Issue #482, opened 2014-05-15
Saving objects without passing "status"


edoardocavazza:
If the "status" of an object already created is not passed to the transaction during the save, BEdita overrides the saved "status" value with the default one. Is this correct or it is a bug?

batopa:
Pull request #565 merged. Closing

PR #565, opened 2014-12-19
#482 Objects' saving fix.

fquffio:
Some strange behaviors occurred when saving objects using the default `Model::save()` method. In some cases, a duplicate save occurred, and some values were overwritten with the default ones.

This isn't a major fix, and it should be quite smooth, but hence objects' saving is involved, I'd prefer somebody else to help me test this.

### DIFF
--- a/bedita-app/app_model.php
+++ b/bedita-app/app_model.php
@@ -1111,11 +1111,12 @@ class BeditaObjectModel extends BeditaSimpleObjectModel {
 		}
 
 		$beObject->create();
-        if ($res = $beObject->save($data, $validate, $fieldList)) {
+        if (!$res = $beObject->save($data, $validate, $fieldList)) {
 			return $res;
 		}
 
 		$data2["id"] = $beObject->id;
+		$this->create(null);
 		$res = parent::save($data2, $validate, $fieldList);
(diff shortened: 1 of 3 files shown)


batopa:
The strange behavior you say occurs only for BEdita object types that extend ```BeditaObjectModel``` class i.e. objects that use only ```objects``` table. It was done because every BEdita object type must have a **hasOne** relation with ```BEObject``` model that uses ```objects``` table too. 
So first it saves BEObject with all its association and then it saves any relation defined in BEdita object type model. It's should be to review, for example, saving the second time only if Model has its associations. But it's better to solve in another issue.

The issue is when ```ForeignDependenceSaveBehavior``` save ```BEObject```. Indeed it does ```BEObject::create()``` and ```BEObject::save()``` but ```Model::create()``` of CakePHP set the default values defined in mysql table in ```$this->data```.

Calling ```Model::create(false)``` it avoids to set the defaults. 
Can you try to use it in ```ForeignDependenceSaveBehavior``` and in ```BeditaObjectModel::save()``` when ```$beObject->create();``` is called?

Also, I will revert the double save behavior for now and in case open another issue on that.

Issue #598, opened 2015-02-18
BeAppModel should look in DB for ObjectType ID if not present in configuration


fquffio:
When saving a BEObject, `BeAppModel::save()` method automatically fills the `object_type_id` key with the ID of the current object type. That method looks for that ID in configuration array, but we should implement a DB lookup fallback in case `$conf->objectTypes` hasn't been populated yet.

PR #599, opened 2015-02-18
#598 ObjectType ID lookup in DB as fallback.

fquffio:
Added method to `BeConfigure` that returns the correct ObjectType ID for the given model name (looks up in configuration array, and in database as a fallback, with basic "caching" to avoid duplicate searches).
Updated `BeAppObjectModel::save()` and `BeditaObjectModel::save()` to use new method when filling the `object_type_id` key.
Tested on both setup and already working instance of BE.
#598

### DIFF
--- a/bedita-app/app_model.php
+++ b/bedita-app/app_model.php
@@ -831,12 +831,10 @@ class BEAppObjectModel extends BEAppModel {
 	}
 
 	function save($data = null, $validate = true, $fieldList = array()) {
-		$conf = Configure::getInstance() ;
-
 		if(isset($data['BEObject']) && empty($data['BEObject']['object_type_id'])) {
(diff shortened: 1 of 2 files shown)


batopa:
Seems ok, but when you write in configuration it should be something like

```php
$objectTypeName = Inflector::underscore($modelName);
$objectType = ClassRegistry::init('ObjectType')->find('first', array(...));
$conf = array(
	'id' => $objectType['ObjectType']['id'],
	'name' => $objectType['ObjectType']['name'],
	'module_name' => $objectType['ObjectType']['module_name'],
	'model' => $modelName
);
Configure.write('objectTypes.{$modelName}.id', $conf);
Configure.write('objectTypes.{$modelName}.{$objectTypeName}', $conf);
```
for consistency with classic object types configuration.

Issue #517, opened 2014-07-31
frontend: use hreflang and different urls for content translations


stefanorosanelli:
https://support.google.com/webmasters/answer/189077?hl=en#
- use tag links in HTML header to mark all traslations available (current page also)  
  `<link rel="alternate" hreflang="es" href="http://es.example.com/" />`
- use different urls for different contents translations/versions

Now we use /lang/xxx at the beginning of urls to change session language and redirect to content, for example using www.example.com/lang/spa/my-content causes:
- change to 'spa' spanish as session language
- redirect to  www.example.com/my-content/ using spanish as language

We should have something like:
- www.example.com/lang/spa/my-content/ that points to the spanish version
- english (or default language) version may be at www.example.com/lang/eng/my-content/ or www.example.com/my-content/
- current behaviour (redirect) stays default in backend
- in frontend we can activate redirect via config parameter
- urls in frontend pages should be in the same (current) language, if available, like  www.example.com/lang/spa/another.content/

fquffio:
Most of the work should have been done by now.
`FrontendController` now handles correctly the `lang` parameter in URLs switching from a language to another.
Language persistence in URL cannot be achieved in an easy way, though. Many of the links in frontends are built the "wrong" way:
```html
<a href="{$html->url('/my_controller/my_action/my_other_param')}">Lorem ipsum</a>
<!-- Instead of the more standard and more stable alternative: -->
<a href="{$html->url(['controller' => 'my_controller', 'action' => 'my_action', 'my_other_param'])}">Lorem ipsum</a>
```
Because of this, we cannot simply prepend `'/lang/xyz'` to all URLs, as it may introduce unexpected behaviors. Hence, it has been introduced a new `BeHtmlHelper` that parses the passed URL using the configured routes, adds the `lang` parameter, and then rebuilds the whole URL passing it back to the `Router` class in the array form. It might look a bit tricky, but it's the best ("cleaner" & simpler) I could figure out by now.
Of course, this needs a soft frontend conversion. A working example would be fquffio/bootstrap@9bfb99cd7ba80e643c65fd55799ae9ec89ffc6ed.

To do:
- [ ] Address and fix an issue with named parameters (that by the way are not massively used in frontends, as far as I know).
- [ ] Add `<meta rel="alternate" />` tags in `<head>` section (this will probably need an additional parameter passed to view, as it would be possible to figure it out only when a single object is displayed).

[...]

PR #616, opened 2015-03-04
#517 SEO-friendly I18n

fquffio:
@batopa @stefanorosanelli Once again, this is a call to arms for you. :wink:
# Changes
- in `FrontendController`, removed redirect in method `lang()`, and changed handling in method `route()`.
- in `BeFrontHelper`, added method `metaAlternate()` that accepts either a valid language string, or an array of valid languages string. It also accepts booleans, where `false` means "no translations", and `true` means "all translations specified in `$conf->frontendLangs`".
- added `BeHtmlHelper` to override `HtmlHelper`'s `url()` and `link()` in order to handle correctly the persistence of the language token in URL even when first argument is passed as a pre-built string (URL token persistence also requires frontends to add a custom routing rule in `router.php`, even if other functionalities' retrocompatibility is guaranteed in case it's missing :neutral_face:).
## Custom routing

In order to correctly handle persistence of `/lang/xyz` in URLs, it is required that multi-language frontends change their routing rules:

``` php
/***** FROM *****/
Router::connect('/*', array('controller' => 'pages', 'action' => 'route'));

/****** TO ******/
Router::connect('/lang/:lang/*', array('controller' => 'pages', 'action' => 'route'), array('lang' => '[a-z]{2,3}', 'persist' => array('lang')));
Router::connect('/*', array('controller' => 'pages', 'action' => 'route'));
```

In case this update is not performed, the site will continue to behave correctly, but internal links will not have `/lang/xyz` as root path, meaning that internal links will not be "fully consistent", **whilst user-displayed language will**, thanks to cookies.


stefanorosanelli:
Patch code seems ok, I need to do some testing before merge... hope to have enough time soon :)

Issue #411, opened 2013-12-05
Update CKeditor to version 4.4


edoardocavazza:
CKeditor 4.3 offers a large number of fixes (from the BEdita current version 4.0.1) and the availability of new widgets (the most important is the render of math in editor and a  dialog for tex insertion).

CKeditor should be compatible with plugins that BEdita currently uses.

didoda:
obsolete

PR #629, opened 2015-03-13
#411 updated ckeditor to 4.4 + math plugin

edoardocavazza:
Updated CKEDITOR to 4.4.x version.

This update introduces many new features to CKEDITOR, but not many breaking changes. BTW, we need to check that all old behaviours are maintained.

Further, we need to test the new math plugin. In order to activate it:
- rename `ckeditor.math.sample` to `ckeditor.js` in `webroot/js/libs/richtexteditors/conf/local`
- in `config/bedita.cfg.php` add `$config['richtexteditor']['local'] = 'ckeditor.js';`


didoda:
obsolete

Issue #671, opened 2015-05-04
Categories' bulk actions


fquffio:
It might be useful to have bulk actions in categories' page, such as bulk delete (that obsoletes atomic deletion) and fast categories merging.
Me might also consider implementing a different way to edit existing categories (either using AJAX or a bulk-save operation) and a lightweight JS filter.

qwerg:
I noticed a 500 error for 'Missing View' on plugins (tickets or questionnaires) after bulk saving or merging.
In core modules it's all ok.

fquffio:
@qwerg Hence many plugin modules overwrite the `forward()` function, for such modules we should add key/item pair:

``` php
"bulkCategories" => array(
    "OK" => "/the_plugin/categories",
    "ERROR" => "/the_plugin/categories"
),
```

to `$REDIRECT` array in `ThePluginController::forward()`. This is due to the fact that the bulk actions use a different controller action than the previous atomic delete, thus needing a custom redirection. For core modules, this has already been done, but for plugin modules I'd rather wait 'till bulk categories actions are a core feature. Should I commit such changes to plugin modules already?

batopa:
Remember that we have to maintain backward compatibility. If changes does not impact plugins installed in old (<= 3.5.1) BEdita we can do it.

fquffio:
@batopa Backward compatibility is fully kept: it's an additional routing rule, that does not affect other ones.

batopa:
@fquffio ok then I agree with you. Updating plugins should be done when the new feature will be merged in main branch

batopa:
I think the merge option is ambiguous.  We should specify what category is maintained or at least write next to the 'merge' button what will happen

fquffio:
@batopa I have improved a bit the information messages. Should be clearer now, please let me know. :smile:

batopa:
Pull request #672 merged

fquffio:
Temporarily reopening as I must update plugin modules as well, to fix misbehavior as of @qwerg comment.

fquffio:
All plugin modules should be ok now. Closing.

PR #672, opened 2015-05-04
Bulk actions and AJAX save for categories

fquffio:
As of #671.


batopa:
I noticed a javascript error editing a category already saved.

```
Uncaught ReferenceError: type is not defined
```

The code is

``` javascript
$('#messagesDiv').empty()
    .html(data.htmlMsg)
    .triggerMessage(type, -1);
```

Issue #670, opened 2015-04-28
Managing dateItems before 1000 AD


qwerg:
In some projects we need to handle dateitems for events and people (card) in remote past.
Current date_items Bedita table is using mysql _datetime_ data type, so we can't manage dates before 1000 AC unless _sqlmode_allow_invalid_dates_ is enabled.
http://dev.mysql.com/doc/refman/5.6/en/sql-mode.html#sqlmode_allow_invalid_dates
Even so, dates BC aren't possible.
So how can we:
- handle dates BC in events and cards (and maybe in other models)
- store a generic 'YEAR' value (2015-00-00??) without months and days
  ?

fquffio:
We could store dates in a `BIGINT` format, like YYYYMMDDHHMMSS (year-month-day-hour-minute-second).
Doing so:
- ordering and comparisons are kept
- BC dates could be treated as negative integers without any issue
- Any "uncertain" date could be stored using some "special" value, which is an invalid date, but follows some convention we can decide
- dates up to 900+M years before and after Christ would be supported.

This would need some caring in building and reading dates, but it should be quite easy to do.
We could also improve performances using a format like the one above transported to binary numbers, instead of using base 10, but this would make dates much less human-readable.

batopa:
To avoid conversion of many data in production databases it's for me preferable to maintain the current format transforming the data type from DATETIME to VARCHAR i.e. 'YYYY-MM-DD HH:MM:SS' and use another field for BC/AC date or the '-' ahead of date.

Using string like that ordering and comparison functions like '<', '>', MAX(), MIN() continue to work properly (also with '-' ahead). I don't know about performance. 

In some rare situations we could save text values (second half of the XII century) losing obviously ordering feature.

fquffio:
@batopa Actually, using the lexicographic order for dates formatted like MySQL datetime seems to carry some problems with dates out of the range 1000-9999 AC:

```
mysql> SELECT * FROM test;
+----+----------------------+
| id | dates                |
+----+----------------------+
|  1 | 2015-08-17 20:30:30  |
|  2 | 20000-10-10 00:00:00 |
|  3 | 0-01-02 03:04:05     |
|  4 | -100-01-02 10:10:10  |
|  5 | -101-10-20 23:23:23  |
|  6 | 300-10-01 17:08:32   |
+----+----------------------+
6 rows in set (0,00 sec)

mysql> SELECT * FROM test ORDER BY dates;
+----+----------------------+
| id | dates                |
[...]

PR #753, opened 2015-10-06
Issue/670 manage historical dates

fquffio:


fquffio:
Tested in #670. Thanks @batopa!

Issue #786, opened 2016-02-10
Update and improve OAuth providers


fquffio:
OAuth providers should:
- [x] be updated in order to handle the currently supported API versions
- [x] be able to assign the created users to a set of specific user groups
- [x] be able to pass a custom redirect URL to the vendor libraries, in order to handle OAuth 2 callbacks in some exotic fashion, where needed

fquffio:
In addition to the already present `key`, `createUser` and `extraUserData` keys in external user auth sources, `groups` and `redirectUrl` have now been added.

Your config file should now look like this:

``` php
<?php

// ...some conf...

/** OAuth config. */
$config['extAuthParams'] = array(
    'facebook' => array(
        'createUser' => true,  // optional, default `false`
        'keys' => array(
            'appId' => 'app-id',
            'secret' => 'secret-key',
        ),
        'extraUserData' => array('some_oauth_scope'),   // optional
        'groups' => array('frontend'),  // optional, default empty array
        'redirectUrl' => $config['serverUrl'] . '/login?authType=facebook',  // optional, default current URL
    ),
    'twitter' => array(
        'createUser' => true,  // optional, default `false`
        'keys' => array(
            'appId' => 'app-id',
            'secret' => 'secret-key',
        ),
        'groups' => array('frontend'),  // optional, default empty array
        'redirectUrl' => $config['serverUrl'] . '/login?authType=twitter',  // optional, default current URL
    ),
    'google' => array(
        'createUser' => true,  // optional, default `false`
        'keys' => array(
            'appId' => 'app-id',
            'secret' => 'secret-key',
        ),
        'extraUserData' => array('some_oauth_scope'),   // optional
        'groups' => array('frontend'),  // optional, default empty array
        'redirectUrl' => $config['serverUrl'] . '/login?authType=google',  // optional, default current URL
    ),
);

// ...some more conf...
```

PR #787, opened 2016-02-10
#786 Update and improve OAuth providers.

fquffio:
See #786.


fquffio:
@edoardocavazza Up!

edoardocavazza:
Missing handling for two behaviors:
- external authentication email is already used by a registered user.
- when `createUser` is set to false in `$config->extAuthParams`, the application crashes when trying to login with a new user.

PR #811, opened 2016-03-23
Issue/791 nickname type cache

stefanorosanelli:


bedita-app/libs/be_object_cache.php
```diff
@@ -92,6 +92,44 @@ private function cacheName($id, array &$options, $label = null) {
...
+    }
+
+    /**
+     * Read id from nickname using cache
+     *
+     * @param  string $nickname object nickname
+     * @return int object id on success, null if $nickname is not found
+     */
+    public function readIdFromNickname($nickname) {
+        if ($this->cacheConfig['engine'] === 'File') {
```
batopa: Maybe better

``` php
if ($this->hasFileEngine()) {
    return null;
}
```

bedita-app/libs/be_object_cache.php
```diff
@@ -92,6 +92,44 @@ private function cacheName($id, array &$options, $label = null) {
...
+
+    /**
+     * Writes $nickname => $id key-value pair in object cache
+     *
+     * @param  string $nickname object nickname
+     * @param  int $id object id
+     * @return boolean true on success, false on failure
+     */
+    public function writeNicknameId($nickname, $id) {
+        if ($this->cacheConfig['engine'] === 'File') {
```
batopa: Maybe better

``` php
if ($this->hasFileEngine()) {
    return false;
}
```

Issue #806, opened 2016-03-15
Implement User and ExternalAuth entities


fquffio:
Basing on the schema structure depicted in [`bedita/bedita4-design`](https://github.com/bedita/bedita4-design), implement basic Table and Entity classes (with tests!) for:
- [x] `Users`
- [x] `ExternalAuths`
- [x] `AuthProviders`
### Questions
- Should we plan additional fields for – let's say – multi-factor authentication (which is not an external auth provider, but rather a second layer of security to the user)? Should we place such additional information in a JSON field?

stefanorosanelli:
About multi-factor auth: this JSON data could be included in the JSON `external_auth.auth_params` field? We may use an explicit key like `MFA` inside that field 
We may also introduce a new/dedicated field for that purpose.

fquffio:
Schema changes proposed:
- `auth_providers.name` could be converted to type `VARCHAR`, and it should be unique
- `auth_providers.url` could be converted to `VARCHAR`, too
- `external_auth.auth_params` and `external_auth.auth_username` could be renamed to `external_auth.params` and `external_auth.username` respectively

### MFA

@stefanorosanelli Uhm… I don't think that the `external_auth` table would fit well. MFA is not a parallel authentication method, but rather an additional step in the authentication process.

As long as I can see, a basic MFA setup (using TOTP generated via a Google Authenticator-compatible app) would consist of:
- a field to store the secret token
- a JSON field to store a set of one-time recovery codes (optional, but _very_ recommended)
- a field to store the last code used (necessary to avoid replay attacks)
- _(optional)_ a JSON field to store enabled U2F FIDO tokens (e.g. YubiKeys)
- _(optional)_ a date field to store information about when MFA was enabled

Of course, instead of adding so many fields to the `users` table, all those columns could be serialized into a single JSON field (even if I wouldn't like this solution) or they could be placed into a separate table structured to handle multiple second-factor authentication methods such as OTPs generated by tokens of several kinds, challenge-response authentication or similar. But honestly I wouldn't see this as a priority, while TOTP generated by Google Authenticator could be an interesting addition to be considered from the very beginning of the development process. :smile:

stefanorosanelli:
Schema changes are ok, proceed

About MFA: I have to read a bit more about the whole TOTP / Google Authenticator process but couldn't those fields, or a single JSON field, be added to `external_auth` table? But I don't see a problem in adding a new table if/when we're going to adopt this mechanism.

PR #819, opened 2016-03-29
Implement authentication-related tables and entities

fquffio:
As of #806, this PR adds `AuthProviders`, `ExternalAuth` and `Users` tables and Entities, together with tests.

It also adds an already deprecated legacy MD5 password hasher, to implement backwards-compatibility with current BEdita 3 users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ cache:
 env:
   matrix:
     - "DB=sqlite db_dsn='sqlite:///:memory:'"
-    - "DB=mysql db_dsn='mysql://travis@127.0.0.1/bedita_test'"
+    - "DB=mysql db_dsn='mysql://travis@0.0.0.0/bedita_test'"
     - "DB=pgsql db_dsn='postgres://postgres@127.0.0.1/bedita_test'"
   global:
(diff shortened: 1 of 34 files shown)


batopa:
A quick observation. 
Maybe `\BEdita\Auth\Database\Type\JsonType` should be stay in `BEdita\Core\Database\Type` namespace. I think it will be used in many Table and make me more sense to stay in core. 
Or is it done to make Auth plugin independent from Core? I.e. Core requires Auth but not vice versa?

fquffio:
@batopa 

> Or is it done to make Auth plugin independent from Core? I.e. Core requires Auth but not vice versa?

I placed it in the `BEdita/Auth` plugin for this very reason. Adding a cross dependency is something I would preferably avoid, as it would require _much_ care in keeping the requirements in sync across `composer.json` files, among other potential issues.

Probably, the most DRY approach would involve an additional plugin (`BEdita/Utils`, or something like that) that both `BEdita/Core` and `BEdita/Auth` depend on. But personally I wouldn't add another plugin unless we realize that the problem we had with `JsonType` is much more widespread than it seems to be right now.

Issue #824, opened 2016-04-11
Add `getPath` cache


fquffio:
Method `FrontendController::getPath()` should have a dedicated caching to avoid repeated and resource-intensive queries on the database.

PR #828, opened 2016-04-18
Path cache implementation

fquffio:
This pull request implements path cache.

For further reference, refer to #824.


stefanorosanelli:
Some minor concerns:
- I think we need to additional cache parameters - namely `$this->publication['id']` and `$this->getStatus()` not just `$objectId` - because we may use a wrong publication or we may have `draft` objects enabled
- we use PHP serialized data in all object cache - I would still use that format for consistency - but for future cache systems (like BE4) JSON should be default choice

Issue #839, opened 2016-04-26
Basic pagination in API response


stefanorosanelli:
Implement paginated response for `/users` endpoint - logic to be reused in other endpoints (where applicable)
JSON API pagination spec here http://jsonapi.org/format/#fetching-pagination

fquffio:
Just a few questions about the spec:
- when serving the first page in the set, `links.prev` should be `null`, right?
- in the same fashion, when serving the last page in the set, `links.next` should be `null`, correct?
- assuming all the results can be displayed in a single page, I guess that both `links.prev` and `links.next` should be `null`, but how about `links.first` and `links.last`? Should `links.last` be `null` as well, or (what I would lean towards) should both links have the same URL to the first page?

stefanorosanelli:
About the spec:
- unavailable **keys** must be `null` or omitted (from the spec) - so it's up to us to decide - is there a solution preferable for a client? @batopa ?
- same logic should be applied to `links.next` and `links.prev`, in case of a single page response I would put same URL on both `links.first` and `links.last`: _last_ page actually is not missing
- there are some additional information we may want to add in a response: total number of items, page size, current page - since they are not links, should they stay in `meta` block?

fquffio:
@stefanorosanelli Completely agree on 2nd and 3rd point. For what concerns the "`null` or omitted" question, I'll wait for an enlightened answer by @batopa and @edoardocavazza. 😉  In the meanwhile, I guess we could consider this a minor change to be made whenever we'll decide which path to follow, couldn't we?

Pagination links and metadata are automatically added whenever the CakePHP's `PaginationComponent` is used. Should we decide to implement our own pagination component or to extend the default one, we might need to update pagination handling in `BEdita/API.JsonApiComponent` in order to restore a seamless integration.

Example response (`GET /users`):

``` json
{
  "data": [
    {
      "id": "1",
      "type": "users",
      "attributes": {
        "username": "blake.jenkins",
        "blocked": false,
        "last_login": null,
        "last_login_err": null,
        "num_login_err": 0,
        "created": "2016-05-09T08:33:01+0000",
        "modified": "2016-05-09T08:33:01+0000"
      }
    },
    {
      "id": "2",
      "type": "users",
      "attributes": {
        "username": "erdman.samir",
        "blocked": false,
        "last_login": null,
        "last_login_err": null,
        "num_login_err": 0,
        "created": "2016-05-09T08:33:01+0000",
[...]

PR #846, opened 2016-05-03
Issue/836 content negotiation

stefanorosanelli:


plugins/BEdita/API/config/bootstrap.php
```diff
@@ -0,0 +1,40 @@
+<?php
+/**
+ * BEdita - a semantic content management framework
```
batopa: Wrong header

plugins/BEdita/API/src/View/JsonApiView.php
```diff
@@ -0,0 +1,71 @@
+<?php
+/**
+ * BEdita - a semantic content management framework
```
batopa: Wrong header

batopa:
I have found an issue with request

``` http
GET /users
```

and no data on database.

The response is an internal server error **500**

``` json
{
    "error": {
        "status": "500",
        "title": "Key `id` is mandatory"
    }
}
```

I will expect a **200** with empty 'data'.

Here is a unit test to simulate it

``` php
public function testEmpty()
{
    $user = TableRegistry::get('Users');
    $user->deleteAll([]);
    $this->configRequest([
        'headers' => [
            'Accept' => 'application/vnd.api+json',
        ]
    ]);

    $this->get('/users');
    $result = json_decode($this->_response->body(), true);

    $this->assertContentType('application/vnd.api+json');
    $this->assertEquals([], $result['data']);
}
```

batopa:
Two things before merge the PR:
1. arranging pagination meta data as described in https://github.com/bedita/bedita/issues/839#issuecomment-217815166
2. as minor issue requests with header `Accept: text/html` (as requests by browser) shown a response containing
   
   ``` json
   {
   "users": [
       {
           "id": 4,
           "username": "hettinger.vida",
           "blocked": false,
           "last_login": null,
           "last_login_err": null,
           "num_login_err": 0,
           "created": "2016-05-09T13:37:30+0000",
           "modified": "2016-05-09T13:37:30+0000"
       },
       {
           "id": 5,
           "username": "sim67",
           "blocked": false,
           "last_login": null,
           "last_login_err": null,
           "num_login_err": 0,
           "created": "2016-05-09T13:37:30+0000",
           "modified": "2016-05-09T13:37:30+0000"
       }
   ]
   }
   ```
   
   Is it expected? I would expect the complete JSON API response.

fquffio:
@batopa Both issues should have been addressed, now. I don't know why Travis CI build hasn't been triggered, but as soon as it will, you may proceed with tests.

HTML view has been achieved using a sub-request: all HTML requests to the API are now redirected to a single action (`\BEdita\API\Controller\AppController::html()`), and a sub-request is triggered with the current URL **but** with JSON Content-Type and Accept headers. This way, the very same serialization logic applies and therefore we are not required to duplicate it in the `Common/html_json.ctp` template anymore.

This approach should work for POST and authenticated requests as well, **but** I haven't configured the sub-request for such scenarios, yet, so we'll need to take care of it as soon as we actually need this feature.

PR #852, opened 2016-05-09
Issue/849 role model

stefanorosanelli:
In `be4-schema.json` _length_ property is now populated, don't know why
I simply did a `db_admin saveSchema` after schema fix

### DIFF
--- a/plugins/BEdita/Core/src/Model/Entity/Role.php
+++ b/plugins/BEdita/Core/src/Model/Entity/Role.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2016 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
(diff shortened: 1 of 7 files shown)


plugins/BEdita/Core/src/Model/Entity/Role.php
```diff
@@ -0,0 +1,44 @@
...
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
+namespace BEdita\Core\Model\Entity;
+
+use Cake\Auth\DefaultPasswordHasher;
```
fquffio: `DefaultPasswordHasher` is never used in this file. This `use` statement could be dropped.

plugins/BEdita/Core/tests/TestCase/Model/Entity/RoleTest.php
```diff
@@ -0,0 +1,88 @@
...
+            'name' => 'patched_role_name',
+        ];
+        $role = $this->Roles->patchEntity($role, $data);
+        if (!($role instanceof Role)) {
+            throw new \InvalidArgumentException();
+        }
+
+        $this->assertEquals(1, $role->id);
+        $this->assertEquals(1, $role->immutable);
+        $this->assertEquals(0, $role->backend_auth);
```
fquffio: I'd prefer to use `assertTrue()` and `assertFalse()` for `$role->immutable` and `$role->backend_auth`, that are supposed to be boolean values. But this is surely a very minor change.

Issue #855, opened 2016-05-11
[API] create /roles endpoint


stefanorosanelli:
- `/roles` - paginated response of roles
- `/roles/1` - single role
- `/roles/1/users` - users having role (paginated)

PR #856, opened 2016-05-11
Implement `/roles` API endpoint

fquffio:
This PR implements `/roles` and `/roles/:role_id/users` API endpoints, as of #855. It also adds `RolesUsersFixture` to `BEdita/Core` plugin.

Please, check carefully that no previously existing functionalities were broken, as some refactors were needed (mainly in routing and pagination links building).


stefanorosanelli:
Seems ok using `json` Accept requests, but weird things happen using `html` Accept:
- exception `Cake\Routing\Exception\MissingRouteException` on endpoint of type `/roles/1`
- pagination links on `/roles` are like

``` json
"links": {
        "self": "http://be4.dev/roles",
        "first": "http://be4.dev/roles?autoRender=0\u0026return=1\u0026bare=1\u0026requested=1",
        "last": "http://be4.dev/roles?autoRender=0\u0026return=1\u0026bare=1\u0026requested=1\u0026page=2",
        "prev": null,
        "next": "http://be4.dev/roles?autoRender=0\u0026return=1\u0026bare=1\u0026requested=1\u0026page=2"
    },
```
- same beahavior on `/users` endpoint  using `html` Accept

stefanorosanelli:
Should we add a unit test to track `/roles/1` like cases using `html` Accept header?

fquffio:
Ouch! Yeah, both are issues arose after 517f9db. I'll take a look.
We might add unit tests to test the HTML display, but I think it would be quite hard to properly test such feature.

stefanorosanelli:
Yes, testing HTML display content is not an easy task, and the HTML response is still not well defined.
But we could simply test response code for urls like `/roles/1` and `/users/1`

Issue #864, opened 2016-05-16
[API] Add link to single item in items list


stefanorosanelli:
In _/roles_ and _/users_ add link to single item using `links.self` in JSON API

PR #866, opened 2016-05-16
#864 feat: links.self on items list + tests

stefanorosanelli:
URLs are created in a very simple way from $type, probably to extend in the future

### DIFF
--- a/plugins/BEdita/API/src/Utility/JsonApi.php
+++ b/plugins/BEdita/API/src/Utility/JsonApi.php
@@ -15,6 +15,7 @@ namespace BEdita\API\Utility;
 use Cake\Collection\CollectionInterface;
 use Cake\ORM\Entity;
 use Cake\ORM\Query;
+use Cake\Routing\Router;
 use Cake\Utility\Hash;
 
 /**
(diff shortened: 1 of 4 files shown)


plugins/BEdita/API/src/Utility/JsonApi.php
```diff
@@ -95,7 +97,14 @@ protected static function formatItem($item, $type = null)
         }
         unset($attribute);
 
-        return compact('id', 'type', 'attributes');
+        if (!$showLink) {
+            return compact('id', 'type', 'attributes');
+        }
+
+        $links = [];
+        $links['self'] = Router::fullBaseUrl() . '/' . $type . '/' . $id;
```
fquffio, line 105: I'd avoid building URLs this way… We might use [named routes](http://book.cakephp.org/3.0/en/development/routing.html#using-named-routes), but this could be a future improvement: this works fine for now. :smile:

stefanorosanelli, line 105: :+1:  like I said in PR description, it's temporary but it works...

Issue #867, opened 2016-05-16
[API] JSON API-compliant `sort` parameter


fquffio:
BEdita API's `PaginatorComponent` should take care of transparently handling JSON API-compliant sorting, as per spec: http://jsonapi.org/format/#fetching-sorting

PR #869, opened 2016-05-16
Implement JSON API-compatible sorting

fquffio:
This PR implements JSON API-compliant sorting, according to [JSON API specification](http://jsonapi.org/format/#fetching-sorting), as required by #867. Specifically:
- `?sort=field` applies ascending order based on `field`
- `?sort=-field` applies descending order based on `field`
- `?sort=non_existing_field` responds with a **400 Bad Request** status

Please note that ordering on multiple fields is **NOT** supported (or at least, is not implemented by this PR), thus `?sort=field1,field2` would result in a 400 response as well.


stefanorosanelli:
Ok 
`?sort=+field` gives a _400_ error - and its ok because is not a JSON API spec

PR #879, opened 2016-05-23
Issue/v4/878 api home document

stefanorosanelli:


plugins/BEdita/API/config/routes.php
```diff
@@ -22,6 +22,11 @@
     ],
     function (RouteBuilder $routes) {
         $routes->connect(
+            '/home',
+            ['controller' => 'Home', 'action' => 'index'],
+            ['_method' => 'GET', '_name' => 'home:index']
```
fquffio: Hence I the home document is actually some sort of meta resource, I'd go for `home` alone, omitting the `:index` suffix.

plugins/BEdita/API/src/Controller/Component/JsonApiComponent.php
```diff
@@ -104,6 +104,7 @@ public function getLinks()
     {
         $links = [
             'self' => Router::reverse($this->request, true),
+            'home' => Router::fullBaseUrl() . '/home',
```
fquffio: Here I think we should totally take advantage of named routes: `Router::url(['_name' => 'api:home'], true)`.

fquffio:
Sorry, I closed this by mistake. PR merged, now.

Issue #911, opened 2016-07-08
API methods to save/remove users and roles


stefanorosanelli:
Implement API POST and DELETE methods to save users and roles entities

PR #912, opened 2016-07-12
Implement CUD actions for `/users` and `/roles` endpoints

fquffio:
This PR implements CUD (Create-Update-Delete) actions for `/users` and `/roles` API endpoints, along with tests. See #911.

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -71,7 +71,7 @@ Router::scope('/', function (RouteBuilder $routes) {
      * You can remove these routes once you've connected the
      * routes you want in your application.
      */
-    $routes->fallbacks('DashedRoute');
+    //$routes->fallbacks('DashedRoute');
 });
 
 /**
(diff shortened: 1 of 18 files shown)


stefanorosanelli:
Everything seems ok, just a couple of things:
- content type `application/json` is not supported but only `application/vnd.api+json` - since in GET methods we are supporting `application/json` just like `application/vnd.api+json` we should apply same rule in POST/PATCH/DELETE (for now)
- we have `JsonApiComponent::allowClientGeneratedIds()` to decide if we allow client generated IDs (see http://jsonapi.org/format/#crud-creating-client-ids) - i don't think we will ever support client IDs so maybe it's better to drop it

@batopa any thoughts on both points?

batopa:
I agree on both points

fquffio:
@batopa @stefanorosanelli Both issues should have been addressed by now.

Issue #914, opened 2016-07-13
Run test suite against multiple database versions


fquffio:
BEdita 4 test suite should be run against several versions of the most popular databases.

Specifically, we should run it with at least MySQL 5.5, 5.6, 5.7; PostgreSQL 9.3, 9.4, 9.5 (and 9.6 beta?); SQLite 3.

batopa:
Merged in PR #916

PR #916, opened 2016-07-14
Run test suite against multiple database versions in Travis using Docker

fquffio:
This PR increments the number of jobs per Travis CI build, as per #914. Tests are run against several versions of each database vendor:
- MySQL versions: 5.5, 5.6, 5.7
- PostgreSQL versions: 9.3, 9.4, 9.5 (tests not run on HHVM)
- SQLite 3

This is achieved using Docker containers for databases. Docker images used are the official ones, and are pulled from the default repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,24 +49,22 @@ matrix:
     - php: hhvm
 
 before_script:
-  - echo "$TRAVIS_PHP_VERSION"
-  - if [ -n "$GH_TOKEN" ]; then composer config github-oauth.github.com ${GH_TOKEN}; fi
-  - composer install --prefer-dist --no-interaction
-
   - |
     if [ "$DB" = 'mysql' ]; then
       docker pull mysql:$DB_VERSION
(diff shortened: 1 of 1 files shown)


fquffio:
Sadly, the systematic error with HHVM build is due to the fact that HHVM is not supported on Trusty builds, which seem to be required for Docker service to be available.

In travis-ci/travis-ci#6091 it is suggested to enable builds on edge infrastructure. As long as we are in this early development phase, I think it could be acceptable to give it a try… @batopa @stefanorosanelli What do you think?

batopa:
Any idea on when will the edge infrastructure become stable?
However we could try it.

fquffio:
Honestly, I have no clue.

fquffio:
To slightly reduce the number of jobs per build, we might limit this multiple-db-versions thing to a single PHP version. For instance, we might run only three builds on PHP 5.5 and 5.6 (SQLite, MySQL, PostgreSQL), plus the complete set only on PHP 7.

fquffio:
@batopa It should be ok by now.

Issue #947, opened 2016-08-05
Upgrade to CakePHP 3.3


batopa:
Also if it is in beta stage we need to upgrade cakephp to take advantage of middlewares for example for #929 and avoid to duplicate effort.

fquffio:
Closing, as PR has been merged.

PR #950, opened 2016-08-24
Upgrade to CakePHP 3.3.*

batopa:
It refers to #947


fquffio:
We have failures on HHVM builds (that are "allowed failures" in our TravisCI config), but this seems to be more of a CakePHP-related problem (see https://travis-ci.org/cakephp/cakephp/builds/154040206), so I'd take it easy.

Issue #926, opened 2016-07-22
API CRUD methods for handle relationships data between users and roles


batopa:
Implement API for

``` http
GET /users/:user_id/relationships/roles
POST /users/:user_id/relationships/roles
DELETE /users/:user_id/relationships/roles
PATCH /users/:user_id/relationships/roles
```

``` http
GET /roles/:user_id/relationships/users
POST /roles/:role_id/relationships/users
DELETE /roles/:role_id/relationships/users
PATCH /roles/:role_id/relationships/users
```

batopa:
Merged by #952

PR #952, opened 2016-08-25
CRUD actions for relationships

fquffio:
This PR implements CRUD actions for relationships, as of #926.


fquffio:
In order to avoid making things too much simple for @batopa, this PR now also includes features as of #954. 😄

batopa:
POST and PATCH don't check the `type` data, so I can do

``` http
POST /roles/1/relationships/users

{
    "data": [
        { "type": "gustavo", "id": "4" }
      ]
}
```

and the relationship between role with id=1 and user with id=4 is created.

batopa:
The command pattern "actions" that save data on db should be handled inside a transaction because if some exception is triggered the data on db are corrupted.

For example, if exists a relationships between roles 1 and users 4 making a request like

``` http
POST /roles/1/relationships/users

{
    "data": [
        { "type": "users", "id": "4" }
      ]
}
```

and an exception is thrown in `\BEdita\Core\Model\Action\AddAssociated::__invoke()` between

``` php
$this->Association->unlink($entity, $relatedEntities);
```

and 

``` php
return $this->Association->link($entity, $relatedEntities);
```

(or fails the `link()` method) the association is removed from db.

Maybe it is enough put in transaction

``` php
// BEdita\API\Controller\ResourcesController::relationships()

if (!$action($id)) {
        throw new InternalErrorException(__('Could not update relationship "{0}"', $relationship));
}
```

fquffio:
@batopa You're right. As far as I can see, the only non-transactional behavior is in `AddAssociated`, right?

batopa:
@fquffio yes it seems to me too

batopa:
`POST`, `PATCH` and `DELETE` requests don't always respond with status code compliant with JSON API specifications http://jsonapi.org/format/#crud-updating-relationship-responses

fquffio:
@batopa Both issues should have been addressed by now.

batopa:
- [x] https://github.com/bedita/bedita/pull/952#issuecomment-243496276 seems not resolved.
- [x] Trying to delete relationships via `PATCH`
  
  ``` http
  PATCH /roles/1/relationships/users
  
  {
     "data": []
  }
  ```
  
  respond with 500 status code with message "Invalid JSON" instead of delete all relationships. If we don't want implement this condition the response should be a `403 Forbidden`.
- [x] Trying to save relationships with repetition respond as `400` with message "Record Not Found"
  
  ``` http
  POST /roles/1/relationships/users
  
  {
      "data": [
          { "type": "users", "id": "2" },
          { "type": "users", "id": "2" }
      ]
  }
  ```
  
  maybe missing an `array_unique` on request data?
- [x] getting relationships of not existing role return a 200 with empty data instead of `404` as from [specification](http://jsonapi.org/format/#fetching-relationships-responses-404) (see the "Note")

fquffio:
For what concerns https://github.com/bedita/bedita/pull/952#issuecomment-243496276, is `409 Conflict` fine? I can see no specific instructions in JSON API spec…

batopa:
`409 Conflict` is correct, I found

> A server MUST return 409 Conflict when processing a POST request in which the resource object’s type is not among the type(s) that constitute the collection represented by the endpoint.
> 
> A server SHOULD include error details and provide enough information to recognize the source of the conflict.

Issue #942, opened 2016-08-04
Shell setup script


stefanorosanelli:
Create a simple shell script for initial setup:
- check database connection
- if no valid db connection
  - ask user to review `app.php` file and insert connection data
  - or ask for host / username / password / driver and update `app.php`
- if database is empty create schema and insert starting data, otherwise ask user to skip action
- check permissions on filesystem
- ask for username and password to define first BEdita4 admin user

stefanorosanelli:
First basic implementation is ready. 
Lot of things could be improved.
- console output styles are basic, we can do better than this :smile: 
- `app.php` is not automatically updated, we simply tell user to check config file
- permissions on filesystem are shown for `logs/` and `tmp/` folders, but no action is taken
- initial password is visible and password is typed only once
- user with id = 1 is modified, no new user is created

fquffio:
If we take advantage of Composer's IO interface, we should have access to a method for asking passwords and hiding the typed answer: https://getcomposer.org/apidoc/master/Composer/IO/IOInterface.html#method_askAndHideAnswer

batopa:
But is ComposerIO interface available without using composer event (post-update, post-install, ...)? I mean for use it normally in CakePHP shell scripts?

If no, Composer uses https://github.com/Seldaek/cli-prompt so we could trying to integrate it for using in CakePHP shell.

fquffio:
PR has been merged. @stefanorosanelli shall we close this issue?

stefanorosanelli:
Ok, issue moved to _Docs_ pipeline, will close afterwards

PR #981, opened 2016-09-16
Issue/v4/942 setup shell

stefanorosanelli:
Setup is working with some limitations (see #942)
Test coverage improved

### DIFF
--- a/plugins/BEdita/Core/src/Shell/BeditaShell.php
+++ b/plugins/BEdita/Core/src/Shell/BeditaShell.php
@@ -106,9 +106,6 @@ class BeditaShell extends Shell
      */
     protected function initSchema()
     {
-        if (!Cache::clear(false, '_cake_model_')) {
-            $this->abort('Unable to remove internal cache before schema check');
-        }
         $currentSchema = Database::currentSchema();
         if (!empty($currentSchema)) {
             $be4Schema = (new JsonConfig())->read('BEdita/Core.schema/be4-schema');
--- a/plugins/BEdita/Core/src/Shell/BeditaShell.php
+++ b/plugins/BEdita/Core/src/Shell/BeditaShell.php
@@ -1,0 +1,220 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2016 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+namespace BEdita\Core\Shell;
+
+use BEdita\Core\Utility\Database;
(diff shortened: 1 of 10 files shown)


plugins/BEdita/Core/src/Shell/BeditaShell.php
```diff
@@ -0,0 +1,217 @@
...
+ *  - internal cache cleanup
+ *
+ * @since 4.0.0
+ */
+class BeditaShell extends Shell
+{
+
+    /**
+     * {@inheritDoc}
+     */
```
fquffio, line 39: Since this method actually does nothing but adding a bit of configuration, I'd add a `@codeCoverageIgnore`, and maybe remove the `BeditaShellTest::testGetOptionParser()`.

We have already used this approach with Table classes (see `RolesTable::initialize()` for instance).

plugins/BEdita/Core/tests/TestCase/Shell/BeditaShellTest.php
```diff
@@ -0,0 +1,132 @@
...
+        $map = [
+            ['username: ', null, 'pippo'],
+            ['password: ', null, 'pippo']
+        ];
+        $this->io->method('ask')
+             ->will($this->returnValueMap($map));
+
+        $info = Database::basicInfo();
+        if ($info['vendor'] != 'mysql') {
+            // TODO: BeditaShell::init works only in MySQL
```
fquffio: Maybe use `$this->markTestSkipped()` or `$this->markTestIncomplete()`?

plugins/BEdita/Core/tests/TestCase/Shell/BeditaShellTest.php
```diff
@@ -0,0 +1,132 @@
...
+            return;
+        }
+
+        $this->BeditaShell->setup();
+
+        $usersTable = TableRegistry::get('Users');
+        $user = $usersTable->get(1);
+
+        $this->assertEquals(1, $user->id);
```
fquffio: Hence `$user` is loaded with `TableRegistry::get('Users')->get(1)`, isn't `$user->id` always 1 regardless of what has happened in `BeditaShell::setup()`?

fquffio (changes requested):

plugins/BEdita/Core/src/Shell/BeditaShell.php
stefanorosanelli: Added `@codeCoverageIgnore` and `@coversNothing` to both `BeditaShellTest::testGetOptionParser()` (same in `DbAdminShell`)

plugins/BEdita/Core/tests/TestCase/Shell/BeditaShellTest.php
stefanorosanelli: used `$this->markTestSkipped` here and in `DbAdminShellTest`

plugins/BEdita/Core/tests/TestCase/Shell/BeditaShellTest.php
stefanorosanelli: Ok, test is uselesses... removed

fquffio (approved):

PR #1002, opened 2016-10-03
Issue/v4/1001 home root

stefanorosanelli:
Some code cleanup/removal

base element do display JSON data in HTML moved to main app - to handle all error cases, also before plugins load


batopa:
I merge the PR.
I'm going to treat the error stuff in #1004

PR #1007, opened 2016-10-07
Issue/v4/1003 status endpoint

stefanorosanelli:


plugins/BEdita/Core/src/Utility/System.php
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Get status information
+     *
+     * @return array Information on environment and datasource/cache connections
+     */
+    public static function status()
+    {
+        $env = 'ok';
+        $errors = [];
+        $check = self::checkPHP();
```
batopa: In general I would prefer the use of `static` keyword instead of `self` to take advantage of [Late State Bindings](http://www.php.net/manual/en/language.oop5.late-static-bindings.php)

plugins/BEdita/API/src/Controller/StatusController.php
```diff
@@ -0,0 +1,36 @@
...
+{
+
+    /**
+     * Show system status info
+     *
+     * @return void
+     */
+    public function index()
+    {
+        $status = System::status();
```
batopa, line 34: Missing check on request method `$this->request->allowMethod('get');`

For example `POST /status` responds as GET request.
Instead it should respond as `405 Method not Allowed`.

batopa (changes requested):

plugins/BEdita/Core/src/Utility/System.php
stefanorosanelli: done

plugins/BEdita/API/src/Controller/StatusController.php
stefanorosanelli: done

batopa (approved):

Issue #1004, opened 2016-10-03
Error response content type and tests


stefanorosanelli:
Something not working as expected in error response (maybe due to cake php upgrade?)

From a JSON-API call:
- 404 errors are returned with `text/html` content type
- 500 errors (like failing db connections) are returned as HTML in body and content type

Also after #1001 
From a browser call:
- 40x and 500 errors are displayed correctly in as JSON in HTML with `debug = 0`
- same errors are displayed using cake templates with `debug=1`

All these cases must be covered by tests
Last point (browser display with `debug=1`) could be postponed

PR #1008, opened 2016-10-11
Error Handling improving

batopa:
It solves #1004 

Assure to update `Error.exceptionRenderer` in `config/app.php` with our `BEdita\API\Error\ExceptionRenderer`.


plugins/BEdita/API/tests/TestCase/Controller/AppControllerTest.php
```diff
@@ -96,20 +250,49 @@ public function contentTypeProvider()
      * @dataProvider contentTypeProvider
      * @covers \BEdita\API\Controller\Component\JsonApiComponent::startup()
      * @covers \BEdita\API\Controller\Component\JsonApiComponent::beforeRender()
+     * @covers \BEdita\API\Error\ExceptionRenderer::render()
      */
-    public function testContentType($expectedCode, $expectedContentType, $accept, array $config = null)
+    public function testContentType($expectedCode, $expectedContentType, $accept, $endpoint, array $config = null)
```
stefanorosanelli: Are we testing too many things? 
Should we have different tests for content type negotiation and generic error responses?

Don't have a solution/proposal but we may probably split this in simpler test methods

Issue #955, opened 2016-08-26
Expose object type as virtual property "type" in entities


fquffio:
To improve consistency and reusability in `JsonApi` utility class, I think it would be a good idea to add a virtual property `type` in object entities, that basically returns the object type name, or "object" as a fallback.

This virtual property could also be writable, and it would provide a simple interface to set the object type on creation.

@batopa @didoda @stefanorosanelli ?

stefanorosanelli:
:+1: if this will be used by the `type` returned by JSON API response it's ok

don't know about `object` as fallback since it's not an actual type

also `Role` entity could have it, and other items that are not objects (like `ObjectType` ?)

fquffio:
@stefanorosanelli Yeah, that was the main purpose of this proposal.

You're right: probably having "object" as a fallback leads to more confusion than having no fallback.

For entities that do not represent a BEdita object (e.g. `Role`, `ObjectType`, …) we could either implement a custom getter-only property, or inflect the type from the table. This way, setting the default type in controller would be only needed when dealing with arrays, but hence most of the time we have hydrated entities, it would require even less configuration.

Basically, I'd change the way the type is retrieved in `JsonApi` utility class to something like:

``` php
if ($item instanceof EntityInterface) {
    if ($item->get('type')) {
        $type = $item->get('type');  // Whether this is an "actual" or virtual property, it's not our business.
    } else {
        $type = $theDefaultTypeSetInController ?: Inflector::underscore(Inflector::pluralize($entity->source()->table()));  // Inflect type from table name as a last resort.
    }
} else {  // $item is array
    $type = !empty($item['type']) ? $item['type'] : $theDefaultTypeSetInController;  // Nothing changes here.
}
```

Issue #971, opened 2016-09-09
Automatic populate object_type_id field in object creation


stefanorosanelli:
In JSON POST data calls avoid to set object_type_id field

Now to create an object you have to use it

``` json
{
    "data": {
        "type": "users",
        "attributes": {
            "object_type_id": 1,
            ......
            "username": "gustavo",
            "password_hash": "gustavo"
        }
    }
}
```

fquffio:
I think this is somehow related to #955.

stefanorosanelli:
That's right,  `type_id` and `type` are both needed - we could probably merge this with #955  
You may proceed if you prefer so

stefanorosanelli:
Seems solved with #955 - moved to `Test`

@fquffio can we close it?

fquffio:
As I wrote in [this comment](https://github.com/bedita/bedita/pull/1026#issuecomment-258476446), it depends on what do we mean with "automatically set". 😄

I guess that current functionality can be considered sufficient. @batopa ?

batopa:
I think it's ok for now. Closing

PR #1026, opened 2016-11-04
Expose `type` virtual property

fquffio:
This PR exposes `type` virtual property in Object entity (and descendants) as of #955. On read, the property is returned as the pluralized form of the object type. The property has also write access (singularized and pluralized forms are both allowed), and it sets the `object_type_id` field accordingly.

JSON API utility library has been slightly refactored to make use of this property, if present. If absent, type is inflected by entity's source table name. As a result, we shouldn't need to set the `_type` view variable anymore.

When using JSON API, the `"type"` key in request payload is used to set the object type without requiring `"object_type_id"` or other fancy stuff to be passed as an attribute. To limit the object types writable by the current controller, one should use `JsonApiComponent` configuration (see `resourceTypes`).


stefanorosanelli:
Does this solve #971 too?

fquffio:
@stefanorosanelli Uhm… yeah, I guess so. It probably depends on what do we mean with "automatically". 🙃  With this proposed merge, we would still need to set `$entity->type = 'object_type_name';`. Maybe we could push it a little further and have the table transparently set the object type name when `$Table->newEntity()` is invoked, but I'm not quite sure this would be a must-have time saver, to be honest. 😕

stefanorosanelli:
To put it simply: if you create an object (like `users`) using POST with `type` but without `objetc_type_id` it just works.
After merge I would close also #971 as duplicate

Issue #967, opened 2016-09-09
Models for endpoint permissions


stefanorosanelli:
Implement Cake models for endpoint permissions system as defined in #920

Issue #968, opened 2016-09-09
Authorization for endpoint permissions


stefanorosanelli:
Implement authorization for endpoint permission using rules and schema from #920 
- handle `all` or `everybody` special role 
- define initial data set for `endpoints` - see #969
- use CakePHP Authorization Handler - http://book.cakephp.org/3.0/en/controllers/components/authentication.html#authorization
- should we handle applications - client apps recognized via api-keys ? or create another issue for that?

fquffio:
After this has been implemented, we should be able to remove `$this->Auth->allow()` in `AppController::initialize()`. Then we could also bind `LoggedUser::setUser()` to the `Auth.afterIdentify` event.

PR #1029, opened 2016-11-08
Models for endpoint permissions

batopa:
It solves #967 introducing basic models (tables and entities) to handle `endpoints`, `applications`, and `endpoint_permissions`.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Entity/Application.php
+++ b/plugins/BEdita/Core/src/Model/Entity/Application.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2016 ChannelWeb Srl, Chialab Srl
+ *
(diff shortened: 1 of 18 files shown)


stefanorosanelli:
Seems fine

Should we already change `ApiKeys` handling? 
http://bedita.readthedocs.io/en/4-develop/configuration.html#apikeys

API KEY check is currently in `AppController::apiKeyCheck()`: should we create a new issue or should this be part of #968?

batopa:
Respect to the current `ApiKeys` handling  the `applications` table missing of `origin` field.

I don't know if  API KEY check should stay in #968. I see the API KEY check as a verification on the application permission rather than on specific endpoint. Maybe I would prefer a middleware for that, so I think a new issue is better.

stefanorosanelli:
Ok, I'll create a new issue for that

Issue #1025, opened 2016-11-03
Distinguish editorial contents from other contents


batopa:
In order to better distinguish contents created by editor users from content created in frontend apps by generic users, we need a way to differentiate the contents.

As *editorial content* we mean:

* a content created from backend
* a content created from frontend configured to be able to create editorial content

We could proceed in two ways:

### Using `objects.publisher` field

We can avoid to add another field using `publisher` field on table `objects`. When populated the content is intended to be created by editor users.

* in backend a configuration would define a default value for `publisher` if no one is specified when an object is created.
* in frontend app users usually should create contents with `publisher` `NULL` unless is specified differently in configuration.

*This field is almost unused but for example is used in [books module](https://github.com/bedita/books) and it seems to me that could create some confusion, specially if a frontend app needs to edit it without make the content as editorial*.

### Adding `objects.editorial` field

This way is cleaner but it expects to add a field to objects table. In this case:

* every object created in backend must have `editorial = 1`.
* every object created in frontend app must have `editorial = 0` unless specified differently in configuration.

---
In both cases we could refine the conf of frontend app defining if just backend users or particular groups may create editorial content.

@stefanorosanelli @fquffio @qwerg @didoda what do you think about it?

stefanorosanelli:
I would use `objects.publisher` field in order to maintain schema structure. There are probably not many use cases needing this feature.
Complexity of both solutions is equivalent, but this one is less _invasive_.

`Editorial` content to me is generally a content created by a user with backend access. In frontend application we may decide other strategies, but as default frontend created contents will be non-editorial.

batopa:
@stefanorosanelli what if a content is created in frontend (not editorial i.e. `objects.publisher IS NULL`) and edited in backend? I would expect that the content remains not editorial but to do that we need two different behaviors on the basis of who has created the object.

In brief editing an object:
- if the object was created in backend (check if `user_created` has backend privilegies) it fills the `publisher` field (if empty) with configured default publisher
- if the object was created in frontend (check if `user_created` doesn't have backend privilegies) it leaves  `publisher` field empty unless edited directly

Anyway we could have some inconsistency when the content is created in frontend from a user with backend privilegies (not editorial by default because is a frontend thing) but later edited in backend. In this situation the content will become an editorial content although it should not.

stefanorosanelli:
I would mark an object as editorial only if we are in a backend environment (through BACKEND_APP constant) creating a NEW object

[...]

PR #1030, opened 2016-11-15
Distinguish editorial contents in backend

batopa:
This PR solves the backend part of #1025 introducing `$config['editorialContents']['defaultPublisher']` to save default `publisher` on object creation and adding the related filter.

### DIFF
--- a/bedita-app/models/behaviors/build_filter.php
+++ b/bedita-app/models/behaviors/build_filter.php
@@ -1200,4 +1200,24 @@ class BuildFilterBehavior extends ModelBehavior {
             }
         }
     }
+
+    /**
+     * Filter editorial contents.
+     *
+     * If `$value` is true it filters by `objects.publisher` not null and not empty
+     * If `$value` is false it filters by `objects.publisher` null or empty
+     *
+     * @param string $s Start quote sql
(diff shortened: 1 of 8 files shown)


bedita-app/models/behaviors/build_filter.php
```diff
@@ -1215,7 +1215,7 @@ protected function allowed_to_userFilter($s, $e, $userid) {
     protected function editorialFilter($s, $e, $value) {
         $field = "{$s}BEObject{$e}.{$s}publisher{$e}";
         if (((bool)$value && $value !== 'false')) {
-            $this->conditions[] = "($field IS NOT NULL && $field <> '')"; 
+            $this->conditions[] = "($field IS NOT NULL AND $field <> '')"; 
         } else {
             $this->conditions[] = "($field IS NULL || $field = '')";;
```
fquffio: I think the second condition should be changed as well (`||` -> `OR`).

stefanorosanelli:
Seems ok

Maybe an INDEX on `objects.publisher` field is necessary?

batopa:
Yes, it would be better. I add the ALTER TABLE to sql upgrade

Issue #1027, opened 2016-11-07
DELETE objects action should move them to trash 


stefanorosanelli:
As discussed in #945 **DELETE** actions on `objects` (and also `users` or other subtypes) should perform a soft delete and move items to *trash*.

Currently we have `objects.status` with `deleted` as thrash option, but we may also use a `deleted` boolean field or a `deleted_at` datetime.

Deleted objects MUST not be visible in `/objects` (and `/users`, `/documents`...) endpoints.

stefanorosanelli:
Did some tests: the best solution IMHO is the use of `objects.deleted` boolean field.
Very easy to setup filters, on restore previous `status` is automatically set.
@bedita/core ?

PR #1032, opened 2016-11-21
Issue/v4/1027 delete trash

stefanorosanelli:
This PR solves #1027 

Conditions on `deleted` field are set in Controllers: I know there are better ways to do that :smiley_cat:


plugins/BEdita/API/src/Controller/UsersController.php
```diff
@@ -153,7 +157,9 @@ public function delete($id)
     {
         $this->request->allowMethod('delete');
 
-        $user = $this->Users->get($id);
+        $user = $this->Users->get($id, [
+            'conditions' => ['deleted' => 0]
+        ]);
         if (!$this->Users->delete($user)) {
```
batopa: The user is really deleted. Add unit test on user delete case

stefanorosanelli: Right.
Solved and modified unit test.
I had to use `ObjectsTable` instead of `UserTable` in `UsersController::delete()` - I'm opening a new issue for that

batopa:
Ok, I would merge it and open a new issue to better handle soft delete. 

I think it's better to use a behavior to have consistency every time a save/find on an object is done. Moreover the behavior could be expose some useful custom finders for example to get just trashed objects.

Issue #1039, opened 2016-11-23
Implement several cache invalidation methods


fquffio:
CakePHP is shipped with a built-in method to invalidate browser's cache for static assets: it basically appends to the URL the timestamp of the asset's last modification time.

We should extend this functionality to handle more generic cases. For instance, we might want to use MD5 or SHA1 hashes, or use a VCS commit hash instead.

PR #1040, opened 2016-11-23
Custom asset timestamp

fquffio:
This PR implements new algorithms to generate hashes to be appended to static resources URLs (see #1039).

By specifying `$config['Asset']['algorithm']` one can choose between:
- `md5` and `md5-short`: calculates MD5 of static asset (or only first 7 characters).
- `sha1` and `sha1-short`: same as above, with SHA1 algorithm.
- `git` and `git-short`: full or short hash commit of Git repository to which file belongs.
- _default_: last modification timestamp.

### DIFF
--- a/bedita-app/views/helpers/be_html.php
+++ b/bedita-app/views/helpers/be_html.php
@@ -33,6 +33,9 @@ class BeHtmlHelper extends HtmlHelper {
      */
     protected $currLang = null;
 
+    /**
+     * Class constructor.
+     */
     public function __construct(array $options = array()) {
         if (!empty($options['currLang'])) {
             $this->currLang = $options['currLang'];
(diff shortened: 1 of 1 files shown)


bedita-app/views/helpers/be_html.php
```diff
@@ -132,14 +144,67 @@ public function hyphen($text, $lang = null, $excludeSelectors = array('.formula'
...
         return $text;
     }
 
+    /**
+     * Append timestamp to an asset to help invalidate browsers' cache.
+     *
+     * @param string $path File path.
+     * @return string
+     */
+    public function assetTimestamp($path) {
```
stefanorosanelli, line 163: Maybe `assetVersion` or `assetRevision`? Timestamp is one of the options

stefanorosanelli (commented):
Code seems ok (apart naming maybe)

How are we going to use this helper method in a template?
Should we maybe introduce `$beHtml->css()` or `$beHtml->script()` ?

bedita-app/views/helpers/be_html.php
fquffio: @stefanorosanelli Sorry, I read your comment just now. Don't know why GitHub notifications aren't working for me. 😕 

Anyway: we don't need `BeHtmlHelper::css()` nor `BeHtmlHelper::script()`: `BeHtmlHelper` extends CakePHP's `HtmlHelper` that implements those methods. Such methods internally invoke `HtmlHelper::assetTimestamp()`, that is overridden by this PR.

Long story short: `$beHtml->script()` and `$beHtml->css()` do the work just fine. If we rename `BeHtmlHelper::assetTimestamp()` to something that fits more the scope of this PR we need to make the magic happen again by overriding `HtmlHelper::css()` and `HtmlHelper::script()` as well.

stefanorosanelli: ok, didn't notice method override

Issue #1031, opened 2016-11-17
[API] Add a way to filter embedded and counted relations


batopa:
In the REST API we need a way in the `ApiBaseController` to set a filter to get a filtered list of embedded or counted relations.

For example, after #1025 we could want to load just *editorial contents* and it can be done on the main objects using `ApiBaseController::$objectsFilter` but we have no way to set a filter getting the related objects embedded/counted.

We can't use `ApiBaseController::$objectsFilter` because the filter on main objects and related objects could be different.

batopa:
Closing as resolved in PR #1041

PR #1041, opened 2016-11-23
Filter embedded and counted relations/children

batopa:
This PR add a way to setup filters for relations and children via controller.  It solves #1031


stefanorosanelli:
PR seems fine, just one thing: how are we dealing with object caching?
Are there possible misuses? Like: wrong data sent to user
Are we potentially create too many caches? Like: one for each user

Not able to test this scenarios right away

batopa:
Mmmm... you are right... anyway even now putting in filter a condition on `BEObject.user_created` we can potentially create a cache for each user using `FrontendController::loadSectionObjects()`.

stefanorosanelli:
Ok, we better deal with this problem in a separate issue - #1042 created
Merging

Issue #1015, opened 2016-10-14
POST /auth param should be `password`


stefanorosanelli:
http://bedita.readthedocs.io/en/4-develop/endpoints/auth.html#login

Current form parameters are: `username` and `password_hash`, we should have `username` and `password` instead

PR #1046, opened 2016-12-21
#1015 feat: password instead of password_hash in login request

didoda:


plugins/BEdita/API/src/Controller/LoginController.php
```diff
@@ -82,6 +82,11 @@ public function login()
     {
         $this->request->allowMethod('post');
 
+        if (!empty($this->request->data)) {
```
stefanorosanelli: I would do
`if (!empty($this->request->data['password']))`
to avoid errors on missing password in POST

Issue #975, opened 2016-09-14
Set `created_by` and `modified_by` properties


fquffio:
When creating or patching an object, `created_by` and `modified_by` should be updated accordingly based on the currently logged in user. So far, they have been arbitrarily set to `1`.

stefanorosanelli:
Through a singleton class like `CurrentUser` where we store/set current user data like `id` and other attributes it would be possible to use that info in the same way from anywhere, also from shell commands or unit test.
@bedita/core ?

fquffio:
PR has been merged. This issue can now be safely closed. 🙃

Issue #1034, opened 2016-11-22
[ORM] `UsersTable::save` not saving `deleted` field


stefanorosanelli:
After #1027 `UsersTable::save` operation not working as expected.
Here a failing unit test method to be added in `UsersTableTest` test case 

```php
    public function testDeleted()
    {
        $user = $this->Users->get(1);
        $user->deleted = true;
        $success = $this->Users->save($user);
        $this->assertTrue((bool)$success);
        $deleted = $this->Users->get(1)->deleted;
        $this->assertEquals(true, $deleted);
    }
```

PR #1052, opened 2017-01-03
Issue/v4/975 created modified by

stefanorosanelli:
This PR solves #975 
    * introduced `LoggedUser` singleton
    * in `AppController::initialize` I've put `LoggedUser::setUser($this->Auth->identify());` - probably not the best place to do that, but I did'nt find information on logged user elsewhere (did we handle this?)
    * did not quite understand how to use behaviors in CTI: should we add them in table models and also in `ObjectTable`? What is the correct assignement in cases like `Users / Profiles / Objects`?
    * `ProfileTableTest::testValidation()` fails removind `$profile->created_by` assignement and adding  `$this->addBehavior('BEdita/Core.UserModified')` in `ProfilesTable`... is it a bug in CTI?

### DIFF
--- a/plugins/BEdita/API/src/Controller/AppController.php
+++ b/plugins/BEdita/API/src/Controller/AppController.php
@@ -13,6 +13,7 @@
 namespace BEdita\API\Controller;
 
 use BEdita\API\Error\ExceptionRenderer;
+use BEdita\Core\Utility\LoggedUser;
 use Cake\Controller\Controller;
 use Cake\Core\Configure;
 use Cake\Event\Event;
@@ -75,6 +76,9 @@ class AppController extends Controller
             'storage' => 'Memory',
         ]);
         $this->Auth->allow();
+        if ($this->name !== 'Login') {
+            LoggedUser::setUser($this->Auth->identify());
+        }
 
         if (empty(Router::fullBaseUrl())) {
             Router::fullBaseUrl(
--- a/plugins/BEdita/Core/src/Model/Behavior/UserModifiedBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/UserModifiedBehavior.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * BEdita, API-first content management framework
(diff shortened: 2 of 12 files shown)


plugins/BEdita/API/src/Controller/AppController.php
```diff
@@ -75,6 +76,9 @@ public function initialize()
             'storage' => 'Memory',
         ]);
         $this->Auth->allow();
+        if ($this->name !== 'Login') {
+            LoggedUser::setUser($this->Auth->identify());
```
fquffio, line 80: After #968, when we will be able to remove `$this->Auth->allow()` just above these lines, this could be probably bound to `Auth.afterIdentify` event.

plugins/BEdita/Core/src/Model/Behavior/UserModifiedBehavior.php
```diff
@@ -0,0 +1,70 @@
...
+ */
+class UserModifiedBehavior extends Behavior
+{
+
+    /**
+     * Default config
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
```
fquffio, line 37: I would have probably preferred an approach more similar to that that can be seen in `TimestampBehavior`, as it provides more flexibility, but nevertheless this could be a future refactor of this behavior if and when we will actually need a more configurable and flexible class.

plugins/BEdita/Core/tests/TestCase/Model/Table/ProfilesTableTest.php
```diff
@@ -129,6 +129,7 @@ public function validationProvider()
     public function testValidation($expected, array $data)
     {
         $profile = $this->Profiles->newEntity($data);
+        // TODO: should work commenting these 2 lines, but it doesn't... why???
```
fquffio, line 132: The problem appears to be in CTI's `ExtensionOf` association. It seems to me that there's a strange bug that causes two very different phenomena: both this and #1034. Hence, this should be a problem that's unrelated to this PR.

fquffio (approved):

Issue #913, opened 2016-07-13
Use database migrations


fquffio:
In order to ensure database is always up to date, we should use database migrations, that make it possible to apply progressive changes in data structure without loss of data.

Despite BEdita 4 is still in an early pre-release phase, we could adopt this method early to become familiar with it before using it in a production-ready product.

fquffio:
#931 introduces several changes to table structure, that would be rather complex to migrate from present structure. As soon as that issue will be solved, we should proceed with migrations.

stefanorosanelli:
In this phase still other **big** changes could (and probably will) be introduced, if that's problematic with `migrations` we could postpone its introduction when we consider db structure more _stable_

Issue #969, opened 2016-09-09
Initial data on DB


stefanorosanelli:
Performing the initial setup as in #942 some initial data set should be inserted on the database in order to get BE4 up and running

Define the initial data set and a (shell?) method to insert it. 
This could also be done using #913 

Initial data should contain
- core object types and properties
- endpoints and their permissions #920 #967 
- default roles

PR #1053, opened 2017-01-04
Use database migrations 🎉 🎊 

fquffio:
This PR solves #913 🎉  — and possibly breaks everything else. 😱 

This is the best attempt I could come up with to break anything we have done so far. 😏 

# Migrations
CakePHP's [migrations plugin](https://github.com/cakephp/migrations) is employed in order to consistently and safely support schema changes on multiple database vendors. The plugin is basically a wrapper for [Phinx](https://github.com/robmorgan/phinx).

## Documentation
Both CakePHP's [migrations plugin](https://book.cakephp.org/3.0/en/migrations.html) and [Phinx](http://docs.phinx.org/en/latest/migrations.html) are extensively documented, so that writing new migrations is a piece of 🍰 . Additionally, CakePHP's migrations plugin integrates with CakePHP's bake plugin to provide super-easy commands like `bin/cake bake migration_diff MyNewMigration` to create a new migration after manually applying schema changes.

Nevertheless, if this PR is to be accepted, we'll need to write some developer docs in our documentation repository.

## Caveats
CakePHP's migrations plugin didn't behave as good as I wished with index names, so this PR adds a bake template to adapt its behavior to our needs.

Also, Phinx seems to have a bug when updating columns that have one or more commas `,` in their SQL definition (see robmorgan/phinx#1006), so as a temporary solution, a custom fixed version is used in place of the official package (see cc04671).

## Initial set of data
When running migrations (e.g. `bin/cake db_admin init`) the minimum data set is automatically loaded into the database. The user can also choose to load additional sets of data (seeds) into their database.

By now, only `plugins/BEdita/Core/config/Seeds/ObjectTypesSeed.php` is available (it adds Documents and News object types). The init script `bin/cake db_admin init` asks the user if they want such additional sets of data to be loaded.

# Side effects
Several components had to be updated, in order for anything to keep working as it was meant to:

- `DbAdminShell` has been widely refactored.
  - It was split into three reusable tasks.
  - Each of these tasks was made it fit for unattended execution, with automated deployments in mind.
  - `CheckSchemaTask` was widely refactored, and additional checks were added to ensure that SQL conventions are enforced.
- `BeditaShell` was almost quite as much revolutionized.
- Test fixtures now rely on `config/Migrations/schema-dump-default.lock` to obtain the latest available schema.
- Possibly others…

## What needs to be tested with special care?
Everything.


fquffio:
Documentation-related tasks:
- [x] [**Developer docs**](https://github.com/bedita/bedita/wiki/%5BBE4%5D-Using-migrations) (_how to create new migrations, test 'em, …_)
- [x] [**End-user documentation of shell commands**](http://bedita.readthedocs.io/en/4-develop/shell.html) (_implemented in bedita/docs#8 and merged in bedita/docs#9_)

stefanorosanelli:
Seems ok to me, did some tests:

 * `bedita setup` shell script seems to checking schema against previous one (the one currently in `4-develop`) - launching it for a second time just to change admin user fails with schema check errors - should we add a unit test for that?

 * in `db_admin init` shell script is not clear if initial data seed is optional or not, some data must be seeded - or should we wait for #969 ?

We could have a `vendor` schema SQL file like before - open an issue for that?

fquffio:
@stefanorosanelli:

- You're right. 😞 Imma fix it, yo! 😎 
- There are two "layers" of data seeds at the moment:
  - the very minimum amount of data that's strictly required for BEdita to operate at a really basic level (e.g.: first user, admin role, User object type and its ancestors) are written in the database regardless of user's choice — it's actually considered part of schema migration.
  - a slightly more extended data seed, that as of this PR includes Document and News object types (mainly as an example), is loaded if and only if the user explicitly asks for it.

    You're right, though: the message is not as clear as it should be. Any suggestions on how it could be improved?
  - in the future, we could perhaps add a third "layer", that could consist of an example directory and document structure just like the initial data set we have in BEdita 3.
- Yes, `.sql` scripts could be generated for current schema. I hope I don't underestimate the effort required, but it shouldn't be that hard, after all. +1 for the proposal.

  I'd develop this feature in a separate issue, though — it sound like something we could put in some shell command similar to `bin/cake spec generate`, but IMO this goes definitely beyond "using database migrations". 😉

stefanorosanelli:
As for data seed layers I would simply insert current data set as default/mandatory without asking: that includes Documents and News types.
We will have an optional extended data set option when we will have something significant. Probably in #969 we can decide this. I agree that having these layers is useful, I would simply add this feature later.

Ok for `.sql` schemas in other issues I'll write that down.

fquffio:
Repeated `bin/cake bedita setup` should now work as expected. There should be no need to add a specific test: tests did already exist, but they were passing "by accident", since while the shell had been refactored, some assertions weren't refactored properly.

Message asking for additional seeds is now more explicit about the fact that it's an optional step: `Would you like to populate your database with an optional set of data?`. I agree that the current additional seed isn't very useful, but I'd rather keep a meaningless placeholder than reverting a change to be re-applied in weeks. If you really wish so, we could maybe comment that part out, or leave a fake prompt and then informing the user that still no optional seed is present.

stefanorosanelli:
`bin/cake bedita setup` behavior is now ok

as for additional seeds: we can leave as it is and comment it out later in case we don't add some optional data in near future

wait for CI tests to succeed to merge

Issue #984, opened 2016-09-16
Implement properties definition on /object_types endpoint


stefanorosanelli:
With /object_types we should be able to handle list and CRUD actions on object types properties.

Probably some `sub-issues` are needed.

fquffio:
PR has been merged.

PR #1060, opened 2017-01-20
Issue/v4/984 properties def

stefanorosanelli:
This PR solves #984 
`/properties` endpoint is introduced and relationships with `object_types` is handled

How to use `properties` will be addressed in a future issue
Also what initial DB data will be needed


plugins/BEdita/API/src/Utility/JsonApi.php
```diff
@@ -130,8 +132,9 @@ function (BelongsToMany $val) {
 
         foreach ($associations as $association) {
             list(, $type) = namespaceSplit(get_class($association));
-            $name = $association->property();
+            $name = Inflector::underscore($association->name());
```
fquffio: Uhm… Why did you change this? Association name might not be the same as the entity's populated property. E.g.:

```
// In a Table's `initialize()` method:
$this->belongsTo('Foos', [
    'propertyName' => 'my_foo_property_name',
]);
```

Also, with "hasOne" associations this will probably fail even with a non-custom configuration, as the property name is in the singular form.

plugins/BEdita/API/src/Controller/ResourcesController.php
```diff
@@ -96,6 +99,9 @@ public function relationships()
         $id = $this->request->param('id');
         $relationship = $this->request->param('relationship');
 
+        $allowAssoc = Inflector::underscore($this->name);
+        $this->set(['_allowedAssociations' => [ $allowAssoc => [$allowAssoc]]]);
```
fquffio: Honestly I don't really like this so much. 😜 It works, but controller name could differ from reverse association's name, especially in objects controller, where we'll have to deal with several models.

I think that this "allowed associations" property should be refactored and exposed at the Entity or Table level. But this could be done in the future. For now, this could probably be changed to something like:

```
// Put these lines just below "$Association = $this->findAssociation($relationship);"

// Try to guess reverse association and implicitly add it to displayed associations.
$reverseAssociation = $Association->target()->association($this->modelClass);
if ($reverseAssociation !== null) {
    $allowAssoc = $reverseAssociation->property();
    $this->set(['_allowedAssociations' => [$allowAssoc => [$allowAssoc]]]);
}
```

fquffio (commented):
I added a couple of notes regarding `allowedAssociations` handling: in both cases things do work fine, so it's more like a personal taste, nothing critical at all.

Also, I'd proceed with a merge of `4-corylus` in this branch before merging this PR, as I fear that a schema migration should be baked to port schema changes applied in this branch.

plugins/BEdita/API/src/Utility/JsonApi.php
stefanorosanelli: Ok, committed

plugins/BEdita/API/src/Controller/ResourcesController.php
stefanorosanelli: Sometimes we would probably want to hide some entity/table associations from controller response, we'll think about that later.
Reverse model associations committed.

stefanorosanelli:
Branch `4-develop` merged in issue branch and migrations updated.

Issue #1049, opened 2016-12-30
Check and modify `unique name` before saving


stefanorosanelli:
After #972 we need to check for actual uniqueness of `uname` field before saving on creation or update.

If `uname` is not unique change it with a simple algorithm (adding hash suffix for instance)

PR #1069, opened 2017-01-31
#1049 feat: uname check and modify

didoda:


plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
```diff
@@ -54,9 +55,47 @@ public function uniqueName(EntityInterface $entity)
...
+    /**
+     * Generate unique name string from $config parameters.
+     * If $regenerate parameter is true, random hash is added to uname string.
+     *
+     * @param \Cake\Datasource\EntityInterface $entity The entity to save
+     * @param array $config parameters to create unique name
+     * @param bool $regenerate if true it adds hash string to uname
+     * @return string uname
+     */
+    public function generateUniqueName(EntityInterface $entity, array $config, $regenerate = false)
```
stefanorosanelli: Since we are going to modify this method, and since it's a critical part of our ORM we need to have a `testGenerateUniqueName` method in `UniqueNameBehaviorTest` - better with `@dataProvider`

stefanorosanelli: For example `$regenerate = true` case is never tested

plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
```diff
@@ -54,9 +55,47 @@ public function uniqueName(EntityInterface $entity)
...
+    {
+        $uname = $config['prefix'] . Text::slug($entity->get($config['sourceField']), $config['replacement']);
+        if ($regenerate) {
+            $hash = sha1(md5($uname));
```
stefanorosanelli: Ok using this hash, but it adds a 40 characters string hash without a dash, difficult to read

I would use less characters, say first or last 6 or add a config to choose hash length with a low default.

plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
```diff
@@ -54,9 +55,47 @@ public function uniqueName(EntityInterface $entity)
         $config = $this->config();
         $uname = $entity->get('uname');
         if (empty($uname) && empty($entity->get('id'))) {
-            $uname = $config['prefix'] . Text::slug($entity->get($config['sourceField']), $config['replacement']);
+            $source = $entity->source();
+            $regenerate = false;
+            do {
```
stefanorosanelli: It works generally, but an unchecked do/while is dangerous

plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
```diff
@@ -54,9 +55,47 @@ public function uniqueName(EntityInterface $entity)
...
+    /**
+     * Verify $uname is unique
+     *
+     * @param string $source table/entity
+     * @param string $uname to check
+     * @return bool
+     */
+    public function uniqueNameExists($source, $uname)
+    {
+        $result = TableRegistry::get($source)->exists(['uname' => $uname]);
```
stefanorosanelli: Probably a `TableRegistry::get('Objects')->exists(['uname' => $uname])` will work, `$source` not needed

plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
```diff
@@ -54,9 +55,47 @@ public function uniqueName(EntityInterface $entity)
         $config = $this->config();
         $uname = $entity->get('uname');
         if (empty($uname) && empty($entity->get('id'))) {
```
stefanorosanelli: What id `$uname` is not empty but conflicting?

Issue #983, opened 2016-09-16
Implement /documents and /news endpoints


stefanorosanelli:
[after #946]

`/documents` and `/news` endpoints should get same response as 
`/objects` with a filter on `documents` or `news` types

as a consequence a filter on types on `/objects` will be implemented (?)

fquffio:
Requires #946.

PR #1071, opened 2017-02-06
Issue/v4/983 implement documents news endpoints

stefanorosanelli:
This PR solves #983 

 * CRUD actions on `/documents`, `/news` and other object type endpoints are implemented

 * `/home` response updated with new dynamic endpoints


plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -61,12 +90,94 @@ public function index()
...
+     *
+     * @return void
+     * @throws \Cake\Network\Exception\BadRequestException Throws an exception if submitted data is invalid.
+     */
+    public function add()
+    {
+        $this->request->allowMethod('post');
+
+        $object = $this->Objects->newEntity($this->request->data);
+        $object->type = $this->request->data('type');
```
batopa, line 119: I think that here is required a check on consistency between `data('type')` and the endpoint used. For example I could send
```http
POST /documents

{
    "type": "users",
    "attributes": {
        "title": "New document"
    }
}
```

A `400 Bad Request` should be expected (missing unit test too)

plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -61,12 +90,94 @@ public function index()
...
+     * @throws \Cake\Network\Exception\ConflictException Throws an exception if object ID in the payload doesn't match
+     *      the URL object ID.
+     * @throws \Cake\Network\Exception\NotFoundException Throws an exception if specified object could not be found.
+     * @throws \Cake\Network\Exception\BadRequestException Throws an exception if submitted data is invalid.
+     */
+    public function edit($id)
+    {
+        $this->request->allowMethod('patch');
+
+        if ($this->request->data('id') != $id) {
```
batopa, line 151: Same validation data on `type` shown above is required

plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -61,12 +90,94 @@ public function index()
...
+     *
+     * @param int $id object ID.
+     * @return void
+     * @throws \Cake\Network\Exception\InternalErrorException Throws an exception if an error occurs during deletion.
+     */
+    public function delete($id)
+    {
+        $this->request->allowMethod('delete');
+
+        $object = $this->Objects->get($id, [
```
batopa, line 186: Missing conditions on object type if specific endpoint is used and 404 Not Found if missing the object to delete

batopa (changes requested):

plugins/BEdita/API/src/Controller/ObjectsController.php
stefanorosanelli: Right, type check added

plugins/BEdita/API/src/Controller/ObjectsController.php
stefanorosanelli: Type check added in edit

plugins/BEdita/API/src/Controller/ObjectsController.php
stefanorosanelli: Type check added here too.
404 not found is handled directly in `$this->Objects->get()` call
Added 404 tests for delete and edit actions

batopa (approved):

Issue #1067, opened 2017-01-27
Plugins for custom models and endpoints


stefanorosanelli:
Custom models and endpoints and generic addons must be implemented as Cake plugins.

Implement basic load/unload mechanism for plugins.

fquffio:
PR has been merged. We can close this issue.

PR #1072, opened 2017-02-08
Issue/v4/1067 custom plugins

stefanorosanelli:
This PR solves #1067 

All custom/3rd party plugins are loaded via `Plugins::loadFromConfig()` reading from `Plugins` configuration key

in `config/app.php` put a `Plugins` key like in `config/app.default.php`

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -370,5 +370,21 @@ return [
      */
     'Accept' => [
         'html' => filter_var(env('ACCEPT_HTML', 'false'), FILTER_VALIDATE_BOOLEAN),
+    ],
+
+    /**
+     * Additional plugins to load with this format: 'PluginName' => load options array
+     * Where options array may contain
+     *
+     * - `debugOnly` - boolean - (default: false) Whether or not you want to load the plugin when in 'debug' mode only
+     * - `bootstrap` - boolean - (default: false) Whether or not you want the $plugin/config/bootstrap.php file loaded.
+     * - `routes` - boolean - (default: false) Whether or not you want to load the $plugin/config/routes.php file.
+     * - `ignoreMissing` - boolean - (default: false) Set to true to ignore missing bootstrap/routes files.
+     * - `autoload` - boolean - (default: false) Whether or not you want an autoloader registered
+     */
+    'Plugins' => [
+        'DebugKit' => ['debugOnly' => true, 'bootstrap' => true],
+        'Migrations' => ['debugOnly' => true],
+//      'MyPlugin' => ['autoload' => true, 'bootstrap' => true, 'routes' => true],
     ]
 ];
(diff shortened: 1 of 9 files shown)


plugins/BEdita/Core/src/Utility/Plugins.php
```diff
@@ -0,0 +1,56 @@
...
+namespace BEdita\Core\Utility;
+
+use Cake\Core\Configure;
+use Cake\Core\Plugin;
+
+/**
+ * Plugins utility class
+ *
+ */
+class Plugins
```
fquffio: I think that having `Plugin` and `Plugins` might lead to some confusion. I'd rather rename this class to `\BEdita\Core\Plugin` and let it extend `\Cake\Core\Plugin`. We could then use `\BEdita\Core\Plugin` everywhere with the added benefits of `loadFromConfig()` method.

config/app.default.php
```diff
@@ -370,5 +370,21 @@
...
+     *
+     * - `debugOnly` - boolean - (default: false) Whether or not you want to load the plugin when in 'debug' mode only
+     * - `bootstrap` - boolean - (default: false) Whether or not you want the $plugin/config/bootstrap.php file loaded.
+     * - `routes` - boolean - (default: false) Whether or not you want to load the $plugin/config/routes.php file.
+     * - `ignoreMissing` - boolean - (default: false) Set to true to ignore missing bootstrap/routes files.
+     * - `autoload` - boolean - (default: false) Whether or not you want an autoloader registered
+     */
+    'Plugins' => [
+        'DebugKit' => ['debugOnly' => true, 'bootstrap' => true],
+        'Migrations' => ['debugOnly' => true],
```
fquffio: **Migrations** should be loaded even if debug mode is turned off. It is required in order to run `bin/cake migrations migrate` on production systems when some update to the database schema is available.

Instead, it should probably be loaded only in CLI environments. Looking at [CakePHP skeleton app](https://github.com/cakephp/app/blob/master/config/bootstrap_cli.php), both **Migrations** and **Bake** are loaded in `config/bootstrap_cli.php`.

fquffio: Another option is to build a wrapper shell around Migration's default shell, and load the plugin only when strictly needed… but loading it in all CLI environments would probably be simpler.

stefanorosanelli: Agreed, we may load *Migrations* and *Bake* explicitly in `config/bootstrap_cli.php` only and remove *Migrations* from app default configuration.

plugins/BEdita/Core/src/Utility/Plugins.php
stefanorosanelli: Done

config/app.default.php
stefanorosanelli: Done

fquffio (approved):

Issue #1078, opened 2017-02-14
Avoid subsections load from selected sections in backend


stefanorosanelli:
Define one or more sections ID via config that will not load their subsection in some publication trees operation in backend. In case of many many (like thousands) subsections these operations are too slow

- AJAX call like `/pages/tree/{id}`
- `/pages/tree/{id}/option`
- `/pages/tree/{id}/checkbox`
- in `model/tree.php` method `getAllRoots(...)`

batopa:
Pay attention that not loading sections in object detail can lead to unwanted behaviors when saving the object as removing the object from hidden sections.

PR #1079, opened 2017-02-14
#1078 feat: avoid subsections load using config excludeFromTreeIds

didoda:
This PR solves #1078 and #1080


bedita-app/app_controller.php
```diff
@@ -1852,6 +1856,29 @@ public function save() {
         $this->{$method}();
     }
 
+    public function readonlyTreePaths($id) {
```
stefanorosanelli: Add PHP doc

Issue #1077, opened 2017-02-13
Upgrade to CakePHP 3.4


batopa:
We need to upgrade to CakePHP 3.4

https://bakery.cakephp.org/2017/02/12/cakephp_340_released.html

https://book.cakephp.org/3.0/en/appendices/3-4-migration-guide.html

fquffio:
As a side effect of upgrading to CakePHP 3.4, the requirement for PHP version rises from 5.5.9 to 5.6.0. We should update that requirement in our `composer.json` as well, and drop Travis CI tests for PHP 5.5.

@batopa @stefanorosanelli Does this look right to you?

stefanorosanelli:
It's ok for me

fquffio:
All the three PRs have been accepted and merged. Closing.

PR #1081, opened 2017-02-14
Upgrade CakePHP to version 3.4

fquffio:
This is the first PR for #1077. It simply bumps the requirement for CakePHP to 3.4 and PHP itself to 5.6, and adjust a couple unit tests to get things working.

There are still a couple of issues with our ExceptionRenderer that should be addressed by cakephp/cakephp#10218, so we should probably wait for a patch release of CakePHP before merging, or deal with the fact that tests will be failing for a short while. 😎


batopa:
It seems ok! I'm waiting for 3.4.1 release to merge the PR.

PR #1087, opened 2017-02-17
Wait until MySQL and PgSQL are up in Travis jobs

fquffio:
This PR _should_ solve a recurrent problem that we have in our Travis CI builds: sometimes MySQL jobs fail because "MySQL server has gone away". Jobs usually pass once restarted.

Being unable to easily debug directly on the system, we can only make assumptions on the root cause of this problem. A possible cause is that the Docker container with MySQL sometimes takes a bit more than usual to start and be able to accept connections, and in the meanwhile tests have already started thus causing an error.

This PR adds wait-for-it, that basically waits up to 15 seconds to see if the specified `host:port` is reachable. As soon as the server becomes reachable, the script continues its execution. If, on the other hand, the server still isn't reachable after 15 seconds, the script exits with a non-zero exit code in the pre-script section that causes an errored (not failed) job.

This approach should have no drawbacks, and it should make us able to discern whether this assumption was actually right.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,12 +49,14 @@ matrix:
     - php: hhvm
 
 before_script:
+  - curl -o wait-for-it.sh https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh && chmod +x wait-for-it.sh
   - |
     if [ "$DB" = 'mysql' ]; then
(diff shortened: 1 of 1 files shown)


stefanorosanelli:
We did receive a nice `zend_mm_heap corrupted` for starters... :smiley: 
At least not the usual `MySQL gone away` message

PR #1089, opened 2017-02-20
Replace deprecated methods

fquffio:
This PR is the second in a set of three. It's related to #1077.

By merging this PR, all uses of deprecated methods in **sources** should be replaced with their recommended counterpart.


plugins/BEdita/API/src/Controller/AppController.php
```diff
@@ -177,34 +177,21 @@ protected function html()
...
-     * Send a 204 No Content response with empty body.
-     *
-     * @return void
-     */
-    protected function noContentResponse()
-    {
-        $this->autoRender = false;
-        $this->response->header('Content-Type: ' . $this->request->contentType());
-        $this->response->statusCode(204);
-        $this->response->send();
```
fquffio, line 208: `Response::send()` method has been deprecated. Therefore I moved all this logic to the invoking points, and a Response object is returned.

Issue #1088, opened 2017-02-18
Unique name not automatically set for documents/news


stefanorosanelli:
Creating new objects like `documents` or `news` - using only `objects` table we are forced to set `uname` field, otherwise create action fails.

Failure example:
```http
POST /documents
{ 
  "data" : {
     "type" : "documents",
     "attributes" : {
         "title" : "A new document",
      },
} 
```

PR #1091, opened 2017-02-20
Issue/v4/1088 uname documents news

stefanorosanelli:
This PR solves #1088


plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
```diff
@@ -54,14 +55,13 @@ class UniqueNameBehavior extends Behavior
...
     public function uniqueName(EntityInterface $entity)
     {
-        $config = $this->getConfig();
         $uname = $entity->get('uname');
         if (empty($uname)) {
-            $uname = $this->generateUniqueName($entity, $config, false);
+            $uname = $this->generateUniqueName($entity);
         }
-        $id = !(empty($entity->get('id'))) ? $entity->get('id') : null;
+        $id = !empty($entity->get('id')) ? $entity->get('id') : null;
```
fquffio: `$entity->get('id')` behaves exactly like this (returns `null` if `id` is not set) and it doesn't add N-path complexity, nor triggers warning because `empty()` wants variables to be passed by reference.

plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
```diff
@@ -116,9 +133,27 @@ public function uniqueNameExists($uname, $id = null)
      * @param \Cake\Event\Event $event The event dispatched
      * @param \Cake\Datasource\EntityInterface $entity The entity to save
      * @return void
+     * @codeCoverageIgnore
```
fquffio: I'd rather keep this line uncovered, even if it's quite trivial. This is not a method that simply adds configuration like `Table::initialize()`…

plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
```diff
@@ -116,9 +133,27 @@ public function uniqueNameExists($uname, $id = null)
...
+     * @param \Cake\Event\Event $event The event dispatched
+     * @param ArrayObject $data The input data to save
+     * @param ArrayObject $options Operation options (unused)
+     * @return void
+     */
+    public function beforeMarshal(Event $event, ArrayObject $data, ArrayObject $options)
+    {
+        if (empty($data['uname']) && empty($data['id'])) {
+            $field = $this->config('sourceField');
+            $fieldValue = !empty($data[$field]) ? $data[$field] : (!empty($data['type']) ? $data['type'] : '');
```
fquffio: I know this might sound kind of pedantic, but I would prefer splitting this into multiple lines, like:
```php
$fieldValue = '';
if (!empty($data[$field])) {
    $fieldValue = $data[$field];
} elseif (!empty($data['type'])) {
    $fieldValue = $data['type'];
}
```

Apart from readability, which is subjective, doing this makes debugging slightly easier — and line coverage would become closer to branch coverage, which is an important metric that's not supported by XDebug, so it's better to avoid complex single-line conditions.

fquffio (changes requested):

plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
stefanorosanelli: I had same doubt writing it, ok
done

plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
stefanorosanelli: done

fquffio (approved):

plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
stefanorosanelli: line removed

Issue #1074, opened 2017-02-11
Evaluate whether controller logic should be refactored into commands/actions


fquffio:
At the present state, some logic that's shared across several controllers is being duplicated.

We have already experimented the commands in our `ResourcesController` to unify logic that's responsible for listing and saving relationships between entities. We should evaluate if and how this "command pattern" can be expanded to cover almost all controller actions, in order to keep our codebase as much DRY as possible.

PR #1095, opened 2017-02-21
Endpoint permissions

fquffio:
This PR implements permissions on endpoint as of #968.

Permissions are different for read and write operations, and can be of four types:
- `false` (`0b00`): no permissions granted
- `true` (`0b11`): full permissions granted
- `'mine'` (`0b01`): permissions granted only on my items — semantic of this is left to controller
- `'block'` (`0b10`): no permissions granted, and ignore all other permissions

In table `endpoint_permissions`, columns `role_id, application_id, endpoint_id` are nullable: in any case, `null` means that that permission applies regardless of the role/application/endpoint. When `role_id` is `NULL`, that permission applies to _any_ user, authenticated _and_ anonymous.

Also, this PR adds:
- a singleton class `CurrentApplication`, that stores the current application. This can be useful to provide context in several places (logging, configuration, …).
- an "anonymous authenticate" authentication method: this was required in order to apply permissions on all users, including those that are not authenticated.

In order to make integration tests work, a massive update of `BEdita\API\Test\TestCase\Controller\*` classes was required.

This PR **does not**:
- implement an `all` role that applies to any authenticated user — how should we do it? should it be automatically inserted in database when initializing it?
- implement custom semantics for `'mine'` (`0b01`) permission — maybe we can wait for #1074?


batopa (approved):
Ok for me, good job!

stefanorosanelli:
For `all` role and `mine` permission we could write two new issues

We should also include some kind of `enable everything` permissions as initial data on DB, to avoid permissions block until we figure out some meaningful defaults (it will take some time, probably)
 
Also the documentation of permission settings is crucial!

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
```diff
@@ -0,0 +1,200 @@
...
+     *
+     * @return \BEdita\Core\Model\Entity\Application|null
+     * @throws \Cake\Network\Exception\ForbiddenException Throws an exception if API key is missing or invalid.
+     */
+    protected function getApplication()
+    {
+        $application = CurrentApplication::getApplication();
+        if ($application === null) {
+            try {
+                CurrentApplication::setFromApiKey(
```
stefanorosanelli, line 107: What if we don't want to use any api key? will it always fail?

fquffio: If `CurrentApplication` singleton already has an application stored in it, API key is not required.

For instance, API key checks can be disabled by setting an application in `config/bootstrap.php` like this:
```php
use BEdita\Core\Model\Entity\Application;
use BEdita\Core\State\CurrentApplication;

// ...
CurrentApplication::setApplication(new Application(['name' => 'Fake app']));
```

This should do the trick, but it's quite obscure. If we plan to have API keys as an optional feature we should develop a more robust and self-explaining solution.

stefanorosanelli: API Keys are not an optional feature, but in many cases like development, test environment or initial setups you may choose not to use them.

Maybe with a simple config like ```['ignoreApikeys' => true]``` or an empty `applications` table we could create a default application at startup. Since we have to remove api keys from config, we may define the expected behavior as we work on that.

Ok for now

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
```diff
@@ -0,0 +1,200 @@
...
+            ->find('byRole', compact('roleIds'))
+            ->find('byApplication', compact('applicationId'))
+            ->find('byEndpoint', compact('endpointIds'));
+    }
+
+    /**
+     * Checks if request can be authorized basing on a set of applicable permissions.
+     *
+     * @param \BEdita\Core\Model\Entity\EndpointPermission[] $permissions Set of applicable permissions.
+     * @return bool|string
```
stefanorosanelli, line 177: possible return values are `true`, `false` and `'mine'` - correct?
it's important to know that `'block'` is never returned

stefanorosanelli:
Very good! Merging

Not so easy to understand and explain the difference between `false` and `'block'` as permissions, we have to write a gooooooood documentation right? :smiling_imp:

PR #1103, opened 2017-02-24
Manage applications, roles, endpoints, and endpoints permissions via shell

### LINKED ISSUE #1096: Manage applications, endpoints and endpoint permissions via shell
We need a way to easily manage Applications, Endpoints and EndpointPermissions entities. While in future we'll want a way to manage them via APIs, an immediate solution that could also be a valuable feature is to provide shell management.

It should be possible to:
- [ ] create a new application (`bin/cake applications create <name> [<description>]`)
- [ ] list existing applications (`bin/cake applications ls [--enabled]`)
- [ ] enable/disable an existing application (`bin/cake applications enable|disable <name|id>`)
- [ ] renew API token for an existing application (`bin/cake applications renew_token <name|id>`)
- [ ] remove an existing application? (`bin/cake applications rm <name|id>`)
- [ ] create a new endpoint (`bin/cake endpoints create <name> [<description>] [--object-type=<name|id>]`)
- [ ] list existing endpoints (`bin/cake endpoints ls [--enabled]`)
- [ ] enable/disable an existing endpoint (`bin/cake endpoints enable|disable <name|id>`)
- [ ] remove an existing endpoint? (`bin/cake endpoints rm <name|id>`)
- [ ] create a new endpoint permission (`bin/cake endpoint_permissions create <read> <write> [--application=<name|id>] [--endpoint=<name|id>] [--role=<name|id>]`)
- [ ] list endpoint permissions (`bin/cake endpoint_permissions ls [--application=<name|id>] [--endpoint=<name|id>] [--role=<name|id>]`)
- [ ] remove an existing endpoint permission? (`bin/cake endpoint_permissions rm [--application=<name|id>] [--endpoint=<name|id>] [--role=<name|id>]`)

I know this quite a bunch of stuff, so it's probably a good idea to split it into multiple sub-tasks… 🤔 Also, we might want to start with a subset of these shell commands, and just add a draft for the remaining with a notice to the end user that the command they're attempting to use is not yet implemented.

didoda:
This fixes #1096


plugins/BEdita/Core/src/Shell/ApplicationsShell.php
```diff
@@ -0,0 +1,177 @@
...
+ * - create: add a new application
+ * - ls: list existing applications
+ * - enable: set enabled an existing application
+ * - disable: set disabled an existing application
+ * - renew_token: renew API token for an existing application
+ * - rm: remove an existing application
+ *
+ * @since 4.0.0
+ */
+class ApplicationsShell extends ResourcesShell
```
stefanorosanelli: Test case for `ApplicationsShell` _coverable_ methods is missing

stefanorosanelli (dismissed):
Test Cases are missing for all shell methods _coverable_ methods. 
Trivial tests may be enough

stefanorosanelli:
Failing tests are code sniffing related, a simple merge from `4-develop` should fix them

stefanorosanelli:
I did a big refactor to simplify the scripts.

We have now a unique `resources` script to handle `applications`, `endpoints`, `roles` and `endpoint_permissions` via `--type` or `-t` option.

Code may be improved but it seems to work and test coverage of the shell script is complete.

The only missing feature is the ability to edit and `endpoint_permission`: it has to be deleted and recreated at the moment.

fquffio (approved):

Issue #1099, opened 2017-02-23
Move migrations data to Core plugin


stefanorosanelli:
After #1075 all `config/Migrations` files should be moved to `plugins/BEdita/Core/config/Migrations`

stefanorosanelli:
@fquffio  after wiki update you can close the issue

fquffio:
[Wiki page](https://github.com/bedita/bedita/wiki/%5BBE4%5D-Using-migrations) updated. I'm closing this issue. Feel free to reopen it if something isn't OK with docs.

Issue #1075, opened 2017-02-11
Bake models for all database tables


fquffio:
We still have some tables in our database that do not have a specific model class.

This is quite fine as long as we do not use such tables, but has some drawbacks anyway. For instance, we are forced to keep our Migrations in the App's namespace, because if they were in *BEdita/Core* plugin (as they are supposed to be) some shell commands like `bin/cake bake migration_diff ...` would not work as expected.

Therefore we should bake Table classes for all database tables that still miss one on their own.

stefanorosanelli:
Missing Tables have been baked
 * default tests are incomplete, keep them like that?
 * entities tests are missing, should we create them?

During `bake` operation i got `<warning>Bake cannot generate associations for composite primary keys at this time</warning>` for `object_relations` and `relation_types` - something to worry about? :smiley:

@batopa @fquffio

PR #1104, opened 2017-02-24
Move migrations to BEdita/Core

fquffio:
This PR solves #1099 by moving all migration classes baked so far to the `BEdita/Core` plugin, where they should stay.

Database administration commands have been updated accordingly.

Also, table `RolesUsers` has now its dedicated model class — it was left out in #1075.


plugins/BEdita/Core/config/Migrations/20170222170901_LocationsTable.php
```diff
@@ -1,4 +1,17 @@
 <?php
+
+/**
+ * BEdita, API-first content management framework
```
stefanorosanelli, line 4: we can skip copyright header for migrations (i guess)

stefanorosanelli:
Seems ok, merging

Issue #1076, opened 2017-02-11
Display list of relationships in JSON API response for BEdita objects


fquffio:
Whenever a JSON API response contains a BEdita object, the object should list its available relationships under the `relationships` key (see [JSON API spec](http://jsonapi.org/format/#document-resource-object-relationships)). E.g.:

```javascript
{
    "id": 42,
    "type": "documents",
    "attributes": {
        // ...
    },
    "relationships": {
        "attach": {
            // ...
        },
        "seealso": {
            // ...
        },
        // ...
    }
}
```

fquffio:
This could probably be obtained using an approach that might also have some benefits in the future: generating dynamic associations.

For instance, let's say that for the object type *Document* there are two available relationships: _attachment_ and _see_also_. When the Table object is instantiated we should somehow add two BTM associations:

```php
$this->belongsToMany('Attachments', [
    'className' => 'Objects',  // Always link to the Objects table
    'through' => 'ObjectRelations',  // Specify "through" model.
    // `through` should probably be an instance of the table, instead of the name,
    // in order to allow filtering by relation type, in case this does not work:
    'conditions' => [
        'ObjectRelations.relation_id' => TableRegistry::get('Relations')->find()->where(['name' => 'attachment'])->first()->id,
    ],
]);
$this->belongsToMany('SeeAlsos', [ // Nice name, huh?
    // Same as above.
]);
```

stefanorosanelli:
Seems a good idea. 
But can we use another naming conventions for those BTM associations?
A good relation name is a predicate in the semantic RDF style triple we are using in BEdita: pluralize a predicate seems weird :smiley:

batopa:
Made some preliminary test.

1. Create two fake relations
 ```sql
 INSERT INTO `relations` (`id`, `name`, `label`, `inverse_name`, `inverse_label`, `description`, `params`) VALUES
(1, 'attachment', 'attachment', 'attached_to', 'attached to', NULL, NULL),
(2, 'seealso', 'see also', 'seealso', 'see also', NULL, NULL);
 ```
2. Manually fill the `object_relations` table
3. Add in `ObjectsTable::initialize()`
  ```php
$this->belongsToMany('AttachmentRelations', [
    'className' => 'Objects',
    'through' => 'ObjectRelations',
    'foreignKey' => 'left_id',
    'targetForeignKey' => 'right_id',
    'conditions' => [
[...]

PR #1105, opened 2017-02-24
Display available relationships in JSON API response

fquffio:
This PR _should_ solve #1076.

We use the approach described in the issue and tested by @batopa : when a BEdita object model is instantiated, associations are dynamically created for each relationship type that that object type can have.

For instance, if we set up a `see_also` relationship that can have **Documents** on the left side, the `Documents` table will have a `SeeAlso` association set up (belongs to many with `Objects` table). Therefore, getting a list of "see also" objects for a document should be possible via `TableRegistry::get('Documents')->get($documentId)->see_also`.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
@@ -38,6 +38,8 @@ class ProfilesTable extends Table
         $this->setPrimaryKey('id');
         $this->setDisplayField('name');
 
+        $this->addBehavior('BEdita/Core.Relations');
+
         $this->extensionOf('Objects', [
             'className' => 'BEdita/Core.Objects'
         ]);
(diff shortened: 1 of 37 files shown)


plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
```diff
@@ -38,6 +38,8 @@ public function initialize(array $config)
         $this->setPrimaryKey('id');
         $this->setDisplayField('name');
 
+        $this->addBehavior('BEdita/Core.Relations');
```
stefanorosanelli, line 41: @fquffio @batopa should this behavior be in `ObjectsTable` only? Or when should we add it on Table classes extending objects?

batopa: I think that It should be added in every table that represents an object, i.e. in ObjectsTable, ProfilesTable, UsersTable, ...

stefanorosanelli: Ok, I checked and on `/users` and `/locations` relations are not shown - so, yes, it has to be added there too - but this raises the other question: what are mandatory behaviors in this table classes? what are not? Ok we can answer later to that :smiley_cat:

batopa (approved):

PR #1107, opened 2017-02-27
Use more developer friendly endpoint permissions

fquffio:
This PR refactors recently introduced endpoint permissions in two ways:

- anonymous applications are now allowed by default. Requests that do not have an API key are treated as anonymous application. This behavior can be disabled through a configuration key that has been introduced and documented in `config/app.default.php`.
- if the table `endpoint_permissions` is completely empty, it is assumed that full permissions are allowed to anyone anywhere.

### DIFF
--- a/plugins/BEdita/API/src/Auth/EndpointAuthorize.php
+++ b/plugins/BEdita/API/src/Auth/EndpointAuthorize.php
@@ -76,6 +76,11 @@ class EndpointAuthorize extends BaseAuthorize
         $application = $this->getApplication();
         $endpoint = $this->getEndpoint();
         $permissions = $this->getPermissions($user, $application, $endpoint)->toArray();
+        if (empty($permissions) && (!$endpoint || $this->getPermissions(false, $application, $endpoint)->count() === 0)) {
+            $this->authorized = true;
+
+            return $this->authorized;
+        }
 
         $this->authorized = $this->checkPermissions($permissions);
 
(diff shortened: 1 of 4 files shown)


plugins/BEdita/API/src/Controller/AppController.php
```diff
@@ -70,7 +70,11 @@ public function initialize()
 
         $this->loadComponent('Auth', [
             'authenticate' => ['BEdita/API.Jwt', 'BEdita/API.Anonymous'],
-            'authorize' => ['BEdita/API.Endpoint'],
+            'authorize' => [
+                'BEdita/API.Endpoint' => [
+                    'disallowAnonymousApplications' => Configure::read('disallowAnonymousApplications'),
```
stefanorosanelli: use ```Configure::read('Security.disallowAnonymousApplications')```
if we want to add this to `Security` key as in `config/app.default.php`

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
```diff
@@ -178,12 +182,21 @@ protected function getPermissions($user, Application $application, Endpoint $end
      */
     protected function checkPermissions(array $permissions)
     {
+        $result = EndpointPermission::PERM_NO;
+        if (empty($permissions)) {
+            $count = TableRegistry::get('EndpointPermissions')->find()->count();
```
stefanorosanelli: Count should be done on current endpoint not on `any` endpoint
If I set a permission only on `/roles` endpoint I will get `401 Unauthorized` also trying to access `/home`

What I would do on empty `$permissions`:
  * if `$endpoint` is `null` (no endpoint and no permissions found) => YES
  * if `$endpoint` is set and no permissions are set for that endpoint (`$count === 0`) => YES
  * otherwise => NO

plugins/BEdita/API/src/Controller/AppController.php
fquffio: Oops! My bad.

PR #1116, opened 2017-03-01
#1116 refactor: remove `pluralized` and add `singular`

stefanorosanelli:
This PR is a refactor of `object_types`:
 * `name` is plural, since actual type names are used in plural form
 * `singular` is the singular version of a type name (used in entities and elsewhere)
 * `pluralized` was removed


plugins/BEdita/Core/config/Migrations/20170301144829_RemovePluralized.php
```diff
@@ -0,0 +1,73 @@
...
+                'null' => true,
+            ])
+            ->update();
+
+        $this->query('UPDATE object_types set singular = name');
+        $this->query('UPDATE object_types set name = pluralized');
+
+        $this->table('object_types')
+            ->changeColumn('singular', 'string', ['null' => false])
+            ->update();
```
fquffio: Why not `$this->table('object_types')->renameColumn('name', 'singular')->update()`, then `$this->table('object_types')->renameColumn('pluralized', 'name')->update()` ?

plugins/BEdita/Core/tests/TestCase/Model/Entity/ObjectTypeTest.php
```diff
@@ -135,39 +135,39 @@ public function testSetName()
...
      * @param string $name Object type name.
-     * @param string|null $pluralized Object type pluralized name.
+     * @param string|null $singular Object type singular name.
      * @return void
      *
-     * @dataProvider getSetPluralizedProvider
-     * @covers ::_getPluralized()
-     * @covers ::_setPluralized()
+     * @dataProvider getSetSingularProvider
+     * @covers ::_getsingular()
```
fquffio: It should probably be `_getSingular()`. I'm not sure whether casing does matter to PHPUnit…

Same problem on next line.

plugins/BEdita/Core/tests/TestCase/Model/Entity/ObjectTypeTest.php
```diff
@@ -190,7 +190,7 @@ public function testGetAlias()
     }
 
     /**
-     * Data provider for `testGetSetPluralized` test case.
+     * Data provider for `testGetSetsingular` test case.
```
fquffio: This is a data provider for `testGetSetTable`. The comment was wrong before, but maybe we can take the chance to fix.

fquffio (commented):
Minor annotations as usual…

plugins/BEdita/Core/config/Migrations/20170301144829_RemovePluralized.php
stefanorosanelli: much better, I found some issues when I first tested it that way, but somehow I was wrong, it works!

plugins/BEdita/Core/tests/TestCase/Model/Entity/ObjectTypeTest.php
stefanorosanelli: yesss

plugins/BEdita/Core/tests/TestCase/Model/Entity/ObjectTypeTest.php
stefanorosanelli: :+1:

Issue #1113, opened 2017-02-28
View related objects in JSON response


stefanorosanelli:
In JSON object response we now have the expected relation links after #1076 :
```json
"relationships": {
  "{relation_name}": {
      "links": {
          "self": "https://api.example.com/objects/{id}/relationships/{relation_name}"
       }
   }
}
``` 
`links` urls `self` and `related` should respond with a paginated list of related objects

`GET /objects/{id}/{relation_name}` must return a list of related objects
`GET /objects/{id}/relationships/{relation_name}` must return same list with basic data, usually just `id` and `type`

PR #1119, opened 2017-03-02
List related objects

fquffio:
This PR implements both `GET /:objectType/:id/:relation` and `GET /:objectType/:id/relationships/:relation` endpoint, and should solve #1113.

In order to display relationships links correctly, a minor refactor of `JsonApi` utility class was necessary. As a result, a `JsonApiTrait` has been introduced to expose some useful properties (e.g. type, or list of available relationships) at the entity level.

Furthermore, routing has been slightly simplified. In #1074 we might be able to simplify it furtherly.

### DIFF
--- a/plugins/BEdita/API/src/Controller/ObjectsController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectsController.php
@@ -19,8 +19,10 @@ use Cake\Network\Exception\ConflictException;
 use Cake\Network\Exception\InternalErrorException;
 use Cake\Network\Exception\NotFoundException;
 use Cake\Network\Exception\NotImplementedException;
+use Cake\ORM\Query;
 use Cake\ORM\TableRegistry;
 use Cake\Routing\Router;
+use Cake\Utility\Inflector;
 
 /**
  * Controller for `/objects` endpoint.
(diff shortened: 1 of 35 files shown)


plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -211,4 +254,42 @@ public function delete($id)
...
+    {
+        $this->request->allowMethod(['get', 'post', 'patch', 'delete']);
+
+        $id = $this->request->getParam('id');
+        $relationship = $this->request->getParam('relationship');
+
+        $Association = $this->findAssociation($relationship);
+
+        switch ($this->request->getMethod()) {
+            case 'PATCH':
```
stefanorosanelli, line 273: should also PATCH and POST throw `NotImplementedException` ?

fquffio: Uhm… they already should. 🤔  Those three `case ...:` without a `break` mean that in either of these three cases, the `NotImplementedException` is thrown (see 4th example [here](http://php.net/manual/en/control-structures.switch.php)).

stefanorosanelli: that's right

PR #1130, opened 2017-03-08
Use command pattern

fquffio:
This PR does what has been described in #1074.

### Controllers

All logic in resources' controllers (`RolesController`, `PropertiesController`, `ObjectTypesController`) has been moved to dedicated "Actions" that are used in the abstract `ResourcesController`. Said concrete controllers now simply extend the abstract one and add a couple of lines of configuration. This should make code maintenance slightly easier, and adding controller for new resources (e.g.: Relations) a piece of cake.

In the same fashion, some logic of the `ObjectsController` has been refactored in dedicated actions, and `UsersController` is now a bare extension of `ObjectsController` that adds just a couple of lines of configuration — basically, `UsersCtrl` is to `ObjectsCtrl` what `RolesCtrl` is to `ResourcesCtrl`.

`TrashController` could have been refactored as well to use the same actions as `ObjectsCtrl`, but hasn't for mental sanity reasons: this PR already brings a lot of changes in code, we can take care of that later, maybe.

### Routing

Routing was slightly simplified as well. Routing rules were merged and adding rules for new controllers should be possible by simply adding one line in `APP/plugins/BEdita/API/config/routes.php`. `JsonApi` utility class had to be changed a bit in order to work with new routing rules. Hopefully, such changes also made Scrutinizer happy.

### Actions

Actions should reflect all actions that one can possibly perform on entities, and (since they belong to the Model layer) can be used in Controllers, Shells, and maybe even other Models. So far, actions exist for:

1. listing entities
2. retrieving a single entity
3. saving or modifying an entity
4. deleting an entity
5. listing entities related to a given entity via a specific relationship
6. completely replacing entities related to a given entity via a specific relationship
7. adding entities related to a given entity via a specific relationship
8. removing a subset of entities related to a given entity via a specific relationship

Due to their different nature, most of these actions will actually have a counterpart dedicated to handling of BEdita objects. At the present state, all actions have their "Object"-counterpart, except: 3 (useless, since it would have been exactly the same action — we can introduce a dedicated one in the future, if needed), 6-7-8 (not yet implemented).

Actions for listing entities and objects already come with a `filter` option, that can perhaps be useful in the future.


plugins/BEdita/API/src/Controller/ResourcesController.php
```diff
@@ -97,15 +94,126 @@ protected function findAssociation($relationship)
...
                 }
             }
         }
 
-        throw new NotFoundException(__('Relationship "{0}" does not exist', $relationship));
+        throw new NotFoundException(__d('bedita', 'Relationship "{0}" does not exist', $relationship));
+    }
+
+    /**
+     * Paginated list of resources.
```
stefanorosanelli: `index()` now is used also for entity creation
I like having simplified routing rules, but It seems to me that `index()` and `view()` do too many different things - breaking Single Responsibility Principle
What about a having `add()` `edit()` `delete()` like before, with routing method?

@batopa what do you think?

plugins/BEdita/API/src/Controller/ResourcesController.php
```diff
@@ -97,15 +94,126 @@ protected function findAssociation($relationship)
...
+        $data = $this->paginate($query);
+
+        $this->set(compact('data'));
+        $this->set('_serialize', ['data']);
+
+        return null;
+    }
+
+    /**
+     * View entity's details.
```
stefanorosanelli: like previous comment - `view()` is doing also `delete()` and `edit()`

plugins/BEdita/API/src/Utility/JsonApi.php
```diff
@@ -77,6 +76,44 @@ protected static function extractType($item, $type)
...
+
+            try {
+                $url = Router::url($options, true);
+            } catch (MissingRouteException $e) {
+                $options['object_type'] = $type;
+
+                try {
+                    $url = Router::url($options, true);
+                } catch (MissingRouteException $e) {
+                    // Do not halt if route is missing.
```
stefanorosanelli, line 109: Could be there a problem with this exception? Should we log something?

plugins/BEdita/API/src/Controller/UsersController.php
```diff
@@ -22,10 +22,8 @@
  * Controller for `/users` endpoint.
```
stefanorosanelli, line 16: Unused `use` statements - not part of this PR, but since class logic is inherited they are not needed

plugins/BEdita/API/src/Controller/ResourcesController.php
fquffio: The problem with the previous approach was that if a request with a not supported HTTP verb reached a known endpoint, the response would have been a "Not Found" instead of a "Method Not Allowed" (i.e. `DELETE /users` would have returned 404).

Indeed, this approach does break SRP, but allows us to correctly detect a method not allowed. Having protected methods `add()`, `edit()` and `delete()` and conditionally invoking them (`if ($this->request->is('post') { $this->add($entity); }`) is an option. However, we had already broken SRP in the `relationships()` method, that handles `GET`, `POST`, `PATCH` and `DELETE`, so we should probably refactor that as well.

plugins/BEdita/API/src/Utility/JsonApi.php
fquffio: Surely it's a weird scenario. I tried re-throwing the exception, and the whole test suite does pass, so we can probably assume that in this case something is going wrong, and the developer wants to be informed about that.

plugins/BEdita/API/src/Controller/ResourcesController.php
stefanorosanelli: we can think of a small refactor later, for now:
 * change doc block in `index()` and `view()`
 * rename `view()` to something like `handle()` ?? or other name, calling `view` to *delete* seems weird :smiley:

batopa: Also to me `index()` and `view()` sound weird because of their multiple actions, specially `view()`, but I think we can refactor their names in another issue.

PR #1133, opened 2017-03-09
#1132 feat: dbadmin massRemove of many objects fix

### LINKED ISSUE #1132: shell dbadmin massRemove - error when too many objects
Shell `dbadmin massRemove - type <objectType>` can be problematic when there are too many objects for <objectType>. We could bypass this issue by:
 - providing an `iterationLimit` on objects to delete (delete `iterationLimit` objects per iteration)
`iterationLimit` could be an argument passed to the shell script, with a default value.

didoda:
It fixes #1132

### DIFF
--- a/bedita-app/vendors/shells/dbadmin.php
+++ b/bedita-app/vendors/shells/dbadmin.php
@@ -885,30 +885,34 @@ class DbadminShell extends BeditaBaseShell {
 		$count = $model->find('count', array('conditions' => array('BEObject.object_type_id' => $objTypeId)));
 		$skipids = array();
 		$conditions = array('BEObject.object_type_id' => $objTypeId);
+		App::import('Component', 'Transaction');
+		$transaction = new TransactionComponent('default');
 		while ($count > 0) {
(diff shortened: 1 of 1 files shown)


bedita-app/vendors/shells/dbadmin.php
```diff
@@ -859,31 +859,78 @@ public function clearMediaCache() {
...
 		}
-		$res = $model->deleteAll("BEObject.object_type_id = '$objTypeId'");
-		if($res == false) {
-			$this->out("Error removing items");
-			return;
+		$counter = 0;
+		$count = $model->find('count', array('conditions' => array('BEObject.object_type_id' => $objTypeId)));
+		$skipids = array();
+		$conditions = array('BEObject.object_type_id' => $objTypeId);
+		while ($count > 0) {
```
batopa: It should be better use a transaction to avoid data inconsistency if some delete fails.

batopa (changes requested):

batopa (approved):

Issue #1126, opened 2017-03-08
Date range display and filter with query string


stefanorosanelli:
* view date range elements in JSON response
 * implement query string filter (after #1074) to select object having date ranges with a filter like: `filter[dates][start]=2017-01-01&filter[dates][end]=2017-03-31`

fquffio:
👍  PR merged. I'm closing the issue.

PR #1138, opened 2017-03-13
Issue/v4/1126 date range display filter

stefanorosanelli:
This PR solves #1126 

Most important changes:

-  we can associate entity dynamically using `object_type.associations` field containing a JSON array of associated entities
- `Events` core object type was added + seed
- `filter[date_ranges]` implementation 
- input datetime data are saved using a subset of ISO8601 format - namely the format addressed here https://www.w3.org/TR/NOTE-datetime - minus YYYY and YYYY-MM formats that may be implemented later if necessary
- datetime type custom database handler was added to implement this datetime input format
- integration tests added in `tests/IntegrationTest` with base class `ApiIntegrationTestCase`: associated entities and filter query string  - NB: these test methods MUST have @coversNothing annotation


plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
```diff
@@ -97,4 +95,52 @@ public function buildRules(RulesChecker $rules)
...
+     */
+    public function findDateRanges(Query $query, array $options, $objectAlias = 'Objects')
+    {
+        $options = array_intersect_key($options, ['start_date' => 0, 'end_date' => 0]);
+        if (!empty($options)) {
+            $conditions = [];
+            $subopts = ['gt' => '>', 'lt' => '<', 'ge' => '>=', 'le' => '<='];
+            foreach ($options as $key => $value) {
+                if (is_array($value) && ($k = key($value)) && !empty($subopts[$k])) {
+                    $conditions[] = sprintf("%s.%s %s '%s'", $this->getAlias(), $key, $subopts[$k], $value[$k]);
```
fquffio: Building conditions this way is highly discouraged, AFAIK, as it opens up a way for SQL injections…
It should probably be something like:

```php
$cond = sprintf("%s.%s %s", $this->getAlias(), $key, $subopts[$k]);
$conditions[$cond] = $value[$k];
```

Or, perhaps even better, use `QueryExpression` instead… 🤔

plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
```diff
@@ -97,4 +95,52 @@ public function buildRules(RulesChecker $rules)
...
+                }
+            }
+
+            if (!empty($conditions)) {
+                $subquery = $this->find()->select(['id'])
+                            ->where(sprintf('%s.object_id = %s.id', $this->getAlias(), $objectAlias))
```
fquffio: ~Why not `$query->innerJoinWith()` ?~

**EDIT:** Ok, I think that this finder shouldn't assume that a join with `objects` has already been done. It should probably be more "agnostic" on the Query object it receives. IMHO it should simply filter `date_ranges` by date, regardless of their `object_type_id`. It is `ObjectsTable`'s duty to `INNER JOIN` with `date_ranges` if and when it is necessary. See comment in `ObjectsTable.php`.

plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
```diff
@@ -97,4 +95,52 @@ public function buildRules(RulesChecker $rules)
...
+                            ->andWhere($conditions);
+                $query = $query->where(function ($exp, $q) use ($subquery) {
+                        return $exp->exists($subquery);
```
fquffio: Wrong indentation.

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -170,4 +175,16 @@ public function findType(Query $query, array $options)
...
+    /**
+     * Find by date range using `DateRanges` table findDate filter
+     *
+     * @param \Cake\ORM\Query $query Query object instance.
+     * @param array $options Array of acceptable date range conditions.
+     * @return \Cake\ORM\Query
+     */
+    public function findDateRanges(Query $query, array $options)
+    {
+        return TableRegistry::get('DateRanges')->findDateRanges($query, $options, $this->getAlias());
```
fquffio: Here I think it would probably be better something like:

```php
return $query->innerJoinWith('DateRanges', function (Query $query) use ($options) {
    return $query->find('dateRanges', $options);
});
```

plugins/BEdita/API/src/Utility/JsonApi.php
```diff
@@ -132,7 +132,13 @@ protected static function extractAttributes(array $item)
...
             $item,
             function (&$attribute) {
-                if ($attribute instanceof \JsonSerializable) {
+                if (is_array($attribute)) {
+                    foreach ($attribute as &$value) {
+                        if (is_array($value)) {
+                            unset($value['id'], $value['object_id']);
+                        }
+                    }
+                } elseif ($attribute instanceof \JsonSerializable) {
```
fquffio: I tried adding `id` and `object_id` to the `DateRange::$_hidden` array, and it seems to work exactly the same, with the benefit of keeping `JsonApi` utility class slightly more agnostic. I'd prefer that approach instead.

plugins/BEdita/API/tests/TestCase/Utility/JsonApiTest.php
```diff
@@ -563,4 +563,44 @@ public function testFallbackLinks()
...
+     * Test {@see \BEdita\Core\Utility\JsonApi::formatData()} and
+     * {@see \BEdita\Core\Utility\JsonApi::extractAttributes()} methods.
+     *
+     * @return void
+     *
+     * @covers ::formatData
+     * @covers ::formatItem
+     * @covers ::extractAttributes
+     */
+    public function testFormatDataArray()
```
fquffio: If you choose to follow the advice for `JsonApi` utility class, this test becomes obsolete — and will be the only one to fail in the test suite, which is correct.

plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
```diff
@@ -97,4 +95,52 @@ public function buildRules(RulesChecker $rules)
...
+     * @return \Cake\ORM\Query
+     */
+    public function findDateRanges(Query $query, array $options, $objectAlias = 'Objects')
+    {
+        $options = array_intersect_key($options, ['start_date' => 0, 'end_date' => 0]);
+        if (!empty($options)) {
+            $conditions = [];
+            $subopts = ['gt' => '>', 'lt' => '<', 'ge' => '>=', 'le' => '<='];
+            foreach ($options as $key => $value) {
+                if (is_array($value) && ($k = key($value)) && !empty($subopts[$k])) {
```
fquffio: So we can have just one condition on each field, right? Filters like `['start_date' => ['>' => '...', '<=' => '...']]` aren't allowed.

fquffio (changes requested):
I have found a couple of issues. I might take care of those, if you wish.

plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
stefanorosanelli: One condition now, but we may extend it if necessary

plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
stefanorosanelli: Yes, other `sanitation` schemes are possible I guess
I was thinking on doing it later, but doing it *now* it's ok :wink:

plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
stefanorosanelli: Ok, do you suggest any change?

plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
stefanorosanelli: That's too bad!! :smiley:

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
stefanorosanelli: Seems ok

plugins/BEdita/API/src/Utility/JsonApi.php
stefanorosanelli: Ok, better use `$_hidden`

batopa (approved):

fquffio (approved):

PR #1139, opened 2017-03-13
feat: dbadmin shell versions, limit object versions

didoda:
This provides method `versions` on `dbadmin` shell.
It limits versions per object, by limit passed as argument, i.e.
```
// remove versions, limiting to a maximum of 10 versions per object
./cake.sh dbadmin versions -limit 10 
```

### DIFF
--- a/bedita-app/vendors/shells/dbadmin.php
+++ b/bedita-app/vendors/shells/dbadmin.php
@@ -1028,6 +1028,80 @@ class DbadminShell extends BeditaBaseShell {
 		$transaction->commit();
 	}
 
+	/**
+	 * Versions of objects util: limit to <n> values per object, by param '-limit'
+	 *
+	 * @return void
+	 */
+	public function versions() {
+		if (empty($this->params['limit'])) {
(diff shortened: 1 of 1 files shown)


bedita-app/vendors/shells/dbadmin.php
```diff
@@ -1028,6 +1028,80 @@ function clonePublication() {
...
+				foreach ($versions as $versionId => $revision) {
+					if ($counter > 0) { // update row => set new revision number to $counter
+						$insertList[] = $versionId;
+					}
+					$counter--;
+				}
+				$objectVersions = array();
+				if (!empty($insertList)) {
+					$revision = 1;
+					$objectVersions = $versionModel->find('all', array(
```
batopa, line 1073: Should it be ordered by revision desc?

batopa: ehm... asc

didoda: by revision asc (default). biggest number => more recent version.

Issue #1122, opened 2017-03-06
Relation parameters


fquffio:
After #1114 we'll also need a proper handling of relation parameters.

Each relation defined can have `relations.params` column populated with a definition of allowed parameters. The proposed format for such definition is [JSON Schema](http://json-schema.org/).

Then, each row in `object_relations` can have its `object_relations.params` column populated if and only if the JSON in `object_relations.params` validates against the schema defined in the related `relations.params` column.

Open questions:
1. What should happen if a relation definition changes when relations of that kind already exist? Should we re-validate all params? How should we behave in case issues arise?
2. How are relation parameters displayed in JSON API response? [According to the spec](http://jsonapi.org/format/#document-resource-identifier-objects) it looks like we should put all these information in the `meta` object in each relationship definition.
3. What's the correct response in case a relationship params fail to validate? **400 Bad Request** or rather silently ignoring such parameters?

fquffio:
@bedita/core Please share your opinions. 😃

stefanorosanelli:
About the 3 questions:
 2. i would put in `meta` as you suggest
 3. if a relationship param fail to validate, i would avoid error (as Robustness Principle suggests) - some sort of warning/information in a `meta` response could be useful, in general, not just here
 
As for 1. probably we will have to revalidate everything or keep everything and give warnings on single requests as params are no more valid. Difficult to say at the moment, but some strategy is needed. 
Or we can send user threatening messages as he tries to change a definition :smiling_imp:

Issue #1114, opened 2017-02-28
CRUD operations on related objects


stefanorosanelli:
[after #1113]
Using URLs like `/objects/{id}/relationships/{relation_name}` CRUD operations on related object are to be implemented.
 
In terms of HTTP methods and URLs:
`POST|PATCH|DELETE /objects/{id}/relationships/{relation_name}`

PR #1150, opened 2017-03-17
BEdita object relations CRUD

fquffio:
This PR solves #1114 and (partially) #1122 by allowing relationships management between BEdita objects.

Relationships between BEdita objects are many-to-many relationships, thus `PATCH`, `POST` and `DELETE` methods have been implemented, as per [JSON API specification](http://jsonapi.org/format/#crud-updating-to-many-relationships).


plugins/BEdita/API/src/Model/Action/UpdateAssociatedAction.php
```diff
@@ -14,12 +14,14 @@
...
 use BEdita\Core\Model\Action\BaseAction;
+use Cake\Database\Expression\QueryExpression;
 use Cake\Datasource\Exception\RecordNotFoundException;
 use Cake\ORM\Association;
+use Cake\ORM\Association\BelongsToMany;
 use Cake\Utility\Hash;
 
 /**
- * Command to add links between entities.
+ * Command to update links between entities.
```
stefanorosanelli, line 24: Why in `BEdita/API` and not in `BEdita/Core`? seems quite *core*

Why same name as the abstract `UpdateAssociatedAction` in *core*?

stefanorosanelli: Some more info in doc block may help

fquffio: This action is only a wrapper around concrete implementations of `BEdita\Core\Model\Action\UpdateAssociatedAction`. We introduced in `BEdita\API` in #926  to parse the `$request` data, fetch all required entities and pass all the pre-populated data structures to the downstream Action. Therefore, I think it fits in BEdita/API, since another invocation of the downstream action wouldn't probably have a Request object ready for use… 🤔 

👍  for improving docblock. I'm on it.

PR #1153, opened 2017-03-20
Replace existing date ranges on save

### LINKED ISSUE #1152: Patching an entity with `date_ranges` adds new ranges instead of replacing
When issuing a `PATCH` request on a BEdita object that already had one or more `date_ranges`, the ranges passed in the request are added to the list of object's ranges, instead of replacing existent ones.

fquffio:
This PR fixes #1152. Now, existing date ranges are replaced when an entity is patched.

This has a drawback: since date ranges IDs are not exposed via API, every time an object is saved _all_ associated date ranges are deleted, and new ones are saved even if nothing at all has changed. I think we should think of some sort of optimization to avoid `date_ranges` table `AUTO_INCREMENT` go crazy. 😄  This should probably be achieved with a `_setDateRanges()` custom setter in `ObjectEntity` class.

### DIFF
--- a/plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectsTableTest.php
+++ b/plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectsTableTest.php
@@ -3,7 +3,6 @@ namespace BEdita\Core\Test\TestCase\Model\Table;
 
 use Cake\ORM\TableRegistry;
 use Cake\TestSuite\TestCase;
-use Cake\Utility\Text;
 
 /**
  * {@see \BEdita\Core\Model\Table\ObjectsTable} Test Case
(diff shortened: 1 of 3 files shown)


plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectsTableTest.php
```diff
@@ -3,6 +3,7 @@
 
 use Cake\ORM\TableRegistry;
 use Cake\TestSuite\TestCase;
+use Cake\Utility\Text;
```
stefanorosanelli: unused import?

fquffio: Oooops. 🙃

stefanorosanelli:
`Replace` strategy seems fine now
Don't see any harm in having ID's iwth potentially high values for now....

Issue #1154, opened 2017-03-20
Unique names are not generated properly in CTI


fquffio:
When a table extends another table via CTI, some behaviors (`UniqueNameBehavior`, for instance) are not triggered properly unless specifically bound to the descendants tables as well.

Currently we have some issues when saving objects with similar titles, as `UNIQUE` constraints on `objects.uname` column prevent latter saves.

stefanorosanelli:
We have the same behavior also with `documents` object type, using only `objects` table: 400 error when using an existing title... 

Another tricky scenario:
 * create a document with empty title - OK
 * create another document with empty title - 400 response

PR #1157, opened 2017-03-22
Avoid errors on uname validation

batopa:
This PR solves #1154

The issue wasn't with CTI but with validation.
`beforeMarshal`  was removed and all the magic happens in `beforeSave()`.

### DIFF
--- a/plugins/BEdita/API/tests/IntegrationTest/UniqueNameTest.php
+++ b/plugins/BEdita/API/tests/IntegrationTest/UniqueNameTest.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2017 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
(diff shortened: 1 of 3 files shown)


plugins/BEdita/API/tests/IntegrationTest/UniqueNameTest.php
```diff
@@ -0,0 +1,79 @@
...
+        $sendRequest = function ($type) use ($attributes) {
+            $data = [
+                'type' => $type,
+                'attributes' => $attributes,
+            ];
+            $endpoint = '/' . $type;
+            $requestBody = json_encode(compact('data'));
+            $this->configRequestHeaders('POST');
+
+            $this->post($endpoint, $requestBody);
```
stefanorosanelli, line 59: More general remark: should we remove data we insert in a unit test?
in some tests we expect e predefined number of objects (from fixtures data)
Here is not a problem, but do we need a rule for that? just wondering...

fquffio: All tables are already automatically truncated and re-populated at the end of each test. Database data cleanup is not our concern.

stefanorosanelli: Yes, but the problem was inside some of our first test cases like `ObjectsControllerTest::testAdd` - probably we should never rely on a predefined number of records in a test

fquffio: Oh, I see what you mean. Yeah, I totally agree! Tests that rely on entities count should probably look like:

```
$count = $Table->find()->count();
$Table->save($myBrandNewEntity);
static::assertSame($count + 1, $Table->find()->count());
```

Issue #1158, opened 2017-03-22
[API] Multilang behavior


stefanorosanelli:
Using a query string like `lang=fra` data JSON response textual fields should be translated in the requested language, if a translation was found. Similar to what happens internally in a BEdita frontend.

PR #1159, opened 2017-03-22
#1158 feat: api get objects query string lang

didoda:
This provides query string `lang` handling for `objects` endpoint.

### DIFF
--- a/bedita-app/controllers/api_base_controller.php
+++ b/bedita-app/controllers/api_base_controller.php
@@ -262,7 +262,7 @@ abstract class ApiBaseController extends FrontendController {
      * @var array
      */
     private $defaultAllowedUrlParams = array(
-        '__all' => array('access_token', 'lang'),
+        '__all' => array('access_token'),
         '_pagination' => array('page', 'page_size'),
         'objects' => array('id', 'filter[object_type]', 'filter[substring]', 'filter[query]', 'embed[relations]', '_pagination', 'lang'),
         'posters' => array('id', 'width', 'height', 'mode')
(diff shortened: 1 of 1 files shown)


bedita-app/controllers/api_base_controller.php
```diff
@@ -726,6 +732,44 @@ public function childrenFilter($conditions = null) {
...
     /**
+     * Setup locale settings
+     * override FrontendController::setupLocale
+     * override AppController::setupLocale
+     *
+     * @see bedita-app/controllers/FrontendController#setupLocale()
+     * @see bedita-app/AppController#setupLocale()
+     */
+    protected function setupLocale() {
+        $lang = $this->lang;
```
stefanorosanelli: Do we need `$this->lang` ?
We could set `$this->currLang` just once, from query string making sure it exists in config: just see if `Configure::read('frontendLangs.' $lang)` exists

bedita-app/controllers/api_base_controller.php
```diff
@@ -726,6 +732,44 @@ public function childrenFilter($conditions = null) {
...
+        if ($this->currLang === null || empty($this->currLang)) {
+            $conf = Configure::getInstance();
+            if (!empty($lang) && array_key_exists($lang, $conf->frontendLangs)) {
+                $this->currLang = $lang;
+            } else {
+                // HTTP autodetect
+                $l10n = new L10n();
```
stefanorosanelli: HTTP lang autodetect from browser is not needed, could be misleading too

stefanorosanelli (changes requested):

bedita-app/controllers/api_base_controller.php
```diff
@@ -726,6 +726,24 @@ public function childrenFilter($conditions = null) {
     }
 
     /**
+     * Setup locale settings
+     * override FrontendController::setupLocale
+     * override AppController::setupLocale
+     *
+     * @see bedita-app/controllers/FrontendController#setupLocale()
```
stefanorosanelli: should be
```
     * @see FrontendController::setupLocale()
     * @see AppController::setupLocale()
```

bedita-app/controllers/api_base_controller.php
```diff
@@ -726,6 +726,24 @@ public function childrenFilter($conditions = null) {
...
+     * @see bedita-app/AppController#setupLocale()
+     */
+    protected function setupLocale() {
+        if (!empty($this->currLang)) {
```
stefanorosanelli: There's a flow problem:
 * `setupLocale()` is invoked first from `AppController::beforeFilter()` when `$this->currLang` is empty
 * `$this->currLang` is set after in `getObjects()` with no config check and write

Also setting `currLang` in view is probably not needed

stefanorosanelli (changes requested):

stefanorosanelli (approved):

PR #1161, opened 2017-03-22
Use specific target table for relations when possible

fquffio:
This PR refactors `RelationsBehavior` in a way that uses a specific target table for relations whenever possible.

For instance, if the relation `my_relation` can have only Profiles on the right side, the behavior now creates a BelongsToMany association with `ProfilesTable` — before the association's target was always `ObjectsTable` regardless of the object types allowed on the other side.

At the moment, this works only for relations that accept only one object type on the opposite side. Surely this behavior can be made more clever, but this can be a start. 🙃

### DIFF
--- a/plugins/BEdita/Core/tests/TestCase/Model/Behavior/RelationsBehaviorTest.php
+++ b/plugins/BEdita/Core/tests/TestCase/Model/Behavior/RelationsBehaviorTest.php
@@ -37,6 +37,7 @@ class RelationsBehaviorTest extends TestCase
         'plugin.BEdita/Core.relation_types',
         'plugin.BEdita/Core.objects',
         'plugin.BEdita/Core.profiles',
+        'plugin.BEdita/Core.locations',
     ];
 
     /**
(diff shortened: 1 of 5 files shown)


plugins/BEdita/Core/tests/TestCase/Model/Behavior/RelationsBehaviorTest.php
```diff
@@ -53,10 +55,43 @@ public function testInitialization()
...
+        static::assertSame(2, $Profiles->objectType()->id);
 
         static::assertInstanceOf(BelongsToMany::class, $Documents->association('Test'));
+        static::assertSame('BEdita/Core.Objects', $Documents->association('Test')->className());
         static::assertInstanceOf(BelongsToMany::class, $Documents->association('InverseTest'));
+        static::assertSame('BEdita/Core.Objects', $Documents->association('InverseTest')->className());
         static::assertInstanceOf(BelongsToMany::class, $Profiles->association('InverseTest'));
+        static::assertSame('BEdita/Core.Objects', $Profiles->association('InverseTest')->className());
+        static::assertInstanceOf(BelongsToMany::class, $News->association('AnotherTest'));
+        static::assertSame('BEdita/Core.Locations', $News->association('AnotherTest')->className());
```
batopa: Missing test on inverse with just an object type.

batopa (changes requested):

batopa (approved):

Issue #1148, opened 2017-03-16
Raise Bad Request error on invalid filter in query string


batopa:
When a request is sent with an invalid filter in query string the response should be a `400 Bad Request`

Example

```http
GET /documents?filter[invalid_filter]=tryThis
```

The same error should be raised if the filter is valid but the request can't be satisfied for some reason, for example invalid params

```http
GET /locations?filter[geo][banana]=12,56
```

Other example may be if the database used doesn't support a function implemented in the filter.

batopa:
Closing after PR #1165

PR #1165, opened 2017-03-24
Issue/v4/1148 filters bad request

stefanorosanelli:
This PR solves #1148 

Probably not all possible cases are covered, db version check is not optimal, but main requirements should be fulfilled


plugins/BEdita/Core/src/Model/Action/ListEntitiesAction.php
```diff
@@ -138,6 +140,13 @@ protected function buildFilter(Query $query, array $filter)
 
                 continue;
             }
+
+            // No suitable filter was found
+            //$this->log('Filter not found ' . $key, 'error');
+            throw new BadRequestException([
```
fquffio: I'm not quite sure it is a good idea to raise a `BadRequestException` from within the model layer.

Perhaps I'd prefer a `new \LogicException('...', 400)`. Or, even better, create a custom `BadFilterCallException` that has `400` as its default code.

stefanorosanelli: Agreed, I had same doubt
Probably a custom exception is the best solution
`LogicException` from PHP guide: <<_Exception that represents error in the program logic. This kind of exception should lead directly to a fix in your code._>>

stefanorosanelli: `BadFilterException` has been introduced

fquffio (approved):

PR #1166, opened 2017-03-25
Superpower date ranges 💪🏼💊

fquffio:
This PR adds some utility methods to `DateRanges` entity class. It also fixes an issue with marshalling DateRanges when using a PostgreSQL backend.

### Comparison methods

Methods are introduced to:
- `isBefore()`: check if a date range is before another date range.
- `isAfter()`: check if a date range is after another date range.

In both cases, if the date ranges being compared do overlap, they are considered "not sortable", thus both `isBefore()` and `isAfter()` methods will return `false`.

### Operations on unions of date ranges

Static methods are introduced to:
- `normalize()`: normalize an array of date ranges — sort by start date, merge overlapping date ranges.
- `union()`: compute set union of two or more arrays of date ranges.
- `diff()`: compute [set difference](https://en.wikipedia.org/wiki/Complement_(set_theory)) between two arrays of date ranges. Special thanks to @didoda.

In all cases "[singleton](https://en.wikipedia.org/wiki/Singleton_(mathematics))" date ranges (i.e. date ranges with `end_date === null`) are considered irrelevant when compared to an interval. For instance, computing the relative complement of the interval `[2017-01-01 00:00:00, 2017-01-10 23:59:59]` with respect to the unit set `{2017-01-05 12:00:00}` will have no effect. Doing the opposite (`{2017-01-05 12:00:00} \ [2017-01-01 00:00:00, 2017-01-10 23:59:59]`) will result in an empty set, as well as subtracting two identical unit sets (`{2017-01-05 12:00:00} \ {2017-01-05 12:00:00}`).

### DIFF
--- a/plugins/BEdita/Core/src/Model/Entity/DateRange.php
+++ b/plugins/BEdita/Core/src/Model/Entity/DateRange.php
@@ -51,7 +51,8 @@ class DateRange extends Entity
      * @param \BEdita\Core\Model\Entity\DateRange $dateRange Date Range being compared.
      * @return bool
      */
-    public function isBefore(DateRange $dateRange) {
+    public function isBefore(DateRange $dateRange)
+    {
         return $this->start_date < $dateRange->start_date && ($this->end_date === null || $this->end_date <= $dateRange->start_date);
     }
 
@@ -61,17 +62,19 @@ class DateRange extends Entity
      * @param \BEdita\Core\Model\Entity\DateRange $dateRange Date Range being compared.
      * @return bool
      */
-    public function isAfter(DateRange $dateRange) {
+    public function isAfter(DateRange $dateRange)
+    {
         return $this->start_date > $dateRange->start_date && ($dateRange->end_date === null || $this->start_date >= $dateRange->end_date);
     }
 
     /**
      * Normalize an array of Date Ranges by sorting and joining overlapping Date Ranges.
      *
-     * @param \BEdita\Core\Model\Entity\DateRange[] $dateRanges
+     * @param \BEdita\Core\Model\Entity\DateRange[] $dateRanges Set of Date Ranges.
      * @return \BEdita\Core\Model\Entity\DateRange[]
      */
-    public static function normalize(array $dateRanges) {
+    public static function normalize(array $dateRanges)
+    {
         if (empty($dateRanges)) {
             return [];
         }
@@ -162,7 +165,6 @@ class DateRange extends Entity
 
                     break;
                 }
-
 
                 $dateRange->start_date = $dateRange2->end_date;
 
--- a/plugins/BEdita/Core/src/Model/Entity/DateRange.php
+++ b/plugins/BEdita/Core/src/Model/Entity/DateRange.php
@@ -21,7 +21,7 @@ use Cake\ORM\Entity;
  * @property int $id
  * @property int $object_id
  * @property \Cake\I18n\Time $start_date
- * @property \Cake\I18n\Time $end_date
+ * @property \Cake\I18n\Time|null $end_date
  * @property array $params
  *
  * @property \BEdita\Core\Model\Entity\ObjectEntity $object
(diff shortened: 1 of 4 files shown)


plugins/BEdita/Core/src/Model/Entity/DateRange.php
```diff
@@ -44,4 +44,191 @@ class DateRange extends Entity
...
+     * for this purpose it assumes the same value as `start_date`.
+     *
+     * **Warning**: this method **does not** take `params` into account.
+     *
+     * @param \BEdita\Core\Model\Entity\DateRange $dateRange Date Range being compared.
+     * @return bool
+     */
+    public function isBefore(DateRange $dateRange)
+    {
+        return $this->start_date < $dateRange->start_date && ($this->end_date === null || $this->end_date <= $dateRange->start_date);
```
didoda: What happens if `$dateRange->start_date` is null or entire object `$dateRange` is null? Maybe these cases should be handled too (if not handled elsewhere)

plugins/BEdita/Core/src/Model/Entity/DateRange.php
```diff
@@ -44,4 +44,191 @@ class DateRange extends Entity
...
+     * for this purpose it assumes the same value as `start_date`.
+     *
+     * **Warning**: this method **does not** take `params` into account.
+     *
+     * @param \BEdita\Core\Model\Entity\DateRange $dateRange Date Range being compared.
+     * @return bool
+     */
+    public function isAfter(DateRange $dateRange)
+    {
+        return $this->start_date > $dateRange->start_date && ($dateRange->end_date === null || $this->start_date >= $dateRange->end_date);
```
didoda: Maybe `$dateRange` null and `$dateRange->start_date` null should be handled

plugins/BEdita/Core/src/Model/Entity/DateRange.php
```diff
@@ -44,4 +44,191 @@ class DateRange extends Entity
...
+     *
+     * @param \BEdita\Core\Model\Entity\DateRange[] $dateRanges Set of Date Ranges.
+     * @return \BEdita\Core\Model\Entity\DateRange[]
+     */
+    public static function normalize(array $dateRanges)
+    {
+        if (empty($dateRanges)) {
+            return [];
+        }
+
```
didoda: `$dateRanges` array is expected to be a "not null" date ranges set. Better checking elements before sorting... otherwise calls like `$dateRange->isBefore` or `$dateRange->isAfter` could have unexpected results. Decide whether getting rid of bad data and continue, or throwing an exception or logging.

plugins/BEdita/Core/tests/TestCase/Model/Entity/DateRangeTest.php
```diff
@@ -0,0 +1,652 @@
...
+                [
+                    'start_date' => '2017-01-05',
+                    'end_date' => null,
+                ],
+                [
+                    'start_date' => '2017-01-05',
+                    'end_date' => null,
+                ],
+            ],
+        ];
```
didoda, line 172: sometimes input data can be bad; maybe some "wrong data" tests could be useful. For instance:
```
[ // wrong interval; end_date before start_date
    'start_date' => '2017-01-05',
    'end_date' => '2017-01-04'
],
[ // unhandled or wrong data format
    'start_date' => '01 2017 31', // odd format
    'end_date' => '2017 02 0' // truncated
],
[ // wrong characters for date
    'start_date' => 'p2*0170x0ab',
    'end_date' => '2017 01 20'
],
[ // wrong interval: start_date null not possible (or is it possible?)
    'start_date' => null,
    'end_date' => '2017 01 20'
],
[ // both start_date and end_date null (uh, what then?)
    'start_date' => null,
    'end_date' => null
],
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/DateRangeTest.php
```diff
@@ -0,0 +1,652 @@
...
```
didoda, line 314: extra tests (like in previous comment)?

plugins/BEdita/Core/tests/TestCase/Model/Entity/DateRangeTest.php
```diff
@@ -0,0 +1,652 @@
...
+                    [
+                        'start_date' => '2017-01-20',
+                        'end_date' => null,
+                    ],
+                    [
+                        'start_date' => '2017-02-07',
+                        'end_date' => '2017-02-10',
+                    ],
+                ],
+            ],
```
didoda, line 425: extra tests including malformed data (as in previous comments)? it can be tricky, but useful to understand what happens when data ranges are not well formed (error? data is ignored? etc.)

didoda (commented):
Everything seems fine. Minor changes are suggested, including null data ranges handling in `DataRange` entity and some extra tests for bad data in `DataRangeTest`.
Good tests ensure us that everything is fine like it seems to be ^_^

fquffio:
@didoda Good points. I have added a `DateRange::checkWellFormed(... $dateRanges)` static method that ensures all of the arguments it receives are "valid" date ranges. I have added tests for this method, but I left other tests untouched, as other methods simply call `static::checkWellFormed()` to ensure everything is alright.

plugins/BEdita/Core/tests/TestCase/Model/Entity/DateRangeTest.php
fquffio: Tests with invalid inputs have been added for newly introduced `DateRange::checkWellFormed()` method, thus they should now be redundant here. 🤔  Do you think it would be better to add them anyway?

didoda: Well... if we expect "blocking exception", I don't think adding redundant tests invalid inputs would be necessary. If we expect that bad data are just ignored, than invalid inputs could be useful.

didoda: `checkWellFormed` is blocking... then no redundancy needed.

Issue #1141, opened 2017-03-14
Error 500 on POST object without auth


stefanorosanelli:
Trying to create an object with a POST without `Authorization` header we get:

 * with `debug` on - an error 500 response with raw **An Internal Server Error Occurred** text and "text/html" content type (no JSON), while in `error.log` we have 
```
2017-03-14 10:11:03 Error: [RuntimeException] You cannot call all() on a non-select query. Use execute() instead.
```
 
* with `debug` off - an error 500 response like this 
```json
{
   "error": {
     "status":"500",
      "title":"An Internal Error Has Occurred."
  }
}
```
while in error log we have 
```
2017-03-14 10:14:47 Error: [PDOException] SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'created_by' cannot be null
```

There's something wrong with our exception renderer
Also response in this case should be 401 not 500

Are these cases not covered by our test suite?

PR #1167, opened 2017-03-27
Refactor tests for use API with auth

batopa:
This PR is preparatory for #1141.

After this PR we can easily setup `Authorization` header in integration test for a specific user. Furthermore it adds the helpers methods that were in the removed class `ApiIntegrationTestCase` to easily build common headers for JSON API requests.

For example, we could perform a POST doing

```php
// using default user (the first one in `UsersFixture`) 
$this->post('/documents', $jsonData);

// using another user
$this->configRequestHeaders('POST', $this->getUserAuthHeader('username', 'password'));
$this->post('/documents', $jsonData);
```

You can also get the JWT tokens if needed:

```php
// for default user (the first one in `UsersFixture`) 
$tokens = $this->authUser();

// for another user
$tokens = $this->authUser('username', 'password');
```

----

Other things done in this PR are:
* fix on `CorsMiddleware` that catched every exceptions also on next middleware. Now correctly it catches just preflight exceptions
* simplify the stack trace with `debug = 1` to be more readable and avoid errors serializing it.


plugins/BEdita/Core/tests/TestCase/Model/Behavior/UniqueNameBehaviorTest.php
```diff
@@ -49,6 +49,18 @@ class UniqueNameBehaviorTest extends TestCase
     public function setUp()
     {
         parent::setUp();
+
```
stefanorosanelli, line 52: Probably a `ModelTestCase` base class with `setUp` and `tearDown` and `LoggedUser` methods could be handy, to avoid repetitions and potential problems on model tests. 

Not necessary now, could be a in a next refactor...

stefanorosanelli (approved):

PR #1172, opened 2017-03-29
Feature/v4/query filter trait

stefanorosanelli:
This PR adds operators to current field filter in `ListEntitiesAction`  via a new `QueryFilterTrait`

this enables us to build query string filters like:
```
?filter[field][gt]=somevalue
?filter[field][]=a&filter[field][]=b
?filter[field1][le]=a&filter[field2][ge]=b
```

### DIFF
--- a/plugins/BEdita/Core/src/ORM/QueryFilterTrait.php
+++ b/plugins/BEdita/Core/src/ORM/QueryFilterTrait.php
@@ -25,7 +25,7 @@ trait QueryFilterTrait
 {
     /**
      * Create query filter using various operators on fields
-     * Options arrya must contain fields as keys and operators like
+     * Options array must contain fields as keys and operators like
      *   - 'gt' or '>' (greather than)
      *   - 'lt' or '<' (less than),
      *   - 'ge' or '>=' (greater or equal)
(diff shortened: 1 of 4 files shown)


plugins/BEdita/Core/src/ORM/QueryFilterTrait.php
```diff
@@ -0,0 +1,122 @@
...
+                    continue;
+                }
+
+                if (!is_array($conditions)) {
+                    $exp = $exp->eq($field, $conditions);
+
+                    continue;
+                }
+
+                $in = [];
```
fquffio, line 72: I'm not sure whether we actually want `IN` conditions to be mixed with other operators… :/ `$filter = [1, 2, 3, 'gte' => 2]` doesn't make much sense to me.

I'd probably go with a more restrictive condition like:
```php
if (Hash::numeric(array_keys($conditions))) {
    $exp = $exp->in($field, $conditions);

    continue;
}
```

batopa: Agree, but you can also make `$filter = ['=' => 1, 'gt' => 2]`or `$filter = ['eq' => 1, 'neq' => 1]` that doesn't make sense :)

PR #1173, opened 2017-03-29
Relationship issues

### LINKED ISSUE #1168: Retrieving relationships for trashed entity
Deleting an entity which is related with other objects should remove (or hide?) its relationships.

### LINKED ISSUE #1169: List relations between certain types gives a 500 error
Creating a relation between `users` and `profiles` gives a 500 error trying to access related objects from URL like `/profiles/{id}/{relation_name}`

Instead a relation between `users` and `documents` seems to work fine.

fquffio:
This PR resolves #1169.

`ListRelatedObjectAction` has been heavily refactored and it now uses a completely different strategy to list related objects. As a result of this refactor, `ListObjectsAction` is now used under the hood. This means that related objects can be filtered using `?filter` query param, and deleted objects are now properly excluded from such lists (this closes #1168).

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
@@ -13,7 +13,13 @@
 
 namespace BEdita\Core\Model\Action;
 
+use BEdita\Core\ORM\Association\RelatedTo;
+use Cake\Datasource\EntityInterface;
+use Cake\Datasource\Exception\RecordNotFoundException;
 use Cake\ORM\Query;
+use Cake\ORM\ResultSet;
+use Cake\ORM\TableRegistry;
+use Cake\Utility\Hash;
 
 /**
  * Command to list associated objects.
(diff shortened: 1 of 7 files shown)


plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
```diff
@@ -29,16 +35,86 @@ class ListRelatedObjectsAction extends ListAssociatedAction
     public function execute(array $data = [])
     {
         $data += ['joinData' => true];
-        $result = parent::execute($data);
-        if (!($result instanceof Query)) {
-            return $result;
+
```
stefanorosanelli, line 38: This method is quite long, maybe split it in shorter methods?
Also in a later/future refactor, not necessarily now.

stefanorosanelli:
An integration test with this scenario is highly recommended, like: list `locations` in a custom relation to `users`. Probably a change in fixtures is needed. 
Could be done in a later PR.

plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
fquffio: It is long indeed. :( However, should we approve this kind of approach when listing relationships, we might decide to refactor `ListAssociatedEntitiesAction` in the same fashion, and some logic should become reusable.

I'd plan this other refactor ASAP.

PR #1174, opened 2017-03-29
Refactor ListAssociatedAction

fquffio:
This PR introduces a refactor similar to the one merged in #1173.

The approach we used to address issues in relations between BEdita objects can be used for associations in general. This PR aims to share most of the logic between `ListAssociatedAction` and `ListRelatedObjectsAction`.

This PR purposely doesn't include any test, since this refactor is supposed to be "internal", thus there should be no visible changes to the external world. However, this PR should simplify things a little, and should make this action slightly more reliable. 🤔


fquffio:
(Ok, _maybe_ a couple of extra tests would be better… 😂 )

PR #1175, opened 2017-03-30
Fix error 500 on post/patch/delete without authenticated user

### LINKED ISSUE #1141: Error 500 on POST object without auth

batopa:
This PR resolves #1141

### DIFF
--- a/plugins/BEdita/API/src/Event/CommonEventHandler.php
+++ b/plugins/BEdita/API/src/Event/CommonEventHandler.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2017 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+namespace BEdita\API\Event;
+
+use BEdita\API\Middleware\CorsMiddleware;
+use BEdita\Core\Utility\LoggedUser;
+use Cake\Core\Configure;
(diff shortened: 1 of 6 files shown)


plugins/BEdita/API/src/Event/CommonEventHandler.php
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param \Cake\Event\Event $event The event object
+     * @param \Cake\Http\MiddlewareQueue $middleware The middleware queue
+     * @return void
+     * @see \BEdita\API\Middleware\CorsMiddleware to more info on CORS configuration
+     */
+    public function buildMiddlewareStack(Event $event, MiddlewareQueue $middleware)
+    {
+        $middleware->insertAfter(
+            'Cake\Error\Middleware\ErrorHandlerMiddleware',
```
fquffio, line 67: As a rule of thumb, I generally prefer `ErrorHandlerMiddleware::class` instead of a string containing the FQNS.

It yields the very same result, so it makes no difference, but it generally makes it easier to spot usages of a particular class in IDEs and code analysis tools. Should ever `ErrorHandlerMiddleware` class become deprecated, for instance, it would make it easier for us to remove the deprecated component.

fquffio (approved):

stefanorosanelli (approved):

PR #1188, opened 2017-04-06
Meta fields

### LINKED ISSUE #1180: `attributes` end `meta` response review
Some core fields currently in `attributes` section of a response should appear instead in `meta`: they are not modifiable via API  calls:

- `"created"`
- `"modified"`
- `"published"`
- `"created_by"`
- `"modified_by"`
- `"locked"` (???)

fquffio:
This PR resolves #1180.

By default, all non-accessible properties (i.e. properties that cannot be modified by a `PATCH` request) are moved from `attributes` to `meta`. Join data, if present, are then merged with the resulting hash. This behavior is implemented in `JsonApiTrait`, so entity classes can override the default if needed.

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
@@ -65,19 +65,21 @@ class ObjectsControllerTest extends IntegrationTestCase
                     'attributes' => [
                         'status' => 'on',
                         'uname' => 'first-user',
-                        'locked' => true,
-                        'created' => '2016-05-13T07:09:23+00:00',
-                        'modified' => '2016-05-13T07:09:23+00:00',
-                        'published' => null,
                         'title' => 'Mr. First User',
                         'description' => null,
                         'body' => null,
                         'extra' => null,
                         'lang' => 'eng',
-                        'created_by' => 1,
-                        'modified_by' => 1,
                         'publish_start' => null,
                         'publish_end' => null,
+                    ],
+                    'meta' => [
+                        'locked' => true,
+                        'created' => '2016-05-13T07:09:23+00:00',
+                        'modified' => '2016-05-13T07:09:23+00:00',
+                        'published' => null,
+                        'created_by' => 1,
+                        'modified_by' => 1,
                     ],
                     'links' => [
                         'self' => 'http://api.example.com/users/1',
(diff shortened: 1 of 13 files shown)


plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
```diff
@@ -65,20 +65,22 @@ public function testIndex()
...
                         'body' => null,
                         'extra' => null,
                         'lang' => 'eng',
-                        'created_by' => 1,
-                        'modified_by' => 1,
                         'publish_start' => null,
                         'publish_end' => null,
                     ],
+                    'meta' => [
+                        'locked' => true,
```
stefanorosanelli, line 77: `objects.locked` and `roles.unchangeable` are special fields: normally not changeable, but a user with `administrator` role should probably be able to change those fields.
How do we deal with that?
How are those properties set?

stefanorosanelli (approved):

plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
fquffio: 🤔 Yeah, we should probably leave those fields in `attributes`.

@batopa @edoardocavazza ?

batopa: Yes, I think so. Maybe some fields should be not visible in `attributes` unless you are `administrator` and if you aren't an admin you can just see it in `meta` (or not even in `meta` if is confusing)

stefanorosanelli: Both fields tell you that a resource is not modifiable. Even if your not an `administrator` is probably a good thing to know.
That said, don't know where to show them.
Having a different layout for different roles could be confusing too...

stefanorosanelli: Other solution: keep `objects.locked` and `roles.unchangeable` in `meta` - and implement special API actions to change them, available to `administrator`only

edoardocavazza: I like the solution to keep those fields in the `meta` field. An administrator can always change those properties with a custom hidden set like we already do with `password_hash`

PR #1192, opened 2017-04-10
Handle dependencies in plugins under the plugins folder

### LINKED ISSUE #1189: Vendor dependency on a plugin
If I need to use a vendor library on a BE4 plugin what should I do:
 * change plugin's `composer.json` ?
 * change main `composer.json' ? even though it'is on bedita/bedita repo?
 * should I run `composer update`? from where?

batopa:
This PR resolves #1189 using https://github.com/wikimedia/composer-merge-plugin

Every plugin put in `plugins` folder and that contains `composer.json` will be hooked to app `composer.json` loading dependencies and autoloading the namespaces.

In addition an unversioned `composer.local.json` may be added to main app to customize the app requirements.

I had issues with `"robmorgan/phinx": "dev-fix\/sqlite-update-column-commas as 0.6.5"` so I left it in app `composer.json`. We'll going to remove it once `cakephp/migrations` dependencies will be updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 #################################################
 /.vagrant/*
 /composer.lock
+/composer.local.json
 /config/app.php
 /coverage/*
 /docs/*
--- a/composer.json
+++ b/composer.json
@@ -33,28 +33,23 @@
         "cakephp/migrations": "~1.0",
         "robmorgan/phinx": "dev-fix\/sqlite-update-column-commas as 0.6.5",
         "cakephp/plugin-installer": "*",
-        "firebase/php-jwt": "~4.0"
+        "wikimedia/composer-merge-plugin": "^1.4"
     },
     "require-dev": {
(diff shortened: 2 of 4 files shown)


fquffio (approved):
Seems alright to me.

I just ran into a weird behavior, however: on the first run of `composer update` after checking out this branch, the update removed some required dependencies, then gave an error without failing. I guess this is caused by the fact that the `wikimedia/composer-merge-plugin` was being installed during that update, so Composer messed things up. Another `composer update` did the work just fine.

I also noticed that when running `composer install` on a fresh install (`rm -rf composer.lock vendor/*`, then `composer install`) runs the `post-install-cmd` script twice, but I couldn't figure out _why_ this is happening. On `4-develop` that script is run just once as expected.

stefanorosanelli:
Same behavior noted on first run of `composer update` after branch swtich....

stefanorosanelli (approved):

Issue #1195, opened 2017-04-13
[API] Pagination size configuration


stefanorosanelli:
As explained here 
http://bedita.readthedocs.io/en/4-develop/endpoints/parameters.html#pagination

we need seom default settings for `default` and `max` page_size
probably an hardcoded `max` of `max` limit is necessary to avoid dangerous configurations

stefanorosanelli:
Config layout proposal:
```php
    'Pagination' => [
        'defaultSize' => 20,
        'maxSize' => 100,
    ],
```

with max possible `maxSize` 500, hardcoded...

fquffio:
CakePHP's [Paginator Component](https://api.cakephp.org/3.4/class-Cake.Controller.Component.PaginatorComponent.html#$_defaultConfig) uses `limit` for default page size and `maxLimit` for maximum requested page size.

didoda:
From `PaginatorComponent` in `vendor/cakephp/cakephp/src/Controller/Component`:
```
protected $_defaultConfig = [
        'page' => 1,
        'limit' => 20,
        'maxLimit' => 100,
        'whitelist' => ['limit', 'sort', 'page', 'direction']
    ];
```
`limit` and `maxLimit` seem fine

stefanorosanelli:
Yes, that's what Cake use but we use `page` and `page_size` in query string and `meta` response. I would use that as reference. 
However, it's a matter of tase :wink:

PR #1196, opened 2017-04-14
 #1195 feat: pagination in config

didoda:
This provides pagination configurable

### DIFF
--- a/plugins/BEdita/API/src/Controller/AppController.php
+++ b/plugins/BEdita/API/src/Controller/AppController.php
@@ -55,11 +55,7 @@ class AppController extends Controller
 
         $this->response = $this->response->withHeader('X-BEdita-Version', Configure::read('BEdita.version'));
 
-        $pagination = Configure::read('Pagination');
-        if (empty($pagination)) {
-            $pagination = [];
-        }
-        $this->loadComponent('BEdita/API.Paginator', $pagination);
+        $this->loadComponent('BEdita/API.Paginator', (array)Configure::read('Pagination'));
 
         $this->loadComponent('RequestHandler');
         if ($this->request->is(['json', 'jsonapi'])) {
--- a/plugins/BEdita/API/src/Controller/AppController.php
+++ b/plugins/BEdita/API/src/Controller/AppController.php
@@ -55,7 +55,11 @@ class AppController extends Controller
 
         $this->response = $this->response->withHeader('X-BEdita-Version', Configure::read('BEdita.version'));
 
-        $this->loadComponent('BEdita/API.Paginator', Configure::read('Pagination'));
+        $pagination = Configure::read('Pagination');
+        if (empty($pagination)) {
+            $pagination = [];
+        }
+        $this->loadComponent('BEdita/API.Paginator', $pagination);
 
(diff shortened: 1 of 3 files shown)


plugins/BEdita/API/src/Controller/AppController.php
```diff
@@ -55,7 +55,7 @@ public function initialize()
 
         $this->response = $this->response->withHeader('X-BEdita-Version', Configure::read('BEdita.version'));
 
-        $this->loadComponent('BEdita/API.Paginator');
+        $this->loadComponent('BEdita/API.Paginator', Configure::read('Pagination'));
```
stefanorosanelli: If no `Pagination` key is found in config an empty array must be passed

(see travis errors)

Also a sample setting in `app.default.php` is missing

plugins/BEdita/API/src/Controller/Component/PaginatorComponent.php
```diff
@@ -34,12 +34,15 @@ class PaginatorComponent extends CakePaginatorComponent
         'whitelist' => ['page', 'page_size', 'sort'],
     ];
 
+    public $absoluteLimit = 500;
```
stefanorosanelli: I would use `protected` to avoid possible easy changes....

stefanorosanelli (changes requested):

stefanorosanelli (approved):

plugins/BEdita/API/src/Controller/Component/PaginatorComponent.php
```diff
@@ -34,12 +34,15 @@ class PaginatorComponent extends CakePaginatorComponent
         'whitelist' => ['page', 'page_size', 'sort'],
     ];
 
+    protected $absoluteLimit = 500;
```
fquffio: Docblock is missing for this class member. 🙃 

However, I'd probably use `const MAX_LIMIT = 500;` instead.

plugins/BEdita/API/src/Controller/AppController.php
```diff
@@ -55,7 +55,11 @@ public function initialize()
 
         $this->response = $this->response->withHeader('X-BEdita-Version', Configure::read('BEdita.version'));
 
-        $this->loadComponent('BEdita/API.Paginator');
+        $pagination = Configure::read('Pagination');
+        if (empty($pagination)) {
+            $pagination = [];
+        }
+        $this->loadComponent('BEdita/API.Paginator', $pagination);
```
fquffio: (This is a _veeeeery minor_ issue…)

In order to reduce complexity (which makes Scrutinizer complain at times) we could avoid the `if` statement by just writing:
```php
$this->loadComponent('BEdita/API.Paginator', (array)Configure::read('Pagination'));
```

If that configuration key is missing, `Configure::read()` returns `null`, and casting NULL to an array results in an empty array (`(array)null === []`), which is perfect. 😎

fquffio (approved):

plugins/BEdita/API/src/Controller/AppController.php
didoda: :+1:

PR #1198, opened 2017-04-21
 #1184 feat: http code 204 and no content-type

### LINKED ISSUE #1184: Delete fails if content type header is missing
Trying to delete an object through an API call with `DELETE` method to the url `/{type}/{id}`, I get a fail response with 500 code status and message: `Invalid header value; must be a string or array of strings`. Following the trace, I found out that it refers to the `Content-type` header.

didoda:
This fixes #1184, removing `content-type` from header, when http status code returned is 204.

### DIFF
--- a/plugins/BEdita/API/src/Controller/TrashController.php
+++ b/plugins/BEdita/API/src/Controller/TrashController.php
@@ -104,7 +104,6 @@ class TrashController extends AppController
         }
 
         return $this->response
-            ->withHeader('Content-Type', $this->request->contentType())
             ->withStatus(204);
     }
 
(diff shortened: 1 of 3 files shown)


plugins/BEdita/API/src/Controller/TrashController.php
```diff
@@ -137,7 +136,7 @@ public function delete($id)
         }
 
         return $this->response
-            ->withHeader('Content-Type', $this->request->contentType())
+            //->withHeader('Content-Type', $this->request->contentType())
```
fquffio: I'd delete this line as well, instead of leaving it commented out.

fquffio (approved):
👍  Having `Content-Type` header set in a `204 No Content` response was a little weird, actually. 😂

PR #1199, opened 2017-04-24
Embed related entities in response

### LINKED ISSUE #1194: Embed related entities in response
In order to leverage HTTP traffic when a client wants to retrieve associated resources, we should implement `?include` query parameter as per [JSON API](http://jsonapi.org/format/#fetching-includes) specification.

fquffio:
This PR implements `?include=` query parameter and resolves #1194, at the cost of a huge refactor of JSON API serialization logic. 😱  A (probably minor) caveat is that nested resources (e.g.: `?include=roles.users`) are not yet supported.

The `?include` query parameter can be used in `GET /:controller` and `GET /:controller/:id` requests, but **not** in `GET /:controller/:id/:relationship` requests, because in this case it would be difficult to decide which relationships are accessible, thus can be `include`d.

The refactor on JSON API serialization logic consists of a `JsonApiSerializable` interface that most entity classes implement. This interface exposes a `jsonApiSerialize($options = 0)` public method that returns a [resource object](http://jsonapi.org/format/#document-resource-objects). `JsonApi` class now relies on this interface to serialize entities instead of its in-house logic, that had to be flexible enough to handle all possible cases, and was quickly becoming very complex.

This refactor became necessary because `JsonApi` class used to rely on a `$type` argument, that was set by the controller through a "magic" view variable (`_type`). This argument was necessary in order to build URLs, display list of relationships, and stuff like this. For included resources, it would have been difficult for the controller to know in advance all possible `_type`s and to pass 'em to the `JsonApi` utility class in a somehow elegant way, therefore we needed some more information at the entity level — each entity needs to know something about how it will be serialized in JSON API format. The `_type` view variable is now unnecessary, and has been removed. 🙃 

I'm not sure at all this is the right way to go. There are surely better approaches, but this is the one I could come up with. Please feel free to give criticisms and helpful suggestions @batopa @didoda @stefanorosanelli !


stefanorosanelli (approved):
Code seems ok.
We should probably find a way to be more abstract and less JSON-API centric implementing patterns like Adapter.

Included relations are not paginated, but they should be.
Probably a limit threshold is sufficient -> new issue for that

batopa:
Maybe the only issue removing the check on primary in `Table::inheritanceBeforeFind()` is that some fix could be executed more times, but we could skip it setting a flag in `\ArrayObject $options` the first time the fixes are executed.

batopa (approved):

PR #1210, opened 2017-04-28
Asynchronous jobs model

### LINKED ISSUE #1201: Async jobs table and model
Create `jobs` table and model for generic asynchronous jobs (mail, notifications, signup,...)
This table could have its own datasource and different table name.

fquffio:
This PR resolves #1201.

Asynchronous jobs are stored in table `async_jobs`. If a datasource connection `async_jobs` is configured, it is assumed that asynchronous jobs are to be managed via that connection. This allows users to use a different storage for application data and asynchronous jobs. However, at this time users need to manually create schema for `async_jobs` table into that datasource.

Asynchronous jobs are identified via their `uuid`, and have a `type` (that is necessary to know which action should be taken in order to run that job), a JSON `payload` to provide additional context, and a `priority` (jobs with higher priority must be run before jobs with lower priorities). They can be scheduled setting `not_before` and/or `not_after`. One can also choose how many times a job can be retried before being marked as failed by setting the `max_attempts` property — by default, job execution will be attempted only once.

Metadata includes `created` and `modified`, and a `completed` field that stores the timestamp at which that job was marked as successfully run.

A `status` property can be read from the entity to easily tell the status of the job. Can be one of **completed** (job already run successfully), **planned** (job to be run in the future, because `not_before > NOW()`), **failed** (either job has expired because `not_after < NOW()`, or `max_attempts = 0`), **locked** (some runner is currently managing that job) or **pending** (job can executed right now, but nobody holds a lock on it).

### Processing jobs
When a job is being run, it **MUST** be locked:
```php
try {
    $asyncJob = TableRegistry::get('AsyncJobs')->lock($uuid, '+10 minutes');
    // The second parameter is optional, and it represents the period of validity for the lock.
    // After that amount of time has expired, it is assumed that the runner is dead, and the lock is
    // released, so that other runners can take care of that job. Default is `+5 minutes`.
} catch (\Exception $e) {
    // The job could not be locked. Either it has expired, it cannot be retried any more,
    // or somebody else already locked it in the meanwhile.
}
```

Right after job execution has ended, whether it was successful or it failed, the lock **MUST** be released:
```php
TableRegistry::get('AsyncJobs')->unlock($asyncJob->uuid, $success);
```

If `$success === false`, job execution failed and the lock is released without further action. Otherwise, the job is marked as "completed".

### Finding jobs
Three custom finders have been implemented to easily retrieve jobs by their status:
```php
$pending = TableRegistry::get('AsyncJobs')->find('pending');
$failed = TableRegistry::get('AsyncJobs')->find('failed');
$completed = TableRegistry::get('AsyncJobs')->find('completed');
```

### Events
Three events are dispatched, and can be listened for (maybe for logging purposes?):
- **AsyncJob.lock**: this event is dispatched when a job is being locked. Event data includes the asynchronous job (`entity`) and the timestamp at which the lock will expire (`expires`).
- **AsyncJob.fail**: this event is dispatched when a lock is released and `$success === false`. Event data includes the asynchronous job (`entity`) and success information (`success`). **Warning**: this event _is not_ triggered for timed-out locks!
- **AsyncJob.complete**: this event is dispatched when a lock is released and `$success !== false`. Event data includes the asynchronous job (`entity`) and success information (`success`).


plugins/BEdita/Core/config/Migrations/20170428100034_AddAsyncJobsTable.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    public function up()
+    {
+
+        $this->table('async_jobs', ['id' => false, 'primary_key' => ['uuid']])
+            ->addColumn('uuid', 'uuid', [
+                'limit' => null,
```
stefanorosanelli, line 18: comments on columns are missing

plugins/BEdita/Core/src/Model/Entity/AsyncJob.php
```diff
@@ -0,0 +1,83 @@
...
+    protected $_virtual = [
+        'status',
+    ];
+
+    /**
+     * Magic getter for status.
+     *
+     * @return string
+     */
+    protected function _getStatus()
```
stefanorosanelli, line 53: This solution with `status` virtual field and a combined of fields to determine actual status works, but having a real `status` field has other advantages:
 * more database *readable* 
 * easier integration with other systems / languages / apis - don't rely on our finders logic

Downside: 
* more logic on `lock` and `unlock` actions 
 * some duplicate information in columns, mostly status details

stefanorosanelli: @batopa whato do you think?

plugins/BEdita/Core/src/Model/Table/AsyncJobsTable.php
```diff
@@ -0,0 +1,270 @@
...
+     *
+     * @param string $uuid UUID of job to be locked.
+     * @param mixed $duration Duration. By default, jobs are locked for 5 minutes.
+     * @return \BEdita\Core\Model\Entity\AsyncJob
+     */
+    public function lock($uuid, $duration = '+5 minutes')
+    {
+        return $this->getConnection()->transactional(function () use ($uuid, $duration) {
+            $entity = $this->get($uuid, ['finder' => 'pending']);
+            $entity->max_attempts -= 1;
```
stefanorosanelli, line 138: do we have to check that `$entity->max_attempts > 0`? 
if not lock should fail

fquffio: `'finder' => 'pending'` already adds the condition `WHERE ... AND max_attempts > 0`, so it shouldn't be necessary.

plugins/BEdita/Core/config/Migrations/20170428100034_AddAsyncJobsTable.php
```diff
@@ -0,0 +1,139 @@
...
+                'default' => null,
+                'limit' => null,
+                'null' => true,
+            ])
+            ->addColumn('not_before', 'timestamp', [
+                'default' => null,
+                'limit' => null,
+                'null' => true,
+            ])
+            ->addColumn('not_after', 'timestamp', [
```
stefanorosanelli: i'd prefer `expires`

plugins/BEdita/Core/config/Migrations/20170428100034_AddAsyncJobsTable.php
fquffio: Right, I'm fixing it.

plugins/BEdita/Core/src/Model/Entity/AsyncJob.php
batopa: Yes, I would prefer a `status` field, it's more clear/readable and it avoids intricated custom finder.

fquffio: Custom finder is required anyways: if a job's `status` is _pending_, it might have become invalid in the meanwhile, if it's _locked_, the lock might have expired, if it's _planned_, it might have become valid, …

plugins/BEdita/Core/config/Migrations/20170428100034_AddAsyncJobsTable.php
fquffio: Fix'd.

plugins/BEdita/Core/src/Model/Entity/AsyncJob.php
fquffio: In other words: we can only trust the `status` column if it's _failed_ or _completed_. All other statuses (_pending_, _locked_, _planned_) might be outdated.

PR #1211, opened 2017-05-02
Enable setup with Sqlite and Postgres

stefanorosanelli:
With this PR it will enable Sqlite and Postgres in `bedita setup` shell script.
User will have to input one of: Mysql (default), Postgres or Sqlite

In case of Sqlite only database file path is requested, with default path `tmp/be4.sqlite` 

With Sqlite `db_admin check_schema` fails, so a second run of  `bedita setup` is not possible.
To fix in other issue after release.

### DIFF
--- a/plugins/BEdita/Core/src/Shell/BeditaShell.php
+++ b/plugins/BEdita/Core/src/Shell/BeditaShell.php
@@ -359,7 +359,8 @@ class BeditaShell extends Shell
             return;
         }
         $this->userInputData['host'] = $this->in('Host?', null, 'localhost');
-        $this->userInputData['port'] = $this->in('Port?', null, '3306');
+        $defaultPort = ($this->userInputData['driver'] === 'Mysql') ? '3306' : '5432';
+        $this->userInputData['port'] = $this->in('Port?', null, $defaultPort);
         $this->userInputData['database'] = $this->in('Database?');
         $this->userInputData['username'] = $this->in('Username?');
         $this->userInputData['password'] = $this->in('Password?');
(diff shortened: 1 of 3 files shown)


fquffio (approved):

plugins/BEdita/Core/src/Shell/BeditaShell.php
```diff
@@ -349,7 +351,13 @@ protected function saveConnectionData()
...
-        $this->userInputData = [];
+        $this->userInputData = ['host' => 'localhost', 'port' => '', 'database' => '', 'username' => '', 'password' => ''];
+        $this->userInputData['driver'] = $this->in('Driver?', ['Mysql', 'Postgres', 'Sqlite'], 'Mysql');
+        if ($this->userInputData['driver'] === 'Sqlite') {
+            $this->userInputData['database'] = $this->in('Sqlite database file path?', null, TMP . 'be4.sqlite');
+
+            return;
+        }
         $this->userInputData['host'] = $this->in('Host?', null, 'localhost');
         $this->userInputData['port'] = $this->in('Port?', null, '3306');
```
didoda: Default port suggested should not be always 3306 (mysql default port). I.e. postgres default port is 5432.

stefanorosanelli: Right, i'll fix it

batopa (approved):

plugins/BEdita/Core/src/Shell/BeditaShell.php
stefanorosanelli: should be ok now

Issue #1203, opened 2017-04-26
Signup endpoint for user registration


stefanorosanelli:
A new `/signup` endpoint is needed to:

* allow users registration/signup
* `POST /signup` will be called anonymously
* `data.attributes` requires `username`, `password` and `email`
* in `meta` key of signup post an `activation_url` or `redirect_url` can be specified, but both could be set via configuration
* `/signup?code={code}` will be responsible for registration confirmation giving also an HTML browser response

fquffio:
The flow should now be complete. @batopa Can we close this issue?

batopa:
Yes, if necessary we'll work on refinements/improvements in specific issues.

PR #1212, opened 2017-05-03
Whitelist for endpoint authorized

batopa:
This PR is propedeutics for #1203 

It introduces a configurable whitelist of endpoints that don't require permission check to be authorized.
An endpoint whitelisted is always authorized unless a specific permission is set on that endpoint for the current application.

In this way we can setup `signup` in whitelist and allow the request `POST /signup` for anonymous users.

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Auth/EndpointAuthorizeTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Auth/EndpointAuthorizeTest.php
@@ -14,11 +14,13 @@
 namespace BEdita\API\Test\TestCase\Auth;
 
 use BEdita\API\Auth\EndpointAuthorize;
+use BEdita\Core\Model\Entity\Endpoint;
 use BEdita\Core\State\CurrentApplication;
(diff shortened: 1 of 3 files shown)


plugins/BEdita/API/tests/TestCase/Auth/EndpointAuthorizeTest.php
```diff
@@ -134,20 +136,40 @@ public function getEndpointProvider()
...
+                new Uri('/disabled/endpoint'),
+            ]
         ];
     }
 
     /**
      * Test getting endpoint from request.
      *
-     * @param int|\Exception $expected Expected endpoint ID.
+     * @param int|\Cake\ORM\Entity $expected Expected endpoint ID or entity.
```
fquffio: To be (_very_) pedantic, this parameter's type is `int|\Cake\ORM\Entity|\Exception`. Or `mixed` to be more concise. 🙃

fquffio (approved):

Issue #1202, opened 2017-04-26
Async jobs business logic and shell script


stefanorosanelli:
Implement shell script to run async jobs, something like:
```bash
bin/cake jobs run [-limit n] [-all]
```

* async jobs logic should be API callable
* new async jobs may be created in plugins
* core async jobs may be overridden via plugins

PR #1213, opened 2017-05-03
Async job services and shell

stefanorosanelli:
This PR solves #1202 

- `ServiceRunner` utility class to run async jobs was introduced
- Each `'service'` saved in `async_jobs` must be run by a class implementing `JobService` interface containing a generic `run` method
- `Service` classes for a specific `'service'` may be registered using `ServiceRunner::register`, otherwise a class matching service name is searched in `Service` namespace of loaded plugins first, than in `BEdita/Core` - for an `'example'` service first a class `\MyPlugin\Service\Example` is searched and finally  `\BEdita\Core\Service\Example`
- `ServiceRunner` provides methods to run all pending jobs, with an optional limit, or a single job via uuid
- a `JobsShell` class has been introduced to run pending jobs, with an optional limit argument
- a dummy `Mail` service class has been created, its logic will be implemented in a future PR or issue

### DIFF
--- a/plugins/BEdita/Core/src/Job/ServiceRunner.php
+++ b/plugins/BEdita/Core/src/Job/ServiceRunner.php
@@ -61,7 +61,6 @@ class ServiceRunner
             }
         }
         if (!$classFound) {
-            Log::write('error', 'service not found: ' . $name);
             throw new \LogicException(__d('bedita', 'Unknown service "{0}"', [$name]));
         }
         $instance = new $classFound();
@@ -82,7 +81,6 @@ class ServiceRunner
     public static function register($name, $instance)
     {
         if (!($instance instanceof JobService)) {
-            Log::write('error', 'bad service class: ' . get_class($instance));
             throw new \LogicException(__d('bedita', 'Bad service class "{0}"', [get_class($instance)]));
         }
         static::$instances[$name] = $instance;
--- a/plugins/BEdita/Core/src/Job/ServiceRunner.php
+++ b/plugins/BEdita/Core/src/Job/ServiceRunner.php
@@ -13,6 +13,7 @@
 
 namespace BEdita\Core\Job;
 
+use Cake\Core\App;
 use Cake\Core\Configure;
 use Cake\Log\Log;
 use Cake\ORM\TableRegistry;
@@ -51,11 +52,11 @@ class ServiceRunner
         }
         $className = Inflector::camelize($name);
         $plugins = array_keys(Configure::read('Plugins'));
-        $plugins[] = 'BEdita\Core';
+        $plugins[] = 'BEdita/Core';
         $classFound = null;
         foreach ($plugins as $plugin) {
-            $fullName = '\\' . $plugin . '\Service\\' . $className;
-            if (class_exists($fullName)) {
+            $fullName = App::className("$plugin.$className", 'Service');
+            if ($fullName) {
                 $classFound = $fullName;
                 break;
             }
--- a/plugins/BEdita/Core/src/Job/ServiceRunner.php
+++ b/plugins/BEdita/Core/src/Job/ServiceRunner.php
@@ -51,7 +51,7 @@ class ServiceRunner
             return static::$instances[$name];
         }
         $className = Inflector::camelize($name);
-        $plugins = array_keys(Configure::read('Plugins'));
+        $plugins = array_keys((array)Configure::read('Plugins'));
         $plugins[] = 'BEdita/Core';
         $classFound = null;
         foreach ($plugins as $plugin) {
--- a/plugins/BEdita/Core/src/Job/ServiceRunner.php
+++ b/plugins/BEdita/Core/src/Job/ServiceRunner.php
@@ -15,7 +15,7 @@ namespace BEdita\Core\Job;
 
 use BEdita\Core\Job\JobService;
 use Cake\Core\Configure;
-use Cake\Log\LogTrait;
+use Cake\Log\Log;
 use Cake\ORM\TableRegistry;
 use Cake\Utility\Hash;
 use Cake\Utility\Inflector;
@@ -27,30 +27,12 @@ use Cake\Utility\Inflector;
  */
 class ServiceRunner
 {
-    use LogTrait;
-
-    /**
-     * Async Jobs table.
-     *
-     * @var \BEdita\Core\Model\Table\AsyncJobs
-     */
-    protected $AsyncJobs = null;
-
     /**
      * Registered instances.
      *
      * @var \BEdita\Core\Job\JobService[]
      */
-    protected $instances = [];
-
-    /**
-     * Default constructor
-     * @codeCoverageIgnore
-     */
-    public function __construct()
-    {
-        $this->AsyncJobs = TableRegistry::get('AsyncJobs');
-    }
+    protected static $instances = [];
 
     /**
      * Get a service class instance for a given $name.
@@ -63,10 +45,10 @@ class ServiceRunner
      * @return \BEdita\Core\Job\JobService Service instance found
      * @throws \LogicException
      */
-    public function getService($name)
+    public static function getService($name)
     {
-        if (!empty($this->instances[$name])) {
-            return $this->instances[$name];
+        if (!empty(static::$instances[$name])) {
+            return static::$instances[$name];
         }
         $className = Inflector::camelize($name);
         $plugins = array_keys(Configure::read('Plugins'));
@@ -80,11 +62,11 @@ class ServiceRunner
             }
         }
         if (!$classFound) {
-            $this->log('service not found: ' . $name, 'error');
+            Log::write('error', 'service not found: ' . $name);
             throw new \LogicException(__d('bedita', 'Unknown service'));
         }
         $instance = new $classFound();
-        $this->register($name, $instance);
+        static::register($name, $instance);
 
         return $instance;
     }
@@ -98,13 +80,13 @@ class ServiceRunner
      * @return void
      * @throws \LogicException
      */
-    public function register($name, $instance)
+    public static function register($name, $instance)
     {
         if (!($instance instanceof JobService)) {
-            $this->log('bad service class: ' . get_class($instance), 'error');
+            Log::write('error', 'bad service class: ' . get_class($instance));
             throw new \LogicException(__d('bedita', 'Bad service instance'));
         }
-        $this->instances[$name] = $instance;
+        static::$instances[$name] = $instance;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


plugins/BEdita/Core/src/Job/ServiceRunner.php
```diff
@@ -0,0 +1,166 @@
...
+        $classFound = null;
+        foreach ($plugins as $plugin) {
+            $fullName = '\\' . $plugin . '\Service\\' . $className;
+            if (class_exists($fullName)) {
+                $classFound = $fullName;
+                break;
+            }
+        }
+        if (!$classFound) {
+            Log::write('error', 'service not found: ' . $name);
```
fquffio: Uncaught exceptions are already written to `error.log`, I'd avoid duplicate entries.

plugins/BEdita/Core/src/Job/ServiceRunner.php
```diff
@@ -0,0 +1,166 @@
...
+    {
+        if (!empty(static::$instances[$name])) {
+            return static::$instances[$name];
+        }
+        $className = Inflector::camelize($name);
+        $plugins = array_keys(Configure::read('Plugins'));
+        $plugins[] = 'BEdita\Core';
+        $classFound = null;
+        foreach ($plugins as $plugin) {
+            $fullName = '\\' . $plugin . '\Service\\' . $className;
```
fquffio: Can `Cake\Core\App::className($className, 'Service')` be used instead of this loop? Things would change a little, because standard `Plugin.ClassName` syntax should be used, but I don't think it would be that bad… 🤔

plugins/BEdita/Core/src/Job/ServiceRunner.php
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * @param string $name The service name you want to register.
+     * @param mixed $instance The instance object to be registered, MUST implement JobService
+     * @return void
+     * @throws \LogicException
+     */
+    public static function register($name, $instance)
+    {
+        if (!($instance instanceof JobService)) {
+            Log::write('error', 'bad service class: ' . get_class($instance));
```
fquffio: (see above) Uncaught exception are already written to `error.log`.

plugins/BEdita/Core/src/Job/ServiceRunner.php
```diff
@@ -0,0 +1,166 @@
...
+     *  - 'success' uuids of failed jobs
+     *
+     * @param int $limit Max number of pending jobs to run.
+     * @return array Result details
+     */
+    public static function runPending($limit = 0)
+    {
+        $results = ['success' => [], 'failure' => []];
+        $pending = TableRegistry::get('AsyncJobs')->find('pending')->select(['uuid']);
+        if ($limit) {
```
fquffio: Since the usage scenario of this is mainly an unattended environment, I guess that in some cases it might happen to accidentally pass 0 as the limit, and I'm not sure whether running _all_ pending jobs is desirable in such case, despite being an edge case. I'd probably use `null` as default parameter value and add a strict check here (`if ($limit !== null)`) to avoid such accidents.

plugins/BEdita/Core/src/Shell/JobsShell.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public function run()
+    {
+        $limit = $this->param('limit');
+        if (!$limit) {
+            $this->out('Running all pending jobs...');
+        } else {
+            $this->out(sprintf('Running pending jobs, limit number: %s', $limit));
+        }
+        $result = ServiceRunner::runPending($limit);
```
fquffio, line 65: (To be investigated in the future) What happens if user sends SIGTERM or SIGHUP? Should the script wait for the job being currently run (if any) to complete?

See: [`pcntl_signal()`](http://php.net/manual/en/function.pcntl-signal.php)

plugins/BEdita/Core/src/Job/ServiceRunner.php
```diff
@@ -0,0 +1,166 @@
...
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
+namespace BEdita\Core\Job;
+
+use BEdita\Core\Job\JobService;
```
fquffio: Unused import.

plugins/BEdita/Core/src/Shell/JobsShell.php
```diff
@@ -0,0 +1,76 @@
...
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+namespace BEdita\Core\Shell;
+
+use BEdita\Core\Job\ServiceRunner;
+use Cake\Cache\Cache;
```
fquffio: Unused import.

fquffio (changes requested):

plugins/BEdita/Core/src/Job/ServiceRunner.php
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public static function runPending($limit = 0)
+    {
+        $results = ['success' => [], 'failure' => []];
+        $pending = TableRegistry::get('AsyncJobs')->find('pending')->select(['uuid']);
+        if ($limit) {
+            $pending->limit($limit);
+        }
+        $count = 0;
+        foreach ($pending as $job) {
```
fquffio: (To be investigated in the future) PDO allows query results to be streamed instead of being buffered (see [working with result sets](https://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#working-with-result-sets)). This could help when running pending jobs in "unlimited" mode and a very large queue.

plugins/BEdita/Core/src/Job/ServiceRunner.php
```diff
@@ -0,0 +1,166 @@
...
+     * Using an optional max number of jobs as 'limit'
+     * Resulting array will contain
+     *  - 'count' number oj jobs executed
+     *  - 'success' uuids of jobs successfully executed
+     *  - 'success' uuids of failed jobs
+     *
+     * @param int $limit Max number of pending jobs to run.
+     * @return array Result details
+     */
+    public static function runPending($limit = 0)
```
fquffio: Maybe a second argument `$minPriority = 0` could be useful. An usage scenario would be a "priority worker" that runs only pending jobs that have a priority higher than, say, 5.

fquffio (changes requested):

plugins/BEdita/Core/src/Job/ServiceRunner.php
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public static function reset()
+    {
+        static::$instances = [];
+    }
+
+    /**
+     * Run a service with $payload data read from async jobs table
+     *
+     * In $options array this parameters are available:
```
didoda: minor doc fix: "these parameters"

plugins/BEdita/Core/src/Job/ServiceRunner.php
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    /**
+     * Run pending jobs reading from async jobs table.
+     * Using an optional max number of jobs as 'limit'
+     * Resulting array will contain
+     *  - 'count' number oj jobs executed
+     *  - 'success' uuids of jobs successfully executed
+     *  - 'success' uuids of failed jobs
```
didoda: minor fixes in doc:
 *  - 'count' number of jobs executed
 [...]
 *  - 'failure' uuids of failed jobs

plugins/BEdita/Core/src/Job/ServiceRunner.php
didoda: I wouldn't add any extra logic, in `runPending`; `$minPriority` would add a possible `starvation` issue at this low level. High level management of jobs (including priority queues or semaphores) could be done elsewhere.
[this comment was a reply to @fquffio comment]

didoda (commented):
I suggested just some minor fixes in methods doc

plugins/BEdita/Core/src/Job/ServiceRunner.php
stefanorosanelli: that's right, done

plugins/BEdita/Core/src/Job/ServiceRunner.php
stefanorosanelli: done

plugins/BEdita/Core/src/Shell/JobsShell.php
stefanorosanelli: done

plugins/BEdita/Core/src/Job/ServiceRunner.php
stefanorosanelli: done

plugins/BEdita/Core/src/Shell/JobsShell.php
stefanorosanelli: That is the desirable behavior, we can probably handle it in `ServiceRunner` class

plugins/BEdita/Core/src/Job/ServiceRunner.php
stefanorosanelli: agreed, we will handle `priority` properly later

plugins/BEdita/Core/src/Job/ServiceRunner.php
stefanorosanelli: Right, this could be an improvement in case of large queues... otherwise a simple pagination will have similar effect

plugins/BEdita/Core/src/Job/ServiceRunner.php
stefanorosanelli: right, some exceptions are not logged but these are...

plugins/BEdita/Core/src/Job/ServiceRunner.php
stefanorosanelli: done

plugins/BEdita/Core/src/Job/ServiceRunner.php
stefanorosanelli: The idea was that both `$limit = 0` and `$limit = null` should run all pending jobs. 
But we can use just `null` - method signature changed accordingly...

plugins/BEdita/Core/src/Job/ServiceRunner.php
stefanorosanelli: Changed.
Both versions seem more or less equivalent...

fquffio (approved):

didoda (approved):
ok

PR #1216, opened 2017-05-05
Implement UUID authentication

### LINKED ISSUE #1209: Implement UUID external auth provider
Implement an external auth provider to allow registration of anonymous user via client UUID.

When the provider is active (row in `auth_providers`), the creation of a user can be done with the request

```http
POST /auth

Authorization: UUID <client-uuid>
```

This request will create a new user with `external_auth.provider_username` and `users.username` equal to `<client-uuid>`.

fquffio:
This PR resolves #1209.

UUID authentication is a way for anonymous users to silently sign-up and save data that is linked to their anonymous, yet unique, profile.

For UUID authentication to be enabled, an **AuthProvider** named `uuid` must be present in `auth_providers` table. If it is, then clients are allowed to do requests like:

```http
POST /auth
Authorization: UUID 7f11c23f-efa7-4880-a167-6aff69eff5af
```

The server will accept the request:
- if a **User** with an **ExternalAuth** of type `uuid` and `provider_username = '7f11c23f-efa7-4880-a167-6aff69eff5af'` already exists, the matching UUID will be considered sufficient to let the user log in.
- if no such user is found, a new **User** is created and it will be automatically given an **ExternalAuth** of type `uuid` and `provider_username` equals to the UUID in the request's Authorization header. This allows the client to log in again as long as it has the original UUID.

The only requirement is that the string following the prefix `UUID ` in the Authorization header is a valid UUID version 4 as per [RFC 4122](http://www.ietf.org/rfc/rfc4122.txt).

Response will contain JWT and renew token as usual, since UUID is a login credential from the server's point of view. Trying to provide an UUID as authorization token in other endpoints will result in a 401: it is only accepted in `/auth`.

### Security consideration
Client should consider UUID used for authentication as sensitive information, just like any JWT token or renew token.

Servers should accept login requests using UUID only via HTTPS, just like any other login request and like any other request in general.

So everybody already knows what to do, right?

### Questions
- I don't really like the piece of code that brutally attaches another authentication component in `LoginController::initialize()`. Suggestions welcome.
- Maybe we should disable UUID authentication for users that have `password_hash != NULL`?


stefanorosanelli:
* Yes, we should disable UUID authentication for users with not null `password`
* **Security** : use of API-KEYs in this case should be recommended or maybe forced... probably having permissions on some applications in `/auth` could be enough

plugins/BEdita/API/src/Controller/LoginController.php
```diff
@@ -39,34 +39,45 @@ public function initialize()
...
                 ],
-                false
-            );
+                'BEdita/API.Jwt' => [
+                    'queryDatasource' => true,
+                ],
+            ];
+
+            // TODO: better way to dynamically add external authentication components?
+            $uuidProvider = TableRegistry::get('AuthProviders')->find()->where(['name' => 'uuid'])->first();
```
stefanorosanelli: Agreed, it's problematic. We should have no knowledge of what type of providers are available.
Why not load **ALL** `auth_providers` ? 
Problem remains with configuration that may vary between providers. 
Why not impose a fixed configuration like this for now and then handle dynamic configurations as we create a new  `authProvider`?

PR #1217, opened 2017-05-05
Allow integer coordinates in "geo" finder

fquffio:
This PR fixes an issue that occurred when using `filter[geo][center]=` with a pair of coordinates, one of which was an integer. E.g.: `10 44.5` caused a `400 Bad Request`.


stefanorosanelli:
Are we sure that `10 44.5` are valid coordinates? 
Should they be `10.0 44.5` instead?
That said the PR seems ok

stefanorosanelli (approved):

fquffio:
I like more `10.0 44.5` too, but in some edge case the client might just omit the decimal part of latitude or longitude, and in that case a `400 Bad Request` response looks a bit too much severe to me, and a possible violation of robustness principle.

stefanorosanelli:
Still not sure if integer numbers are acceptable, but since there are not clear answers to that it's ok to be tolerant :wink:

PR #1222, opened 2017-05-08
Check for DB feature support instead of checking DB version

fquffio:
In some cases, checking DB version might not be enough: some functions might have been added via additional extensions that we just can't know of.

Therefore, when checking if backend has support for GIS functions like `ST_Distance_sphere` it is probably better to use an [EAFP](https://docs.python.org/2/glossary.html#term-eafp)-ish (Easier Ask Forgiveness than Permission) approach and try to run a test query that uses the required function — true EAFP cannot be achieved easily since queries are lazily executed. If that query fails the hard way, the database does not support GIS the way we need it, if it succeeds we can go on.

This also allows MySQL-compatible backends like MariaDB or Amazon Aurora to be properly checked, since their version number might not be aligned with the corresponding MySQL version.


fquffio:
This PR also extracts all `findGeo`-related logic to a dedicated Behavior in order to make logic slightly more reusable.

stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Behavior/GeometryBehavior.php
```diff
@@ -0,0 +1,184 @@
...
+     * Store geometry support for current connection.
+     *
+     * @var bool
+     */
+    protected $hasGeoSupport;
+
+    /**
+     * Get database expression to find distance between two points on a spheroid.
+     *
+     * @param string|float[] $point1 First point. Can be either a field name or a pair of floats.
```
didoda: I would add some more detail about points as floats, for example: "pair of floats representing respectively latitude and longitude, i.e. (44.496876, 11.343164)"

plugins/BEdita/Core/src/Model/Behavior/GeometryBehavior.php
```diff
@@ -0,0 +1,184 @@
...
+     * @param string|float[] $point2 Second point. Can be either a field name or a pair of floats.
```
didoda: same observation as in previous comment

plugins/BEdita/Core/src/Model/Behavior/GeometryBehavior.php
```diff
@@ -0,0 +1,184 @@
...
+            [],
+            'float'
+        );
+    }
+
+    /**
+     * Parse coordinates.
+     *
+     * @param mixed $point Coordinates.
+     * @return float[]
```
didoda: I'd add more detail about point as float array (again, as described in previous comments ;))

plugins/BEdita/Core/src/Model/Behavior/GeometryBehavior.php
```diff
@@ -0,0 +1,184 @@
...
+                new FunctionExpression('ST_GeomFromText', $point2),
+            ],
+            [],
+            'float'
+        );
+    }
+
+    /**
+     * Parse coordinates.
+     *
```
didoda, line 78: Some detail about "validation"? For instance: "A point has valid coordinates when latitude absolute value is greater than 90, lotitude absolute value is greater than 180"

plugins/BEdita/Core/src/Model/Behavior/GeometryBehavior.php
```diff
@@ -0,0 +1,184 @@
...
+            ]);
+        }
+
+        return $point;
+    }
+
+    /**
+     * Check if current DB supports geo operations.
+     *
+     * @return void
```
didoda: In a "check" function I'd expect a `@return boolean`.
However, `try/catch` logic does the job.

plugins/BEdita/Core/src/Model/Behavior/GeometryBehavior.php
```diff
@@ -0,0 +1,184 @@
...
+     * Find objects by geo coordinates.
+     *
+     * Create a query to filter objects using geo data: location objects are
+     * ordered by distance, from the nearest to the farthest using a center geo point.
+     *
+     * Accepted options are:
+     *   - 'center' with point coordinates, latitude and longitude
+     *   - 'from' with point coordinates, latitude and longitude
+     *   - 'radius' with positive decimal number
+     *
```
didoda: center-radius-from pattern search meaning could be explained with more details
I'd explicitate options meanings in context, i.e.:
* Accepted options are:
*    - 'center' with point coordinates, latitude and longitude | center point for the search
*    - 'radius' with positive decimal number | radius to compute circle around the center, containing coordinates
*    - 'from' with point coordinates, latitude and longitude | another location point, used to compute distances from positions found

Maybe we could also indicate a dummy real example use:
"find all locations close to Trafalgare Square (center), within 2 km (radius), refering how far these location are from King Cross (from)

didoda (commented):
Some doc and code review.

fquffio:
@didoda I have applied the changes you have requested. :)

PR #1224, opened 2017-05-10
signup endpoint first step: create user, job and send mail for activation

batopa:
The PR add addtional step to #1203 

After this PR a user can partially signup :smile: 

* a user with status `draft` will be created
* a job is added to `async_jobs`
* a confirmation email is send to the user

### Considerations

1. If no error occurs the response will be a `202 Accepted` with no body but I don't know if it's correct.
2. At the moment trying to signup a user with existing `username` or `email` responds with a `400 Bad Request` instead of `409 Conflict`.

### DIFF
--- a/plugins/BEdita/Core/src/Mailer/UserMailer.php
+++ b/plugins/BEdita/Core/src/Mailer/UserMailer.php
@@ -20,7 +20,6 @@ use Cake\ORM\TableRegistry;
 
 /**
  * Mailer class to send notifications to users
- * [Temporary dummy implementation]
  *
  * @since 4.0.0
  */
@@ -30,12 +29,12 @@ class UserMailer extends Mailer
      * Welcome message
      *
      * @param array $options Email options: 'to' (recipient)
-     * @return void
+     * @return $this
      * @codeCoverageIgnore
      */
     public function welcome($options)
     {
-        $this
+        return $this
             ->setTemplate('BEdita/Core.welcome')
             ->setLayout('BEdita/Core.default')
             ->setTo($options['to'])
@@ -50,7 +49,7 @@ class UserMailer extends Mailer
      * - `activationUrl` the activation url to follow
      *
      * @param array $options Email options
-     * @return void
+     * @return $this
      * @throws \LogicException When missing some required parameter
      */
     public function signup(array $options)
@@ -81,7 +80,7 @@ class UserMailer extends Mailer
             'appName' => $appName
         ]);
 
-        $this->setTemplate('BEdita/Core.signup')
+        return $this->setTemplate('BEdita/Core.signup')
             ->setLayout('BEdita/Core.default')
             ->setEmailFormat('both')
             ->setTo($user->email)
--- a/plugins/BEdita/Core/src/Mailer/UserMailer.php
+++ b/plugins/BEdita/Core/src/Mailer/UserMailer.php
@@ -51,6 +51,7 @@ class UserMailer extends Mailer
      *
      * @param array $options Email options
      * @return void
+     * @throws \LogicException When missing some required parameter
      */
     public function signup(array $options)
     {
--- a/plugins/BEdita/Core/src/Mailer/UserMailer.php
+++ b/plugins/BEdita/Core/src/Mailer/UserMailer.php
@@ -13,7 +13,10 @@
 
 namespace BEdita\Core\Mailer;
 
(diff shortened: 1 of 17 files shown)


plugins/BEdita/Core/src/Model/Action/SaveEntityAction.php
```diff
@@ -47,8 +47,10 @@ protected function initialize(array $config)
      */
     public function execute(array $data = [])
     {
-        $entity = $this->Table->patchEntity($data['entity'], $data['data']);
-        $success = $this->Table->save($entity);
+        $entityOptions = !empty($data['entityOptions']) ? (array)$data['entityOptions'] : [];
+        $saveOptions = !empty($data['saveOptions']) ? (array)$data['saveOptions'] : [];
```
fquffio: `(array)Hash::get($data, 'saveOptions')` makes Scrutinizer think this method has lower cyclomatic complexity. 🙃

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -0,0 +1,232 @@
...
+     *
+     * It needs to validate some data that don't concern the User entity
+     * as `activation_url` and `redirect_url`
+     *
+     * @param array $data The data to validate
+     * @return array
+     */
+    protected function validate(array $data)
+    {
+        $validator = new Validator();
```
fquffio, line 95: Maybe a validation step is worth adding to `BaseAction` itself…

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -0,0 +1,232 @@
...
+        try {
+            $options = [
+                'params' => [
+                    'userId' => $user->id,
+                    'activationUrl' => $this->getActivationUrl($job, $urlOptions),
+                ]
+            ];
+            $this->getMailer('BEdita/Core.User')->send('signup', [$options]);
+        } catch (SocketException $e) {
+            // @todo insert async job to retry sending email
```
fquffio, line 211: How about writing a custom `AsyncTransport` instead?

On a simple server setup, the default `Mail` or `Smtp` transport would be set, so that emails are sent synchronously.
On a setup that is able to properly manage asynchronous jobs, the custom `Async` transport is set so that every email is not sent immediately, but rather its details are written to the `async_jobs` table, and the job runner will take care to use those very same options on a real transport class (`Mail` or `Smtp` or whatever).

This way, on a multi-instance setup one could avoid telling SMTP secrets to the web servers, and set those pieces of configuration only on workers. 😏

plugins/BEdita/Core/src/Mailer/UserMailer.php
```diff
@@ -38,4 +41,50 @@ public function welcome($options)
...
+        $appName = ($currentApplication !== null) ? $currentApplication->name : 'BEdita';
+        $subject = __d('bedita', '{0} registration', [$appName]);
+
+        $this->set([
+            'user' => $user,
+            'activationUrl' => $options['params']['activationUrl'],
+            'appName' => $appName
+        ]);
+
+        $this->setTemplate('BEdita/Core.signup')
```
fquffio: Here it would probably be better to `return` the whole chain, so that a mail preview can be set up with DebugKit (see http://yourlocalbeditaurl.localhost/debug_kit/mail_preview).

fquffio (approved):

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
batopa: Yes, I think so. Maybe should we create another issue for that?

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
batopa: It makes sense. Please could you open a new issue?

plugins/BEdita/Core/src/Model/Action/SaveEntityAction.php
batopa: done

plugins/BEdita/Core/src/Mailer/UserMailer.php
batopa: done

PR #1233, opened 2017-05-15
Update test coverage

stefanorosanelli:
This PR updates and improves test coverage on some important files with low coverage

 * `TrashController` improvedi with Actions
 * some `@codeCoverageIgnore` and `@covers` added
 * deprecated `SaveSchemaTask` removed


plugins/BEdita/Core/src/ORM/Locator/TableLocator.php
```diff
@@ -37,10 +37,6 @@ class TableLocator extends CakeLocator
      */
     protected function _getClassName($alias, array $options = [])
     {
-        if (empty($options['className'])) {
-            $options['className'] = Inflector::camelize($alias);
-        }
-
```
didoda: ok for removing this part. Consequently `$options['className']` is considered necessary parameter.
we could add in function doc a comment about that.
for instance:
* @param array $options Table options array. It must contain 'className'.

plugins/BEdita/Core/tests/TestCase/ORM/Inheritance/QueryTest.php
```diff
@@ -219,6 +219,7 @@ public function aliasFieldProvider()
      * @dataProvider aliasFieldProvider
      * @covers ::fixAliasField()
      * @covers ::extractField()
+     * @covers ::aliasChecker()
```
didoda, line 222: should we add `* @covers ::inheritedTables()`?

(it doesn't increment code coverage, but it's another method called by `fixAliasField()`)

didoda (commented):
Minor suggestions...

plugins/BEdita/Core/src/ORM/Locator/TableLocator.php
stefanorosanelli: those lines were removed because they are useless, same code is called in parent class
https://api.cakephp.org/3.2/source-class-Cake.ORM.Locator.TableLocator.html#152

stefanorosanelli: but is probably wiser to leave it, as it is in base class implementation... i'll do it

stefanorosanelli: done

plugins/BEdita/Core/tests/TestCase/ORM/Inheritance/QueryTest.php
fquffio: For this purpose, the [`@uses`](https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.uses) annotation would probably fit better. However, we don't currently use it — it is meant for very strict test suites in which one is required to explicitely list _all_ the methods the test will flow through. If code executed during a test happens to be not listed amongst the `@uses`, the test [is marked as risky](https://phpunit.de/manual/current/en/risky-tests.html#risky-tests.unintentionally-covered-code). 🙈  It would be clearly unfeasible in our case.

fquffio (approved):

PR #1234, opened 2017-05-15
Allow custom JSON input

stefanorosanelli:
This PR add custom JSON input in situations where JSON-API input makes no sense.

Simplest way to activate it is ovverride controller `initialize` like this: 
```php
public function initialize()
    {
        parent::initialize();
        $this->JsonApi->setConfig('customJsonInput', true);
    }
```

`JsonApiComponent::parseInput()` will then allow custom json input if `application/json` is used as request content type


stefanorosanelli:
A better solution on controller `initialize()` - it just works

```php 
public function initialize()
    {
        parent::initialize();
        if ($this->request->is(['json'])) {
            $this->RequestHandler->setConfig('inputTypeMap.json', ['json_decode', true], false);
        }
    }
```

Issue #1207, opened 2017-04-27
Reduce the dimension of JWT access token


batopa:
The JWT generated to grant access to user is overweight. We generate it using all user data but we can reduce its weight using just useful and not sensitive information.

If client needs all user data it can use `GET /auth`.

I think we could store in JWT the user id and roles, something like

```json
{
  "id": 1,
  "roles": ["editor", "basic"]
}
```

or

```json
{
  "id": 1,
  "roles": [
    {
      "id": 14,
      "name": "editor"
    }
    {
      "id": 27,
      "name": "basic"
    }
  ]
}
```
@bedita/core what do you think?

stefanorosanelli:
Agree, I would probably add `username` also but for other data `GET /auth` is the way to go.

batopa:
Maybe we could define a default and make it configurable.

PR #1235, opened 2017-05-16
#1207 feat: reduce jwt access token

didoda:
This implements #1207

### DIFF
--- a/plugins/BEdita/API/src/Controller/LoginController.php
+++ b/plugins/BEdita/API/src/Controller/LoginController.php
@@ -93,23 +93,15 @@ class LoginController extends AppController
             throw new UnauthorizedException(__('Login not successful'));
         }
 
-        $fields = Configure::read('Security.jwt.user.fields') ?: ['id', 'roles', 'username'];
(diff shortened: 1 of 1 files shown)


fquffio:
I'm not sure that we want this to be configurable… JWT content is mostly used for internal purposes, so I'd rather go with a hardcoded list of fields… :/

didoda:
In #1207 @batopa commented: "Maybe we could define a default and make it configurable."
@fquffio suggests no config (easy to achieve, with trivial change in code).
If a specific config is not necessary now, I'd rather get rid of it and think about it in future (when needed).

fquffio:
Ok, I didn't read @batopa 's comment. I don't know… Is it necessary to make it configurable? I'd prefer avoid adding configuration keys until we actually need 'em — the more configuration keys we add now, the more likely it is to forget about them and re-invent the wheel somewhen in the future. 🙃

stefanorosanelli (approved):

batopa:
@fquffio in some situations the username could be considered a sensitive information, for example if it's an email, so I would avoid to store it in JWT. But as you said we can think about it later.

PR #1238, opened 2017-05-17
Enable login using JSON

stefanorosanelli:
This PR enables login using JSON input

```http
POST /auth
{
   "username": "{name}",
   "password": "{password}"
}
```

`Content-Type: application/json` is needed in this call whereas `Accept: application/json`is also acceptable here.

Use of `application/x-www-form-urlencoded` remains valid.

JSON API input parsing was modified:
* an empty json is accepted as `[]` - this happens when `Content-Type` is set but no data is passed
* in case of invalid JSON malformed JSON API a 400 Bad Request is returned


didoda (approved):

plugins/BEdita/API/src/Controller/Component/JsonApiComponent.php
```diff
@@ -75,23 +76,26 @@ public function initialize(array $config)
...
 
             return JsonApi::parseData((array)$json['data']);
         } catch (\InvalidArgumentException $e) {
-            if (Configure::read('debug') && !(empty($json))) {
-                throw $e;
-            }
-
-            return [];
+            throw new BadRequestException([
+                'title' => __d('bedita', 'Bad JSON input'),
```
didoda, line 96: Scrutinizer complains about this part:

https://scrutinizer-ci.com/g/bedita/bedita/inspections/47895a8d-f5cf-4ce2-ae68-8b2d53756a43/issues/files/plugins/BEdita/API/src/Controller/Component/JsonApiComponent.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

stefanorosanelli: since it's `Minor` and `Documentation` and frankly questionable I would skip it...

plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
```diff
@@ -76,19 +87,26 @@ public function testSuccessfulRenew(array $meta)
...
     public function testFailedLogin()
     {
-        $this->configRequestHeaders('POST');
+        $this->configRequestHeaders('POST', ['Content-Type' => 'application/json']);
 
         $this->post('/auth', ['username' => 'first user', 'password' => 'wrongPassword']);
 
         $this->assertResponseCode(401);
+
+        // using default 'application/vnd.api+json'
```
fquffio: I'd prefer to split these tests in two, one with invalid credentials, the other with wrong content type. No data providers: just two test cases for better understanding of the problem in case it fails. Same for `testSuccessfulLogin` test case, first with a content type, then with the other one.

However, this is a very minor concern.

fquffio (approved):

plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
stefanorosanelli: Makes sense, I'm splitting both methods

stefanorosanelli: Done

Issue #1237, opened 2017-05-17
Display application specific error code


stefanorosanelli:
In error response we need to display a `code` item in JSON API `error` response 
see http://jsonapi.org/format/#error-objects

PR #1239, opened 2017-05-17
Add error code

stefanorosanelli:
This PR solves #1237 

Throwing any Exception extending `CakeException` we can now add a `'code'` element that will be displayed in JSON API error output together with `status`, `title` and `detail`

An exception like:
```php
            throw new BadRequestException([
                'title' => 'bad request',
                'detail' => 'this was bad',
                'code' => 'error_code',
            ]);
```

will display:
```json
"error": {
   "status" : "400",
   "title": "bad request",
   "detail": "this was bad",
   "code": "error_code"
}
```

### DIFF
--- a/plugins/BEdita/API/src/Error/ExceptionRenderer.php
+++ b/plugins/BEdita/API/src/Error/ExceptionRenderer.php
@@ -141,7 +141,7 @@ class ExceptionRenderer extends CakeExceptionRenderer
         }
 
         $errorAttributes = $error->getAttributes();
-        if (empty($errorAttributes['code'])) {
+        if (empty($errorAttributes['code']) || !is_scalar($errorAttributes['code'])) {
             return '';
         }
 
--- a/plugins/BEdita/API/src/Error/ExceptionRenderer.php
+++ b/plugins/BEdita/API/src/Error/ExceptionRenderer.php
@@ -43,9 +43,10 @@ class ExceptionRenderer extends CakeExceptionRenderer
     {
         $isDebug = Configure::read('debug');
 
-        $code = $this->_code($this->error);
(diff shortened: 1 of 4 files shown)


plugins/BEdita/API/src/Error/ExceptionRenderer.php
```diff
@@ -43,9 +43,10 @@ public function render()
     {
         $isDebug = Configure::read('debug');
 
-        $code = $this->_code($this->error);
-        $message = $this->_message($this->error, $code);
+        $status = $this->_code($this->error);
```
didoda, line 46: A couple of minor observations about `$status` and `$code`:

1) why renaming `$code` to `$status`? it could be `http code` (=> `$httpCode`)
2) Regarding following statement: `$code = $this->appErrorCode($this->error);`:
function `appErrorCode` returns "string Error code". Then instead of `$code`, it could be `$errorCode` or `$appErrorCode`.

Disambiguisation of variables names could help in following `$this->jsonError($status, $title, $detail, $code, $trace)`
[=> `$this->jsonError($httpCode, $title, $detail, $appErrorCode, $trace)`]

didoda (approved):
It seems ok. Just minor observations about variables names.

plugins/BEdita/API/src/Error/ExceptionRenderer.php
stefanorosanelli: naming changes reflect error json output from JSON API spec http://jsonapi.org/format/#error-objects
http error code will be displayed as `status` etc etc...

see also `jsonError()` method to better understand naming choice

fquffio (approved):

PR #1247, opened 2017-05-23
Fix wrong field alias with query string search on deeply-inherited tables

fquffio:
This PR is basically an ugly workaround to alias fields the right way.

Visiting `GET /users?filter[query]=gustavo` now causes a 500 Internal Server Error because some fields are improperly aliased with `Profiles` instead of `Objects`. This happens for any table inheritance chain deeper than two levels.

Not any more! Thanks to this ugly PR, you shall be able to correctly search for Gustavo. (Hint: he's at the toilet.)

### DIFF
--- a/plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
@@ -159,8 +159,25 @@ class SearchableBehavior extends Behavior
             ]);
         }
 
+        $table = $this->getTable();
+        $tables = array_reverse(
+            array_merge(
+                [$table],
+                ($table instanceof InheritanceTable) ? $table->inheritedTables() : []
+            )
+        );
+        $aliasField = function ($field) use ($tables) {
+            /* @var \Cake\ORM\Table $table */
+            foreach ($tables as $table) {
+                if ($table->hasField($field, false)) {
+                    return $table->aliasField($field);
+                }
+            }
+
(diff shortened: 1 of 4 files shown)


plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
```diff
@@ -159,8 +159,25 @@ function ($word) use ($minLength) {
...
+        );
+        $aliasField = function ($field) use ($tables) {
+            /* @var \Cake\ORM\Table $table */
+            foreach ($tables as $table) {
+                if ($table->hasField($field, false)) {
+                    return $table->aliasField($field);
+                }
+            }
+
+            return $field;
```
stefanorosanelli: `codecov` complains that this line is not tested, could we add a simple test case for it?

if problematic we can skip this...

fquffio: This was an edge case, so hopefully this line was never hit. However, I have slightly refactored the closure, and now coverage should be 100%. 🎉

fquffio:
This PR also adds a `?q=` shorthand for `?filter[query]=`. This might come in handy in many scenarios. However, it is not fully JSON API compliant, since JSON API _recommends_ to always use `_` or `-` in non-standard query parameters.

I hope this won't keep us awake at night, anyways.

stefanorosanelli (approved):

Issue #1241, opened 2017-05-19
Reset and change password 


stefanorosanelli:
Implement recover/reset actions via `/auth` endpoint

1. request a password reset, creating an async job to be removed after some period (15/30 minutes?)
request could look like:
```http
POST /auth/recover
Content-Type: application/json

{
  "username": "{username}"
}
```
Should we add an `"email"` field in case username is not an email?

2. perform an actual password change using async job `{token}`

```http
POST /auth/recover
Content-Type: application/json

{
  "token": "{token}",
  "password": "{new-password}",
  "login": true
}
```
if `"login"` is true perform a login and return jwt token in meta to be used immediately

PR #1253, opened 2017-05-26
Login credentials update

stefanorosanelli:
This PR solves #1241 

Implementation is slightly different from original issue description

1. request a credential change, creating an async job that lasts for 24h, only contact method available is email:
```http
POST /auth/change

{
  "contact": "{my email}",
  "change_url": "{change url}"
}
```
Note:
 * in future versions we may add phone numbers and  other methods
 * this implies that `email` should be unique, as it is...
 * `{change url}` is a custom provided URL - user will receive an email with a link `{change url}?uuid={uuid}` this custom page will allow credentials change
 * response is `204 No Content` on success, could also be `200`, probably not `201` with location heaader since no information on the created resource should be given....

2. perform an actual credentials change using async job `{uuid}` 

```http
PATCH /auth/change
{
  "uuid": "{uuid}",
  "password": "{new password}",
  "login": true
}
```
Note:
 * if `"login"` is true login is performed and jwt tokens are returned in `meta` for immediate use; this key is optional; if missing `"login": false` is assumed
 * response will always contain user data


fquffio (approved):

plugins/BEdita/Core/src/Model/Action/ChangeCredentialsAction.php
```diff
@@ -0,0 +1,105 @@
...
+        $asyncJob = $this->AsyncJobs->get($data['uuid'], ['finder' => 'incomplete']);
+
+        if (empty($asyncJob->payload['user_id'])) {
+            throw new \LogicException(__d('bedita', 'Parameter "{0}" missing', ['payload.user_id']));
+        }
+
+        $user = $this->Users->get($asyncJob->payload['user_id'], ['contain' => ['Roles']]);
+        $user->password_hash = $data['password'];
+
+        return $this->Users->getConnection()->transactional(function () use ($user, $asyncJob) {
```
batopa, line 98: Since `AsyncJobs` could use a different connection from `Users` we would handle the transaction by hand doing something similar  to that done in `SignupUserAction`.

plugins/BEdita/Core/src/Model/Action/ChangeCredentialsRequestAction.php
```diff
@@ -0,0 +1,145 @@
...
+    public function validate(array $data)
+    {
+        $validator = (new Validator())
+            ->email('contact')
+            ->notEmpty('contact')
+            ->requirePresence('contact')
+
+            ->url('change_url')
+            ->notEmpty('change_url')
+            ->requirePresence('change_url');
```
batopa: I think that custom url validation is required to permit something as `myapp://` as in `SignupUserAction`.

batopa (commented):
Overall it's ok!

Maybe we could consider (in other issue) to refactor some common properties and methods between `Signup*Action` and `ChangeCredentials*Action` in a `Trait`.

plugins/BEdita/Core/src/Model/Action/ChangeCredentialsAction.php
stefanorosanelli: In this case there's nothing to rollback, a failure will stop following operations

Transaction simply removed

plugins/BEdita/Core/src/Model/Action/ChangeCredentialsRequestAction.php
stefanorosanelli: Right: a new `CustomUrlValidationProvider` class was added.

plugins/BEdita/Core/tests/TestCase/Model/Validation/CustomUrlValidationProviderTest.php
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * @coversDefaultClass \BEdita\Core\Model\Validation\CustomUrlValidationProvider
+ */
+class CustomUrlValidationProviderTest extends TestCase
+{
+
+    /**
+     * Test validation.
+     *
+     * @param array $expected Expected validation errors.
```
batopa: No need for any `@param`

batopa (changes requested):

plugins/BEdita/Core/tests/TestCase/Model/Validation/CustomUrlValidationProviderTest.php
stefanorosanelli: you're right
done

batopa (approved):

edoardocavazza:
Maybe the `change_url` should be in meta field as for `/signup` endpoint

stefanorosanelli:
You mean in `meta` section of the change request POST response?
It's on purpose to avoid security issues: token will be sent only using provided `contact`

edoardocavazza:
Something like this:

```http
POST /auth/change

{
    "data": {
        "type": "users",
        "attributes": {
            "email": "{my email}"
        },
        "meta": {
            "change_url": "{change url}"
        }
    }
}
```

similar to the `/signup` structure:

```http
POST /signup

{
    "data": {
        "type": "users",
        "attributes": {
            "username": "{my username}",
            "password": "{my password}",
            "email": "{my email}"
        },
        "meta": {
            "activation_url": "{activation url}",
            "redirect_url": "{redirect url}"
        }
    }
}
```

PR #1255, opened 2017-05-29
User verified field

### LINKED ISSUE #1254: User verification refactor
- new `users.verified` boolean column (or datetime?) to track user activation/validation
- field is `false` or `NULL` (if datetime) on signup, upon email confirmation becomes `true` or `current datetime`

- on **signup** `objects.status` of the user can be `draft` or `on` depending on configuration to define
- on **activation** if `status` is `draft` should be changed to `on` (like it is now), but `409 error` must be avoided if user status was already `on` with `verified` `false` or `NULL`

fquffio:
This PR adds a column `verified` to `users` table and resolves #1254.

Users' `verified` property represents the timestamp at which user verified their contact method. If this is `NULL`, the user hasn't verified their email yet.

In order to make update smoother, the database migration takes care of updating existing users whose `status = 'on'`.

Signup and activation processes now take into account this `verified` field, and act accordingly.


plugins/BEdita/Core/tests/TestCase/Model/Action/SignupUserActionTest.php
```diff
@@ -15,6 +15,7 @@
 
 use BEdita\Core\Model\Action\SignupUserAction;
 use BEdita\Core\Model\Entity\User;
+use Cake\Core\Configure;
```
batopa, line 18: Unused import

batopa (changes requested):

batopa (approved):

PR #1256, opened 2017-05-30
Add mail preview for password recovery message

fquffio:
This PR adds a mailer preview for email sent after a password recovery request.

It also includes a minor fix to docblock `@return` annotation. Excluding this, all other changes are in a class used for development purposes, hence no tests are present.

### DIFF
--- a/plugins/BEdita/Core/src/Mailer/Preview/UserMailerPreview.php
+++ b/plugins/BEdita/Core/src/Mailer/Preview/UserMailerPreview.php
@@ -17,7 +17,7 @@ use Cake\Utility\Text;
 use DebugKit\Mailer\MailPreview;
 
 /**
- * Mail preview for UserMailer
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Mailer/Preview/UserMailerPreview.php
```diff
@@ -13,11 +13,14 @@
 
 namespace BEdita\Core\Mailer\Preview;
 
+use Cake\Utility\Text;
 use DebugKit\Mailer\MailPreview;
 
 /**
  * Mail preview for UserMailer
```
didoda: `\BEdita\Core\Mailer\UserMailer` is not used, then I'd rather say "Mail preview for user mailer"

didoda (approved):
Very minor suggestion about doc

PR #1258, opened 2017-05-30
Service 'plugin.name' notation

stefanorosanelli:
This is a trivial PR to enable `my_plugin.service` notation for services.

Since `Inflector::camelize('my_plugin.service')` gives `"MyPlugin.service"` instead of `"MyPlugin.Service"` this may cause failures.

### DIFF
--- a/plugins/BEdita/Core/src/Job/ServiceRegistry.php
+++ b/plugins/BEdita/Core/src/Job/ServiceRegistry.php
@@ -49,11 +49,15 @@ class ServiceRegistry
             return static::$instances[$name];
         }
 
-        $className = Inflector::camelize($name);
-        $fullClassName = App::className($className, 'Job/Service', 'Service');
-        if ($fullClassName === false && strpos($className, '.') === false) {
-            $fullClassName = App::className('BEdita/Core.' . $className, 'Job/Service', 'Service');
+        $plugin = 'BEdita/Core.';
+        if (strpos($name, '.') !== false) {
+            list($plugin, $name) = explode('.', $name);
+            $plugin = Inflector::camelize($plugin) . '.';
         }
+
+        $className = $plugin . Inflector::camelize($name);
+        $fullClassName = App::className($className, 'Job/Service', 'Service');
+
         if ($fullClassName === false) {
             throw new \LogicException(__d('bedita', 'Unknown service "{0}"', [$name]));
         }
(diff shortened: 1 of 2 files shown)


didoda (approved):

plugins/BEdita/Core/src/Job/ServiceRegistry.php
```diff
@@ -49,11 +49,15 @@ public static function get($name)
...
-        if ($fullClassName === false && strpos($className, '.') === false) {
-            $fullClassName = App::className('BEdita/Core.' . $className, 'Job/Service', 'Service');
+        $plugin = 'BEdita/Core.';
+        if (strpos($name, '.') !== false) {
+            list($plugin, $name) = explode('.', $name);
+            $plugin = Inflector::camelize($plugin) . '.';
         }
+
+        $className = $plugin . Inflector::camelize($name);
+        $fullClassName = App::className($className, 'Job/Service', 'Service');
```
fquffio, line 59: This changes the semantics slightly… Using this approach it is impossible to have a service in the app (i.e. not within any plugin), which is not a scenario we are planning, but might be undesirable anyways, especially if and when we will decide to publish `BEdita/Core` and `BEdita/API` as standalone plugins to be re-used in non-standard ways.

I'd probably have changed only:
```diff
- $className = Inflector::camelize($name);
+ $className = implode('.', array_map([Inflector::class, 'camelize'], explode('.', $name)));
```

This should fix the problem without adding this restriction on app services. However, if we decide that app services can be safely excluded, this is good to go.

fquffio (commented):
Looks good, but changes semantics slightly and introduces a (probably minor) restriction.

If we feel like this restriction can be safely ignored, it's a go-go for me.

plugins/BEdita/Core/src/Job/ServiceRegistry.php
stefanorosanelli: Since BE4 has a plugin architecture it's not recommended to do anything in the `app` - the right way to customize something is via plugins: i would exclude `app` services

fquffio: I do not quite agree it's the "right" way. Using an approach similar to `cakephp/app` (users actually "fork" the app repository instead of merely cloning it, then install plugins into it) might have benefits as BEdita 4 becomes production ready — versioned `composer.lock`, customized configuration, …
In this case, placing logic in the app would be acceptable.

However, 👍  for now.

fquffio (approved):

PR #1261, opened 2017-06-05
Move HTML rendering to middleware layer

### LINKED ISSUE #1259: Move HTML rendering to middleware layer
At the present stage, developer-friendly HTML rendering happens partly in `AppController`, partly in `ExceptionRenderer`.

In my opinion, this is ugly and should be changed. We could move this feature in a middleware, in order to unify logic and make things simpler to maintain and test.

fquffio:
This PR resolves #1259. All HTML rendering logic was moved to a new `HtmlMiddleware`, in order to keep all code for that feature in a single place.

The middleware is added _at the top_ of the middleware queue¹ if, and only if, BEdita is configured to accept HTML requests. This means that in most production environments the middleware should just not be present in the queue, and the execution flow will be completely unaware of this developer feature.

Since HTML rendering used to affect many classes (`AppController`, `JsonApiComponent` and `ExceptionRenderer`, namely), a large number of files is involved in this PR. 😞 Additionally, many of the previous unit tests now just don't make sense, so they have been moved to integration tests.

Last but not least, `ExceptionRenderer` was collecting coverage from many test cases (all controller test classes were covering all the methods in `ExceptionRenderer` without any restriction), and `JsonApiComponent` was being covered in some unexpected places as well. Several test cases have been added, and several `@covers` annotations have been removed, in order to tidy up things a little.

------------------

¹ `HtmlMiddleware` is added at the top of the middleware queue, above `ErrorHandlerMiddleware`. This is required in order to render successful and error responses consistently, but it also means that if an exception is thrown _by_ `HtmlMiddleware`, a standard JSON response will be returned — you can see what happens by triggering any exception in `HtmlMiddleware::__invoke()`. Since this is very unlikely to happen, and it still is a developer-oriented feature, I'd consider it an acceptable risk and just leave it as is. If we want HTML rendering to be bullet-proof, however, we must add some try/catch blocks, which still won't catch syntax errors and similar in PHP 5.x.


fquffio:
As a welcome side effect, if additional plugins are loaded in BEdita, and they add custom controllers and routes, HTML rendering is available without re-defining templates or whatever in these plugins.

Since requests are now proxied almost unchanged, and "decorated" when all is said and done, all endpoint permissions are honoured as expected.

Also, requests other than `GET` can now be rendered in HTML. This opens the way for some in-browser client that's able to perform basic POST/PATCH/DELETE requests. 😏

fquffio:
I thought that maybe this kind of functionality might be moved to an external plugin `bedita/debug_kit` that adds some helpers for developers. This plugin might be installed and available by default in dev instances just like CakePHP's debug kit, but not installed in production environments unless explicitly asked.

This plugin might also include additional features like new panels for debug kit toolbar and the faker shell as well.

@stefanorosanelli @batopa @didoda What do you think?

didoda:
I agree: prod envs shouldn't have "dev" stuffs. External plugin for developers tools is a good practice.

batopa:
I agree too. The HTML rendering is just for developers, so I like the idea to place It in a separated plugin

didoda (approved):

stefanorosanelli:
Ok, having a separate plugin for devs stuff seems fine, but is there a dependency on `DebugKit`? Could they be loaded separately?

fquffio:
@stefanorosanelli It depends on what we want to have in the developer's plugin. If HTML rendering is the only feature, it doesn't require `cakephp/debug_kit`. If we plan to add, say, custom panels to the DebugKit toolbar (showing available object types and relations, for instance), or we also want to move mailer preview in the dev plugin, then it does require CakePHP's debug kit.

However, it might not be a mandatory dependency, but rather a suggestion, so that HTML rendering is always available, while debug kit panels and mailer previews are available only if `DebugKit` plugin is also loaded.

fquffio:
HTML rendering has now been moved to [`bedita/dev-tools`](https://github.com/bedita/dev-tools), a plugin in which we should place any development-related piece of code.

config/bootstrap.php
```diff
@@ -233,8 +234,15 @@
     ['bootstrap' => true, 'routes' => true, 'ignoreMissing' => true]
 );
 
-/**
- * Load common plugins like 'Migrations' and 'DebugKit' or other custom / 3rd party plugins
- * via configuration key 'Plugins'
+if (Configure::read('debug')) {
```
batopa: I would prefer to decide in `app.php` if I want to load the plugin.

```php
'Plugins' => [
        'BEdita/DevTools' => ['debugOnly' => true, 'bootstrap' => true],
]
```

In this way I could activate it just if I need it, optionally also with debug off.

plugins/BEdita/API/tests/TestCase/Event/CommonEventHandlerTest.php
```diff
@@ -43,7 +50,7 @@ public function testImplementedEvents()
     }
 
     /**
-     * test build middleware stack
+     * Test build middleware stack when HTML is acceptable.
```
batopa: :thinking: why "when HTML is acceptable"?

batopa (changes requested):

fquffio:
Both issues found by @batopa should have been addressed by now.

stefanorosanelli (approved):

PR #1264, opened 2017-06-06
Trigger events on user signup, activation or password recovery

### LINKED ISSUE #1263: Trigger events on signup, activation and password recovery
Some user-related actions should trigger proper events, to allow developers to attach their listeners and execute custom code on relevant events.

My proposal:

- `Auth.signup`, triggered on successful user signup, regardless of activation being required or not
- `Auth.signupActivation`, triggered when a user verifies their primary contact method
- `Auth.signupExpiration`, triggered when time expires for user signup, and the user is therefore deleted — can't be done at the present stage, since `SignupService` doesn't exist yet
- `Auth.credentialsChangeRequest`, triggered when a user initiates the process of password recovery
- `Auth.credentialsChange`, triggered when a user successfully recovers their password

fquffio:
This PR resolves #1263 and adds the documented events, so that developers can attach their listeners and hook into the flow.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
@@ -18,6 +18,9 @@ use BEdita\Core\Model\Entity\User;
 use BEdita\Core\Model\Validation\CustomUrlValidationProvider;
 use BEdita\Core\Utility\LoggedUser;
 use Cake\Core\Configure;
+use Cake\Event\Event;
+use Cake\Event\EventDispatcherTrait;
+use Cake\Event\EventListenerInterface;
 use Cake\I18n\Time;
 use Cake\Mailer\MailerAwareTrait;
 use Cake\Network\Exception\BadRequestException;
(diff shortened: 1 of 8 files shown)


plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -29,31 +32,35 @@
...
      * The AsyncJobs table
      *
      * @var \BEdita\Core\Model\Table\AsyncJobsTable
      */
-    protected $AsyncJobs = null;
+    protected $AsyncJobs;
 
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
```
stefanorosanelli, line 56: why `{@inheritdoc}` instead of `{@inheritDoc}` ?
https://www.phpdoc.org/docs/latest/guides/inheritance.html#the-inheritdoc-tag
both should work, but the camelcase version seems correct

stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
fquffio: Gotcha! 😱  It's merely cosmetic, I admit. It's just because PHPStorm complains about the camelCased version at times… 👀

PR #1265, opened 2017-06-07
 feat: spec shell script to generate yaml files

didoda:
This provides shell `spec` to generate yaml file using a speficied target file, merging yaml files from specified folder.
It adds `symfony/yaml` component to `vendors`, in order to perform merge and save of yaml target file.


fquffio (approved):
I'm quite scared of adding Composer stuff in a legacy system. 😱  I'd probably employ some kind of home-made PSR-0/4 class loader, if possible.

However this looks very self-contained, so hopefully everything's gonna be alright. 😉

didoda:
Good point, @fquffio. Actually I first look for a Cakephp 1.3 component implementing symfony yaml class, and I didn't find any. I didn't want to loose too much time in adapting symfony yaml to "old" Bedita.
So I tryed to install via composer in a specified folder, and then use `autoload`, as specified in yaml installation instructions. We could get rid of all composer stuff, and extra files, and just leave the classes. I know it's not a "clean" solution, but it's inexpensive.

batopa:
What about moving all in an addon? In this way we can keep clean the core.

didoda:
Feature developed in addon https://github.com/bedita/yaml_addons

PR #1266, opened 2017-06-07
National Identification and VAT number

stefanorosanelli:
Two very common fields are added to `profiles`: it makes sense to have both as core fields.

- https://en.wikipedia.org/wiki/National_identification_number
- https://en.wikipedia.org/wiki/VAT_identification_number 

No validation is done, is delegated to the client, could be particularly complicated.
Field dimensions are somehow overestimaded, but since they're `varchar` I don't see particular problems with them.


fquffio (approved):

plugins/BEdita/Core/config/Migrations/20170607114052_AddIdVatNumbers.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        $this->table('profiles')
+            ->addColumn('national_id_number', 'string', [
+                'comment' => '',
+                'null' => true,
+                'default' => null,
+                'limit' => 32,
+            ])
+            ->addColumn('vat_number', 'string', [
+                'comment' => '',
```
fquffio: Comments for these two fields are missing. However column name is quite self-explanatory.

fquffio (approved):

plugins/BEdita/Core/config/Migrations/20170607114052_AddIdVatNumbers.php
stefanorosanelli: you're right, I forgot...

didoda (approved):

PR #1268, opened 2017-06-15
Api key via query string

stefanorosanelli:
Although generally not recommended in some situations passing `api_key` in query string can be useful or necessary: for instance in `webhooks` like method calls where you have poor or no control over request headers.

Probably this possibility should be application dependent.

Minor refactor on `$this->endpoint` and `$this->application` included + improved coverage.

### DIFF
--- a/plugins/BEdita/API/src/Auth/EndpointAuthorize.php
+++ b/plugins/BEdita/API/src/Auth/EndpointAuthorize.php
@@ -170,7 +170,7 @@ class EndpointAuthorize extends BaseAuthorize
         if ($this->application === null) {
             $apiKey = $this->request->getHeaderLine($this->_config['apiKeyHeaderName']);
             if (empty($apiKey)) {
-                $apiKey = $this->request->getQuery($this->_config['apiKeyQueryString']);
+                $apiKey = strval($this->request->getQuery($this->_config['apiKeyQueryString']));
             }
             if (empty($apiKey) && empty($this->_config['blockAnonymousApps'])) {
                 return null;
--- a/plugins/BEdita/API/src/Auth/EndpointAuthorize.php
+++ b/plugins/BEdita/API/src/Auth/EndpointAuthorize.php
@@ -140,8 +140,7 @@ class EndpointAuthorize extends BaseAuthorize
     protected function unauthenticate()
     {
         $controller = $this->_registry->getController();
-
-        return $controller
+        $controller
             ->Auth->getAuthenticate('BEdita/API.Jwt')
             ->unauthenticated($controller->request, $controller->response);
     }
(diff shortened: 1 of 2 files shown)


plugins/BEdita/API/src/Auth/EndpointAuthorize.php
```diff
@@ -146,31 +159,36 @@ public function isAnonymous($user)
...
     protected function getApplication()
     {
-        $application = CurrentApplication::getApplication();
-        if ($application === null) {
-            $header = $this->request->getHeaderLine($this->_config['apiKeyHeaderName']);
-            if (empty($header) && empty($this->_config['blockAnonymousApps'])) {
+        $this->application = CurrentApplication::getApplication();
+        if ($this->application === null) {
+            $apiKey = $this->request->getHeaderLine($this->_config['apiKeyHeaderName']);
+            if (empty($apiKey)) {
```
fquffio: (suggestion) I'd probably add a `&& $this->_config['apiKeyQueryString'] !== null`, so that API key via query string can be quickly disabled by setting that configuration to `null`.

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
```diff
@@ -123,12 +135,13 @@ public function authorize($user, ServerRequest $request)
...
      *
-     * @return void
+     * @return mixed
      */
     protected function unauthenticate()
     {
         $controller = $this->_registry->getController();
-        $controller
+
+        return $controller
```
fquffio: `BaseAuthenticate::unauthenticated()` method must not return any value, according to its documentation. This return is not needed, and not used, as far as I can tell. 🤔

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
```diff
@@ -146,31 +159,36 @@ public function isAnonymous($user)
...
     {
-        $application = CurrentApplication::getApplication();
-        if ($application === null) {
-            $header = $this->request->getHeaderLine($this->_config['apiKeyHeaderName']);
-            if (empty($header) && empty($this->_config['blockAnonymousApps'])) {
+        $this->application = CurrentApplication::getApplication();
+        if ($this->application === null) {
+            $apiKey = $this->request->getHeaderLine($this->_config['apiKeyHeaderName']);
+            if (empty($apiKey)) {
+                $apiKey = strval($this->request->getQuery($this->_config['apiKeyQueryString']));
```
fquffio: (minor) I generally prefer type-casting, thus `(string)$var` instead of `strval($var)`. They should yield the same results, so it's just a matter of taste.

fquffio (changes requested):
Seems good. I have just a doubt with a returned value that seems unnecessary, a suggestion for quickly disabling this feature if and when we'll want it to be disabled, and a minor style suggestion.

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
stefanorosanelli: That's right, I did it just to make sure code coverage is 100%, maybe there's a better way to do it

Removing it anyway

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
stefanorosanelli: :+1:

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
stefanorosanelli: ok

fquffio (approved):

PR #1269, opened 2017-06-17
Use subqueries for CTI

fquffio:
This PR introduces a major refactor to how we handle CTI. @batopa and I have worked on this in the past two days to work around a couple of weird scenarios that caused events not to be dispatched correctly for several model-related operations.

Previously, we used to inherit from parent tables using `INNER JOIN`s and later merging results in the descendant's entity. This caused troubles because we had to take care of fields being aliased correctly, and in complex queries this could cause conflicts when one table joined the party more than once with the same alias because it was being inherited from by two different models.

Now, we "enclose" all inheritance stuff in a single subquery that is used in the `FROM` clause. This makes things significantly simpler as we don't need to traverse queries and mess up with selected fields or fields that might be used in conditions, but rather "expose" a single table in lieu of the whole inheritance chain.

Basically, we turned this:

```sql
SELECT Objects.title AS Objects__title, Profiles.email AS Profiles__email, Users.username AS Users__username
    FROM users AS Users
    INNER JOIN profiles AS Profiles ON Profiles.id = Users.id
    INNER JOIN objects AS Objects ON Objects.id = Profiles.id
    WHERE Objects.status = 'on'; -- ...plus post-processing to merge Object&Profile into User entity
```

…into this:

```sql
SELECT Users.title AS Users__title, Users.email AS Users__email, Users.username AS Users__username
    FROM (
        SELECT *
            FROM users
            INNER JOIN profiles ON profiles.id = users.id
            INNER JOIN objects ON objects.id = profiles.id
    ) AS Users
    WHERE Users.status = 'on';
```

In order to do this, we got rid of the `ExtensionOf` association, that is no longer used, so save and delete are now handled in a slightly different way.

These changes should be fully backwards-compatible. The only non-breaking change in a method's signature is for `Table::extensionOf()` method, that no longer uses the second argument.

### Pros

 - Easier to maintain.
 - Less error-prone.
 - Tables that inherit from another table are now seen as a single table in the main query. This means that it should be easier to hack into queries for developers.
 - No more nasty things happen in a `beforeFind` event. This means that, if needed, table inheritance can be even opted-out just for one query by explicitly setting the `FROM` clause.
 - Types are now handled correctly in `WHERE` clause (for instance, previously we had to write `0` instead of `false` to add conditions on boolean columns — not any more).
 - Associations are now inherited. :tada: Whether inheriting behaviours in the same fashion is to be discussed.

### Cons

Potentially performances. Joins are generally more efficient than sub queries. However, after some basic benchmarking on MySQL 5.7.18, the two approaches seem to be comparable in terms of query execution time. My test was with a very basic inheritance scenario (only two tables) and 10M records.

I'm going to release a benchmark shell in a public Gist so that we can run performance tests on several setups using real BEdita objects (Documents, Profiles, Users and Locations) and compare average and peak times for both save and find operations — `4-cactus` against this branch.


fquffio:
## Benchmark results

All tests have been run using [BenchmarkShell](https://gist.github.com/fquffio/811fb4e038425fb12672cf7af1d14ead) on the same machine under similar load conditions.

Tests:

1. Time to save an object (following tests are run on a data set generated by this test). 10k samples:
    `bin/cake benchmark insert 10000`
2. Time to load and hydrate ~2k users. 1k samples:
    `bin/cake benchmark select users 1000`
3. Time to load and hydrate <10 locations, filtering by an indexed field and a not indexed field. 1k samples:
    `bin/cake benchmark select locations 1000 'filter[country_name]=Italy&filter[status]=on'`

(more tests to come)

### MySQL 5.7.18

| # | `4-cactus` | `refactor/use-subqueries-for-CTI` | ? |
|---|------------|-----------------------------------|---|
| 1 | Avg: 0.01861s <br> Max: 0.07736s | Avg: 0.01302s <br> Max: 0.03961s | ✅ |
| 2 | Avg: 0.01704s <br> Max: 0.38284s | Avg: 0.01451s <br> Max: 0.01792s | ✅ |
| 3 | Avg: 0.00255s <br> Max: 0.01867s | Avg: 0.00230s <br> Max: 0.00581s | ✅ |

### MySQL 5.6.35

| # | `4-cactus` | `refactor/use-subqueries-for-CTI` | ? |
|---|------------|-----------------------------------|---|
| 1 | Avg: 0.01793s <br> Max: 0.11711s | Avg: 0.07946s <br> Max: 1.57220s | 🏴❓  |
| 2 | Avg: 0.03088s <br> Max: 0.34302s | Avg: 0.04933s <br> Max: 0.06458s | ❌ |
| 3 | Avg: 0.00370s <br> Max: 0.02117s | Avg: 0.02884s <br> Max: 0.05555s | 🏴 |

### SQLite 3.16.0

| # | `4-cactus` | `refactor/use-subqueries-for-CTI` | ? |
|---|------------|-----------------------------------|---|
| 1 | Avg: 0.01679s <br> Max: 0.06485s | Avg: 0.01732s <br> Max: 0.06246s | ❌ |
| 2 | Avg: 0.03694s <br> Max: 0.08215s | Avg: 0.04099s <br> Max: 0.07864s | ❌ |
| 3 | Avg: 0.00147s <br> Max: 0.01659s | Avg: 0.00140s <br> Max: 0.01574s | ✅ |

### PostgreSQL 9.6

| # | `4-cactus` | `refactor/use-subqueries-for-CTI` | ? |
|---|------------|-----------------------------------|---|
| 1 | Avg: 0.02253s <br> Max: 0.08103s | Avg: 0.01851s <br> Max: 0.08297s | ✅ |
| 2 | Avg: 0.01588s <br> Max: 0.29765s | Avg: 0.01487s <br> Max: 0.35224s | ✅ |
| 3 | Avg: 0.00182s <br> Max: 0.02506s | Avg: 0.00170s <br> Max: 0.02453s | ✅ |

stefanorosanelli:
About performance & benchmarks: 
 * we will get rid of MySQL 5.6 sooner or later (when 8.0 is out, or before probably)
 * SQLite performance is basically the same, and we probably won't use it in production scenarios with lots of data

I don't see any problems.

plugins/BEdita/Core/src/Model/Entity/Profile.php
```diff
@@ -41,4 +41,15 @@
  */
 class Profile extends ObjectEntity
 {
+
+    /**
+     * {@inheritDoc}
+     */
+    protected $_hidden = [
```
stefanorosanelli: are these field no more inherited from `ObjectEntity` ?

fquffio: They still are. This piece of code is completely useless. 🙃  I'm reverting these changes.

plugins/BEdita/Core/src/Model/Entity/User.php
```diff
@@ -62,6 +62,8 @@ class User extends Profile
      * {@inheritDoc}
      */
     protected $_hidden = [
+        'created_by_user',
```
stefanorosanelli: Remove this `$_hidden` array like in `Profile`?
Probably only `password_hash` has to be added? In constructor?

stefanorosanelli (approved):

didoda (approved):
Everything seems ok. I can't think about more possible tests; this provides a core feature, so side effects not visible now would be arisen soon: we should deal with that in the next weeks.

PR #1274, opened 2017-06-22
Fix save of object types

fquffio:
This PR is a hotfix for a bug introduced in #1269.


plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
```diff
@@ -628,13 +628,14 @@ public function testEdit()
         $this->assertResponseCode(200);
```
stefanorosanelli, line 628: I would add here and in `UsersControllerTest::testEdit()` a check on JSON response, redundant maybe...
```php
        $this->assertEquals($data['id'], $result['data']['id']);
        $this->assertEquals($data['type'], $result['data']['type']);
        $this->assertEquals($data['attributes']['title'], $result['data']['attributes']['title']);
```

fquffio: Done

stefanorosanelli (approved):

PR #1280, opened 2017-07-05
ExpiredTokenException with application code

stefanorosanelli:
In case of expired JWT token a custom exception is thrown and a custom application code sent in error response to allow safer identification by client apps

Some problems:
 * where should applications code like `'be_token_expired'` be placed? An error codes class/interface?
 * had to change `testAuthenticate` method to test exception attributes, I did not found another way to test `$exception->getAttributes()` using `expectException*` methods


batopa (commented):
Seems ok for me. Maybe adding a specific test for `ExpiredTokenException` class should be better.

Issue #1252, opened 2017-05-24
Many warnings are polluting error log


fquffio:
Right now, `error.log` is a bit polluted by warning messages that are caused by a behaviour that I personally consider rather normal and not harmful.

The scenario is a bit knotty:
- let's suppose there is a **plugin** that adds a table for an object type called "MyObjectType"
- let's also suppose that there is a relation called "MyRelation" between "MyObjectType" and any other object type
- let's suppose that it does exist an object of type "MyObjectType" with ID = 2
- issuing a request `GET /my_object_type/2/my_relation` or `GET /my_object_type/2/relationships/my_relation` will cause the server to return a correct response, but a warning will be written to `BEDITA/logs/error.log` even if everything apparently did go well

Steps to reproduce from the console:
```
$ bin/cake console

>>> // First, we must restore the logger, since `ConsoleShell` drops all loggers before initializing.
>>> \Cake\Log\Log::setConfig('error', ['className' => 'Console', 'stream' => 'php://stdout'])

>>> // Next, we need to initialize `ListRelatedObjectsAction`
>>> $MyObjectType = \Cake\ORM\TableRegistry::get('MyObjectType')
>>> $action = new \BEdita\Core\Model\Action\ListRelatedObjectsAction(['association' => $MyObjectType->MyRelation]);

>>> // Now, we execute the action. This will cause a warning to be written to error log (in this case, stdout).
>>> $action->execute(['primaryKey' => 2, 'filter' => []])->count()
2017-05-24 22:04:14 Warning: Record not found in table "object_types"
```

This is caused by the fact that in `TableLocator` we log a warning every time an attempt is made to load a table whose name doesn't meet some requirements. Before 2526b2f this didn't happen, since the error was being logged only in case it wasn't a `RecordNotFoundException`.

However, I'm not quite sure how to proceed. 🤔  Surely it's a weird thing, but I'm not quite sure that it should be logged. Frankly, I'd _at least_ lower the logging level in case the exception is a `RecordNotFoundException`, but I'd probably prefer to revert to the old behaviour (record not found -> not logged at all).

What do you think?

didoda:
"a table whose name doesn't meet some requirements" and `RecordNotFoundException` don't sound relate with each other, I mean "semantically".
'Record not found' should not be an issue, it should be just an empty set result, shouldn't it?
If it is so, then I agree with reverting to previous behaviour
```
if (!($e instanceof RecordNotFoundException)) {
    $this->log($e, 'warning');
[...]

PR #1281, opened 2017-07-05
Avoid unwanted TableLocator log

stefanorosanelli:
Quick & dirty way to remove unwaned error logs locading related objects.

There's surely a better way to do it :wink:

### DIFF
--- a/plugins/BEdita/Core/src/ORM/Locator/TableLocator.php
+++ b/plugins/BEdita/Core/src/ORM/Locator/TableLocator.php
@@ -28,13 +28,6 @@ class TableLocator extends CakeLocator
     use LogTrait;
 
     /**
-     * Aliases to exclude from loading as object type name
-     *
-     * @var array
-     */
-    protected $excludeAlias = ['_InverseAssociation'];
-
-    /**
      * Gets the table class name.
      *
      * @param string $alias The alias name you want to get.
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/ORM/Locator/TableLocator.php
```diff
@@ -52,12 +59,14 @@ protected function _getClassName($alias, array $options = [])
             return $className;
         }
 
-        try {
-            $objectTypes = $this->get('ObjectTypes');
-            $objectType = $objectTypes->get($alias);
-            $options['className'] = $objectType->table;
-        } catch (\Exception $e) {
-            $this->log($e->getMessage(), 'warning');
+        if (!in_array($alias, $this->excludeAlias)) {
```
fquffio: `if (substr($alias, 0, 1) !== '_')`, maybe? We could then consider all aliases starting with `_` as "reserved", even though this should be quite rare.

fquffio:
References #1252 .

plugins/BEdita/Core/src/ORM/Locator/TableLocator.php
stefanorosanelli: Agreed, more general solution.
We should definitely add rules on custom object types creation.

fquffio: It is already impossible to have an object type whose name starts with an underscore, since object type aliases are forcibly camelized through inflector.

stefanorosanelli: done

PR #1282, opened 2017-07-05
Allow `HEAD /status`

fquffio:
This MR fixes a 405 Method Not Allowed error on `/status` endpoint when request method is HEAD instead of GET.

It also introduces a minor refactor to how requests are accepted even though they haven't a suitable `Accept` header: instead of overwriting the request detector, if a request would be rejected because of its `Accept` header, it is overwritten and set as `application/json`.

### DIFF
--- a/plugins/BEdita/API/src/Controller/StatusController.php
+++ b/plugins/BEdita/API/src/Controller/StatusController.php
@@ -13,6 +13,7 @@
 namespace BEdita\API\Controller;
 
 use BEdita\Core\Utility\System;
+use Cake\Core\Configure;
 
 /**
  * Controller for `/status` endpoint.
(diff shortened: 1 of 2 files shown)


plugins/BEdita/API/src/Controller/StatusController.php
```diff
@@ -29,10 +28,8 @@ class StatusController extends AppController
      */
     public function initialize()
     {
-        if (!$this->request->is('jsonapi')) {
-            Request::addDetector('json', function (Request $request) {
-                return true;
-            });
+        if (!$this->request->is('jsonapi') && ((!Configure::read('debug') && !Configure::read('Accept.html')) || !$this->request->is('html'))) {
```
stefanorosanelli: Only for readability I'd prefer
```php 
       $htmlResponse = $this->request->is('html') && (Configure::read('debug') || Configure::read('Accept.html'));
       if (!$this->request->is('jsonapi') && !$htmlResponse) {
```

stefanorosanelli (approved):

PR #1283, opened 2017-07-05
Fix missing CORS headers on errors

batopa:


plugins/BEdita/API/src/Middleware/CorsMiddleware.php
```diff
@@ -12,17 +12,15 @@
...
+use Cake\Network\CorsBuilder;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
 /**
  * Handle cross-origin HTTP requests setting the proper headers.
  *
  * The response of preflight request (OPTIONS) is delivered directly after the headers are applied.
- * For simple requests the CORS headers are applied to the response then it is passed to next middleware.
+ * For simple requests the CORS headers are applied before send response.
```
fquffio: (minor) I'd suggest "before sending response" or "before response is sent".

plugins/BEdita/API/src/Middleware/CorsMiddleware.php
```diff
@@ -229,8 +135,9 @@ protected function buildCors(ServerRequestInterface $request, ResponseInterface
         }
 
         $response = $corsBuilder->build();
-        if (!empty($origin) && !$response->hasHeader('Access-Control-Allow-Origin')) {
-            throw new ForbiddenException('Origin not allowed');
+
+        if ($response->getHeaderLine('Access-Control-Allow-Origin') != '*') {
```
fquffio: (minor) Here I'd prefer a strict check `!==`. Since `*` is not a string that could be type-casted to a strange value, it is probably pretty much the same, so it's just a matter of taste. :)

(Same at line 129)

fquffio (approved):

plugins/BEdita/API/tests/TestCase/Event/CommonEventHandlerTest.php
```diff
@@ -13,6 +13,7 @@
 namespace BEdita\API\Test\Event;
```
fquffio: (Not part of this PR) Namespace is wrong: `BEdita\API\Test\TestCase\Event`.

Also, these imports are unnecessary:
```
use Cake\Network\Exception\UnauthorizedException;
use Cake\ORM\TableRegistry;
```

plugins/BEdita/API/src/Middleware/CorsMiddleware.php
batopa: done

plugins/BEdita/API/src/Middleware/CorsMiddleware.php
batopa: done

plugins/BEdita/API/tests/TestCase/Event/CommonEventHandlerTest.php
batopa: done

PR #1285, opened 2017-07-12
Save entity log & test

stefanorosanelli:
This simple PR improves logs on save entity action errors and coverage to related unit tests.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/SaveEntityAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SaveEntityAction.php
@@ -56,7 +56,7 @@ class SaveEntityAction extends BaseAction
         if ($success === false) {
             $errors = $entity->getErrors();
             if (!empty($errors)) {
-                $this->log(sprintf('Entity save errors: %s', print_r($errors, true)), 'warning');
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Model/Action/SaveEntityAction.php
```diff
@@ -56,15 +56,15 @@ public function execute(array $data = [])
         if ($success === false) {
             $errors = $entity->getErrors();
             if (!empty($errors)) {
-                $this->log('Entity save failed', 'debug', compact('entity'));
+                $this->log(sprintf('Entity save errors: %s', print_r($errors, true)), 'warning');
```
fquffio: Probably `json_encode` is better. Services that ingest logs often have parsers for JSON (even if mixed with other plain test), but not for PHP's `print_r` format.

fquffio (approved):

plugins/BEdita/Core/src/Model/Action/SaveEntityAction.php
stefanorosanelli: seems right
done

Issue #1292, opened 2017-07-18
First /admin implementation


stefanorosanelli:
``/admin`` endpoint deals with administrative operations on current project

Managed resources are: 
  * configuration properties
  * async jobs
  * applications including api keys
  * custom endpoints

PR #1293, opened 2017-07-18
Admin endpoint

stefanorosanelli:
This PR solves #1292 

First minimal implementation of `/admin` endpoint 

Through `/admin/{resource}` and usual POST, GET, PATCH, DELETE methods some resources are managed namely:
  * `/admin/applications` 
  * `/admin/async_jobs`
  * `/admin/ednpoints`
  * `/admin/config`

No relationships are displayed for `applications` and `endpoints` - no available for `config` and `async_jobs` 

A new trait `JsonApiAdminTrait` has been introduced to override default JSON API links and relationships display.

Some checks must be introduced at model/table level to avoid dangerous operations like overriding API KEY in use or writing malformed configurations. This will be addressed in a next issue.


plugins/BEdita/API/tests/IntegrationTest/AdminResourcesTest.php
```diff
@@ -0,0 +1,134 @@
...
+                [
+                    'content' => 'Another Configuration',
+                 ],
+            ],
+        ];
+    }
+
+    /**
+     * Test CRUD on admin resources
+     *
```
didoda, line 86: missing `@param $type`

didoda (approved):

plugins/BEdita/API/tests/IntegrationTest/AdminResourcesTest.php
stefanorosanelli: :+1:      
done!

plugins/BEdita/API/tests/IntegrationTest/AdminResourcesTest.php
```diff
@@ -0,0 +1,135 @@
...
+     * @var array
+     */
+    public $fixtures = [
+        'plugin.BEdita/Core.config'
+    ];
+
+    /**
+     * Data provider for `testResource`
+     */
```
batopa: Missing `@return`

plugins/BEdita/API/tests/IntegrationTest/AdminResourcesTest.php
batopa: Missing `@return` too

plugins/BEdita/API/tests/IntegrationTest/AdminResourcesTest.php
```diff
@@ -0,0 +1,135 @@
...
+        $this->assertContentType('application/vnd.api+json');
+
+        $locationHeader = $this->_response->getHeaderLine('Location');
+        $this->assertNotEmpty($locationHeader);
+        $resourceId = substr($locationHeader, strrpos($locationHeader, '/') + 1);
+
+        // VIEW
+        $this->configRequestHeaders();
+        $this->get("$endpoint/$resourceId");
+        $result = json_decode((string)$this->_response->getBody(), true);
```
batopa: unused `$result` variable. Missing a test on it?

plugins/BEdita/API/tests/TestCase/Controller/AdminControllerTest.php
```diff
@@ -0,0 +1,127 @@
...
+            'type' => 'applications',
+            'attributes' => [
+                'name' => 'another-app',
+                'description' => 'Another app',
+            ],
+        ];
+
+        $this->configRequestHeaders('POST', $this->getUserAuthHeader());
+        $this->post('/admin/applications', json_encode(compact('data')));
+        $result = json_decode((string)$this->_response->getBody(), true);
```
batopa: `$result` not used

plugins/BEdita/Core/tests/TestCase/Model/Entity/JsonApiAdminTraitTest.php
```diff
@@ -0,0 +1,84 @@
...
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
+namespace BEdita\Core\Test\TestCase\Model\Entity;
+
+use BEdita\Core\Utility\JsonApiSerializable;
```
batopa: Useless `use` statement for `JsonApiSerializable` and `Hash`

batopa (changes requested):

plugins/BEdita/Core/tests/TestCase/Model/Entity/JsonApiAdminTraitTest.php
stefanorosanelli: done

plugins/BEdita/API/tests/TestCase/Controller/AdminControllerTest.php
stefanorosanelli: done

plugins/BEdita/API/tests/IntegrationTest/AdminResourcesTest.php
stefanorosanelli: removed for now

a detailed view test would be preferable, maybe not in IntegrationTest?

plugins/BEdita/API/tests/IntegrationTest/AdminResourcesTest.php
stefanorosanelli: added

plugins/BEdita/API/tests/IntegrationTest/AdminResourcesTest.php
stefanorosanelli: added

fquffio:
I like the approach to have a single controller manage multiple resources, however I don't really like having a single `AdminController`. `admin/` prefix sounds like something that might become really complex and "wide".

Rather than a single `AdminController` I'd have a `Controller/Admin` namespace with a `ResourcesController` in it. Routes would then be connected using a [prefix](https://book.cakephp.org/3.0/en/development/routing.html#prefix-routing).

This is not blocking: we could switch to this approach in the future if and when we'll actually need this.

stefanorosanelli:
Agreed, an `/admin` prefix and a `Controller\Admin` namespace are both good ideas.

We could to it, among with other necessary improvements, in one or more new issues.

batopa (approved):

PR #1295, opened 2017-07-19
Project.name config

stefanorosanelli:
This PR adds a `Project` configuration key to contain project information.

`Project.name` will be used in email subjects to users as default.

### DIFF
--- a/plugins/BEdita/Core/tests/TestCase/Mailer/UserMailerTest.php
+++ b/plugins/BEdita/Core/tests/TestCase/Mailer/UserMailerTest.php
@@ -352,8 +352,6 @@ class UserMailerTest extends TestCase
      */
     public function testGetProjectName($expected, $configured)
     {
-        $current = Configure::read('Project.name');
-
         Configure::write('Project.name', $configured);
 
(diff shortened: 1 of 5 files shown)


plugins/BEdita/Core/tests/TestCase/Mailer/UserMailerTest.php
```diff
@@ -368,5 +367,8 @@ public function testGetProjectName($expected, $application = null)
         $viewVars = $this->Email->getViewVars();
         static::assertArrayHasKey('projectName', $viewVars);
         static::assertEquals($expected, $viewVars['projectName']);
+
+        // restore conf
+        Configure::write('Project.name', $current);
```
fquffio: This is not needed. Configuration is automatically restored in tearDown.

fquffio (approved):

plugins/BEdita/Core/tests/TestCase/Mailer/UserMailerTest.php
stefanorosanelli: Right, done

didoda (approved):

PR #1301, opened 2017-07-28
Include relations in object types cache

### LINKED ISSUE #1226: Add cache for relation and relation types
Right now a simple request triggers about 30 SQL queries, most of which happen on `relations` and `relation_types` table. This is far too much and could quickly cause problems: we should definitely employ some caching mechanism to avoid repeated calls and cache BEdita metadata across subsequent requests.

fquffio:
This PR fixes #1226 by including information about relations in object types cache.

In complex scenarios, where relations are being navigated in depth, we are currently performing the same query over and over to find out which relations are attached to the current object type. In such scenarios this PR might significantly reduce the amount of queries being performed — a simple test in a very complex scenario lowered the amount of queries by almost 75%. 😱

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
@@ -18,8 +18,6 @@ use Cake\Cache\Cache;
 use Cake\Database\Expression\QueryExpression;
 use Cake\Database\Schema\TableSchema;
 use Cake\Datasource\Exception\RecordNotFoundException;
-use Cake\Event\Event;
-use Cake\ORM\Entity;
 use Cake\ORM\Query;
 use Cake\ORM\RulesChecker;
 use Cake\ORM\Table;
@@ -85,7 +83,6 @@ class ObjectTypesTable extends Table
             'conditions' => [
                 $through->aliasField('side') => 'left',
             ],
-            'cascadeCallbacks' => true,
         ]);
         $through = TableRegistry::get('RightRelationTypes', ['className' => 'RelationTypes']);
         $this->belongsToMany('RightRelations', [
@@ -96,7 +93,6 @@ class ObjectTypesTable extends Table
             'conditions' => [
                 $through->aliasField('side') => 'right',
             ],
-            'cascadeCallbacks' => true,
         ]);
     }
 
@@ -201,7 +197,7 @@ class ObjectTypesTable extends Table
 
         if (empty($options)) {
             $options = [
-                'key' => self::getCacheKey($primaryKey),
+                'key' => sprintf('id_%d_rel', $primaryKey),
                 'cache' => self::CACHE_CONFIG,
                 'contain' => ['LeftRelations.RightObjectTypes', 'RightRelations.LeftObjectTypes'],
             ];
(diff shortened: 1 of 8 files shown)


plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -197,10 +197,13 @@ public function get($primaryKey, $options = [])
             $primaryKey = $allTypes[$primaryKey];
         }
 
-        $options += [
-            'key' => 'id_' . $primaryKey,
-            'cache' => self::CACHE_CONFIG,
-        ];
+        if (empty($options)) {
+            $options = [
+                'key' => 'id_' . $primaryKey,
```
stefanorosanelli: this is minor minor minor .... since this cache contains relations information maybe a better name could be `related_1` or `1_related` ?

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
```diff
@@ -343,15 +354,15 @@ public function testInvalidateCacheAfterDelete()
...
+        static::assertNotFalse(Cache::read('id_1', ObjectTypesTable::CACHE_CONFIG));
+        static::assertNotFalse(Cache::read('map', ObjectTypesTable::CACHE_CONFIG));
+        static::assertNotFalse(Cache::read('map_singular', ObjectTypesTable::CACHE_CONFIG));
 
         $this->ObjectTypes->delete($entity);
 
-        $this->assertFalse(Cache::read('id_1', ObjectTypesTable::CACHE_CONFIG));
-        $this->assertFalse(Cache::read('map', ObjectTypesTable::CACHE_CONFIG));
-        $this->assertFalse(Cache::read('map_singular', ObjectTypesTable::CACHE_CONFIG));
+        static::assertFalse(Cache::read('id_1', ObjectTypesTable::CACHE_CONFIG));
```
stefanorosanelli: When an object type is removed or its name modified should we invalidate also the relations cache of other object types that can be related to this one?
Since this is not a frequent operation we could simply remove all `id_*` caches

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
fquffio: I agree. What about `id_1_with_relations`? `related_1` looks to me a bit like if that cache contains _only_ informations about that object type's relations. 🤔

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
fquffio: Yes, you're right. 😞  I'm fixing it.

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
stefanorosanelli: I'd prefer to keep it short as possibile: `id_1_rel` or `id_1_relations` ?

fquffio:
Ok, both issues should have been addressed by now. Renaming the cache key was pretty easy, but correctly invalidating any "relatable" object type was much harder.

So far, the models attempt to invalidate as little cache keys as possible. However, since things are now rather complex, and since they might get even harder in the future when we'll need to cache properties and stuff as well, I'm wondering if we should rather invalidate the whole object types cache every time an operation is performed, as Stefano suggested in his review.

This would be sub-optimal, but surely it would make things simpler to understand, possibly to maintain, and quite likely it wouldn't severely affect applications: relations and object types aren't invalidated often, and their cache is quick to be recreated.

@batopa @stefanorosanelli ?

stefanorosanelli (approved):

stefanorosanelli:
I'd prefer the simple approach to invalidate whole cache for the exact reasons you just explained..

batopa:
In general I would prefer to invalidate the cache more often even it's not optimized. I would possibly avoid to complicate things that could become a mess to understand. So I am for keep it simple

fquffio:
We all agree, then. 👍  I'm making the required changes, then I'll ask you both to review changes once more.

stefanorosanelli (approved):

batopa (approved):

PR #1305, opened 2017-08-01
Allow base64-encoded uploads

fquffio:
This PR allows clients to upload files using base64-encoded payloads instead of raw binary streams.

Clients that wish to upload a base64-encoded payload **MUST** include the header `Content-Transfer-Encoding: base64` in their request.

This feature is available _only_ for `POST /streams/upload/:fileName` endpoint.


fquffio:
After discovering [this section](http://php.net/manual/en/filters.convert.php) of PHP documentation, it seems that the correct stream filter to use is always `convert.base64-decode`. Therefore, I have removed the fallbacks to `string.base64` stream filter and `base64decode()` function, that are probably useless and practically impossible to test.

PR #1310, opened 2017-08-02
Issue/v4/1308 trash filter

### LINKED ISSUE #1308: Add filters to `/trash` endoint
`/trash` endpoint should provide some of the common filters on `GET` requests  

https://docs.bedita.net/en/latest/endpoints/parameters.html#filters

`Type`, `Field` and `Query` filters are needed.

stefanorosanelli:
This PR fixes #1308

* in `TrahController` actions are used instead of `Table` calls
* a new integration test was added
* on hard delete failure in `DeleteEntityAction` now an exception is thrown, and related test was updated

### DIFF
--- a/plugins/BEdita/API/src/Controller/TrashController.php
+++ b/plugins/BEdita/API/src/Controller/TrashController.php
@@ -13,6 +13,8 @@
 namespace BEdita\API\Controller;
 
 use BEdita\Core\Model\Action\DeleteObjectAction;
+use BEdita\Core\Model\Action\GetObjectAction;
+use BEdita\Core\Model\Action\ListObjectsAction;
 use BEdita\Core\Model\Action\SaveEntityAction;
 use Cake\Network\Exception\ConflictException;
 use Cake\Network\Exception\InternalErrorException;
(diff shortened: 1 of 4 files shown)


plugins/BEdita/Core/src/Model/Action/DeleteEntityAction.php
```diff
@@ -41,6 +43,11 @@ protected function initialize(array $data)
      */
     public function execute(array $data = [])
     {
-        return $this->Table->delete($data['entity']);
+        $success = $this->Table->delete($data['entity']);
```
fquffio: Would `$this->Table->deleteOrFail($data['entity'])` do the job anyways?

It throws a `PersistenceFailedException` that provides some further details in logs, and still results in a **500 Internal Server Error** response, so IMHO it should be preferred.

plugins/BEdita/API/src/Controller/TrashController.php
```diff
@@ -13,6 +13,8 @@
 namespace BEdita\API\Controller;
 
 use BEdita\Core\Model\Action\DeleteObjectAction;
+use BEdita\Core\Model\Action\GetObjectAction;
+use BEdita\Core\Model\Action\ListObjectsAction;
 use BEdita\Core\Model\Action\SaveEntityAction;
 use Cake\Network\Exception\ConflictException;
 use Cake\Network\Exception\InternalErrorException;
```
fquffio: This import is now unused and can be removed.

fquffio (changes requested):

plugins/BEdita/Core/src/Model/Action/DeleteEntityAction.php
stefanorosanelli: agreed

plugins/BEdita/API/src/Controller/TrashController.php
stefanorosanelli: ok

fquffio (approved):

batopa (approved):

PR #1318, opened 2017-08-17
 #1186 feat: relations link to available objects

### LINKED ISSUE #1186: Link for available objects to relate with another one
In order to display a list of available objects to connect through a relationship with another one, maybe would be useful an additional entry in the `links` section of a relationship API call which contains the url to retrieve the list.

Examples:
```js
// documents/5/relationships/authors
{
    "data": [...],
    "meta": {...},
    "links": {
        "self": "...",
        "home": "...",
        "available": ".../objects?filter[type][]=profiles&filter[type][]=authors",
     }
}
```
or
```js
// users/1/relationships/roles
{
    "data": [...],
    "meta": {...},
    "links": {
        "self": "...",
        "home": "...",
        "available": ".../roles",
     }
}
```

didoda:
This PR fixes #1186 

Function `getRelationships` (in `JsonApiTrait` and `ObjectEntity`) provides a new url `available`, that points to objects available by relation. So each relation part response will have 3 links: `related`, `self`, `available`. I.e.:
```
// http://be4api/roles
[...]
"relationships": {
    "users": {
        "links": {
        "related": "http://be4api/roles/2/users",
        "self": "http://be4api/roles/2/relationships/users",
        "available": "http://be4api/users"
    }
}
[...]
// http://be4api/boats
[...]
"relationships": {
    "boats_booked": {
        "links": {
            "related": "http://be4api/boats/13786/boats_booked",
            "self": "http://be4api/boats/13786/relationships/boats_booked",
            "available": "http://be4api/objects?filter[type][0]=bookings"
        }
    },
    "owned_by": {
        "links": {
            "related": "http://be4api/boats/13786/owned_by",
            "self": "http://be4api/boats/13786/relationships/owned_by",
            "available": "http://be4api/objects?filter[type][0]=users"
        }
    }
}
[...]
```


plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
```diff
@@ -128,12 +129,14 @@ public function testIndex()
                             'links' => [
                                 'related' => 'http://api.example.com/documents/2/test',
                                 'self' => 'http://api.example.com/documents/2/relationships/test',
+                                'available' => 'http://api.example.com/objects?filter%5Btype%5D%5B0%5D=documents&filter%5Btype%5D%5B1%5D=profiles',
```
stefanorosanelli: can we use the more readable version `filter[type][]=documents&filter[type][]=profiles` ?

plugins/BEdita/Core/src/Model/Entity/ObjectEntity.php
```diff
@@ -185,6 +185,33 @@ protected function getRelationships()
                 ],
                 true
             );
+            $entityDest = $this->getTable()->associations()->getByProperty($relationship)->getTarget()->newEntity();
+            if ($entityDest instanceof JsonApiSerializable) {
+                $destObj = $entityDest->jsonApiSerialize(JsonApiSerializable::JSONAPIOPT_BASIC);
+                if (empty($destObj['type'])) {
+                    $options = ['name' => $relationship];
+                    $result = TableRegistry::get('ObjectTypes')
```
stefanorosanelli: Is there a way to avoid this query (performed on every item, I guess) using some sort of cache, like object types cache?

stefanorosanelli (changes requested):

plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
fquffio: It's that version, only URL-encoded. Also, due to some reason, PHP likes to encode array keys as well when building query strings, thus it becomes `filter[type][0]=documents&filter[type][1]=profiles`.

plugins/BEdita/Core/src/Model/Entity/ObjectEntity.php
fquffio: There is. Object types cache already includes information about relations and relatable object types. You might take a look at `RelationsBehavior` to see how it's done in there, or I might take care of this myself in the afternoon.

didoda: I'll take a look at `RelationsBehavior` ;) if I keep lost, I'll ask for help ^_^

didoda: I didn't find any cache related part in `RelationsBehavior`... maybe I misunderstood the comments

plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
batopa: Since the URL must be encoded I think it is correct. Very ugly, but correct.

From JsonAPI spec:

> Note: The above example URI shows unencoded [ and ] characters simply for readability. In practice, these characters must be percent-encoded, per the requirements in RFC 3986.

We could make the tests more readble writing something as

```php
'available' => sprintf(
    'http://api.example.com/objects?%s',
    http_build_query(['filter' => ['type' => ['documents', 'profiles']]])
)
```

Just a little problem with that. CakePHP uses `http_build_query()` to generate encoded query string that by default uses `RFC 1738` that converts spaces with `+` instead of `%20`.

plugins/BEdita/Core/src/Model/Entity/ObjectEntity.php
```diff
@@ -185,6 +185,33 @@ protected function getRelationships()
...
+                    $available = Router::url(
+                        [
+                            '_name' => 'api:objects:index',
+                            'object_type' => 'objects',
+                            'filter' => ['type' => array_values($result)],
+                        ],
+                        true
+                    );
+                } else {
+                    $available = Router::url(
```
fquffio: These lines are not covered by unit tests. It should be possible to cover them by adding a case to the existing test case's data provider that asserts the presence of the `relationships.roles.links.available` key for a User.

fquffio (changes requested):

plugins/BEdita/Core/src/Model/Entity/ObjectEntity.php
didoda: It seems that test is already covered by `BEdita\API\Test\TestCase\Controller\ObjectsControllerTest->testIndex (/var/www/workspace/bedita4/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php:363)`

fquffio: @didoda Yes, but that test case is not specifically covering these lines. Instead of extending that test case coverage, I'd duplicate that test in `ObjectEntityTest` class: code duplication is fine within tests. 🙃

didoda: I just put a breakpoint in that line and test passed from there... covering means something else than "passing from there", doesn't it? Then I'll write a test method in `ObjectEntityTest`.

fquffio: Code coverage is measured only for classes and/or methods explicitly set using `@covers` annotation. This helps preventing unintentionally covered code ("coverage by accident"), where a test case that's supposed to test the functionality of a component A does indeed cover code of another component B, thereby not pointing out the need for specific tests for component B.

As a rule of thumb, a `MyWhateverTest` class **SHOULD** only `@covers` methods from `MyWhatever` class in order to avoid confusion, while integration tests should `@coversNothing` since they are "high level" tests that are more focused on the correct interaction between several components rather than each line of code from a single component.

In this case, `ObjectsControllerTest` does only `@covers` methods from `ObjectsController`, therefore lines of code executed in other components (including `ObjectEntity`) are discarded, as those lines are "used"<sup>1</sup> rather than "covered".

See [specifying covered methods](https://phpunit.de/manual/current/en/code-coverage-analysis.html#code-coverage-analysis.specifying-covered-methods) in PHPUnit docs for reference.

------------

<sup>1</sup> "Used" components **SHOULD** be always explicitly stated via the `@uses` annotation in every test case. However, since in our case we use so many components in every operation, this is practically unfeasible. We have therefore configured PHPUnit in a way that it does not enforce the use of `@uses` annotation. If we were to use the "strict" mode, tests that execute lines of code from components not declared via that annotation [would be marked as "risky"](https://phpunit.de/manual/current/en/risky-tests.html#risky-tests.unintentionally-covered-code).

fquffio (approved):

plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
fquffio: @stefanorosanelli I should have fixed this problem using the solution suggested by Alberto.

stefanorosanelli (approved):

PR #1320, opened 2017-08-21
FIx unable to save date ranges in some cases

### LINKED ISSUE #1299: Unable to save date ranges in some CTI cases
Creating a new model/table that extends `Locations` or `Profiles` we are unable to save `date_ranges` data  in a classic **POST** call like

```json
{
    "data": {
	  "type": "dockings",
           "attributes": {
        	"title": "New Title",
                "date_ranges": [
                   {
                     "start_date": "2017-08-01T23:00:00+00:00",
                     "end_date": "2017-08-30T13:00:00+00:00",
                     "params": null
                  }
            ]
        }
    }
}
```

Resulting object has empty `date_ranges`. 
If same model extends `Objects` data are saved correctly.

However: adding `date_ranges` to `Users` and saving works.... :thinking: ... weird behavior

batopa:
This PR fixes #1299

The fix is small and passes all tests but it is a change quite sensitive and I don't know if there was a reason for how it was done :scream:

### DIFF
--- a/plugins/BEdita/Core/src/ORM/Inheritance/Table.php
+++ b/plugins/BEdita/Core/src/ORM/Inheritance/Table.php
@@ -66,8 +66,8 @@ class Table extends CakeTable
      */
     public function extensionOf($associated)
     {
-        // If it is an alias, and it already exists in the registry, immediately load the instance.
-        if (is_string($associated) && TableRegistry::exists($associated)) {
+        // If it is an alias, immediately load the instance.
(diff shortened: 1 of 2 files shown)


stefanorosanelli (approved):
Did some tests and it works now

fquffio:
Only life will teach us if this was done for a good reason.

If it was done for a good reason and tests were missing it would have been a mistake anyways.

PR #1323, opened 2017-08-22
Travis coverage and HHVM 

stefanorosanelli:
This PR removes coverage analysis from all builds and HHVM support

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,14 +24,7 @@ jobs:
   fast_finish: true
 
   include:
-    # Reference tests set with coverage: PHP 7.1 and MySQL 5.7
-    - stage: reference
-      php: 7.1
-      env: "DB=mysql DB_VERSION=5.7 db_dsn='mysql://root@localhost/bedita_test'"
-      script: vendor/bin/phpunit --coverage-clover=clover.xml # Run PHPUnit with code coverage.
-      after_success: bash <(curl -s https://codecov.io/bash) # Upload test coverage reports to CodeCov.
-
-    # Minimal tests set: SQLite on PHP 5.6/7.0/7.1, PostgreSQL on PHP 7.1, code sniffer.
+    # Minimal tests set: SQLite on PHP 5.6/7.0, MySQL 5.7 on PHP 7.0 and code sniffer.
     - stage: minimal
       php: 5.6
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -161,7 +155,7 @@ before_script:
   - if [ "$DB" = 'pgsql' ]; then psql -c 'CREATE DATABASE bedita_test;' -U postgres; fi
   - if [ "$DB" = 'pgsql' ] && [ "$DB_POSTGIS" = '1' ]; then psql -c 'CREATE EXTENSION postgis;' -U postgres; fi
 
-script: vendor/bin/phpunit --coverage-clover=clover.xml # Run PHPUnit.
```
batopa, line 164: The clover.xml is used for Codecov https://codecov.io in `after_success` there is the script to upload it.
Could it be created just in a build?

batopa (changes requested):

.travis.yml
stefanorosanelli: That's right, I missed it

stefanorosanelli: a new `reference` stage is introduced with PHP 7.1, MySQL 5.7 and code coverage with codecov upload - all other builds will not have coverage

fquffio: I'd keep it in the "minimal" stage. A "reference" stage with only one job will not allow other jobs to run in parallel, thus increasing the overall time required for the build to complete.

Also, since coverage varies slightly from database to database, I'd like to have coverage for at least one job on every database vendor — say, all PHP 7.1 test jobs. If not, things like the fact that a proper error is raised when the database does not support geospatial functions would never be covered, since MySQL _has_ such support indeed.

batopa (approved):

stefanorosanelli:
Build stages have been changed this way:

* **Minimal** SQLite on PHP 5.6/7.0, MySQL 5.7 on PHP 7.0 and code sniffer.
* **Reference** (includes coverage + codecov upload) PHP 7.1 with MySQL 5.7, SQlite and PostgreSQL 9.6 + PostGIS
* **Complete** SQLite on PHP 5.6 with lowest dependencies, MySQL 5.7 on PHP 5.6, PostgreSQL 9.6 on PHP 5.6/7.0, MySQL 5.6 and PostgreSQL 9.5 on PHP 7.1, static code analyzer.
* **Deploy** unchanged

PR #1324, opened 2017-08-22
 #1322 feat: admin endpoint only as administrator

### LINKED ISSUE #1322: `/admin` must be accessible to `administrator` role only
To access `/admin` endpoint user must be logged and have `administator` role - even if no permissions are set on `/admin` endpoint

didoda:
This PR fixes #1322 

Only administrator users can call `/admin` endpoint: system raises 401 if user is anonymous or it's not an administrator.

### DIFF
--- a/plugins/BEdita/API/src/Auth/EndpointAuthorize.php
+++ b/plugins/BEdita/API/src/Auth/EndpointAuthorize.php
@@ -17,6 +17,7 @@ use BEdita\Core\Model\Entity\Application;
 use BEdita\Core\Model\Entity\Endpoint;
 use BEdita\Core\Model\Entity\EndpointPermission;
 use BEdita\Core\State\CurrentApplication;
+use BEdita\Core\Utility\LoggedUser;
 use Cake\Auth\BaseAuthorize;
 use Cake\Datasource\Exception\RecordNotFoundException;
 use Cake\Http\ServerRequest;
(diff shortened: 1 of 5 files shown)


plugins/BEdita/API/src/Auth/EndpointAuthorize.php
```diff
@@ -17,6 +17,7 @@
 use BEdita\Core\Model\Entity\Endpoint;
 use BEdita\Core\Model\Entity\EndpointPermission;
 use BEdita\Core\State\CurrentApplication;
+use BEdita\Core\Utility\LoggedUser;
```
stefanorosanelli: unnecessary, see below

plugins/BEdita/API/src/Controller/AdminController.php
```diff
@@ -65,6 +65,9 @@ public function initialize()
         if (isset($this->JsonApi)) {
             $this->JsonApi->setConfig('resourceTypes', [$this->resourceName]);
         }
+
+        $this->Auth->getAuthorize('BEdita/API.Endpoint')->setConfig('blockAnonymousUsers', true);
```
stefanorosanelli: unnecessary

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
```diff
@@ -96,6 +97,20 @@ public function authorize($user, ServerRequest $request)
             $this->unauthenticate();
         }
 
+        // if 'administratorOnly' configuration is true logged user must have administrator role
```
stefanorosanelli: I suggest this simpler change just before  `if ($this->isAnonymous($user) && $this->authorized !== true)`

```php
if ($this->authorized && $this->getConfig('administratorOnly')) {
 $this->authorized = in_array(self::ADMIN_ROLE, Hash::extract($user, 'roles.{n}.id'));
} 
```

a definition of `const ADMIN_ROLE = 1;` should be added to class

Explanation: 
 * if user is unlogged `$this->authorized` will be false and user will get a 401 anyway
 * if user is logged but without administrator role response will be 403

plugins/BEdita/API/tests/TestCase/Controller/AdminControllerTest.php
```diff
@@ -69,11 +69,17 @@ public function testIndex()
...
-        $this->configRequestHeaders();
+        // GET /admin with anonymous user
+        $this->configRequestHeaders('GET');
+        $this->get('/admin/applications');
+        $result = json_decode((string)$this->_response->getBody(), true);
+        static::assertResponseCode(401);
+        static::assertContentType('application/vnd.api+json');
+        // GET /admin with authenticated user
+        $authHeader = $this->getUserAuthHeader();
+        $this->configRequestHeaders('GET', $authHeader);
```
stefanorosanelli: a testcase of a user logged with no administrator role is needed - response must be 403

stefanorosanelli (changes requested):

stefanorosanelli (approved):

PR #1329, opened 2017-08-24
 #1325 feat: 403 Forbidden on delete role/user 1

### LINKED ISSUE #1325: Admin role and user not removable
* it should be impossible to remove user 1 and role 1
* also association between user 1 and role 1 is not changeable

only possible modifications may occur via custom query, not using common `delete` model methods

didoda:
This PR fixes #1325 

On role/user 1 delete and role-1/user-1 association delete: system returns 403 Forbidden

### DIFF
--- a/plugins/BEdita/API/src/Controller/ResourcesController.php
+++ b/plugins/BEdita/API/src/Controller/ResourcesController.php
@@ -25,7 +25,6 @@ use BEdita\Core\Model\Action\SetAssociatedAction;
 use Cake\Core\InstanceConfigTrait;
 use Cake\Network\Exception\BadRequestException;
 use Cake\Network\Exception\ConflictException;
-use Cake\Network\Exception\ForbiddenException;
 use Cake\Network\Exception\InternalErrorException;
 use Cake\Network\Exception\NotFoundException;
 use Cake\ORM\Association\BelongsTo;
@@ -33,7 +32,6 @@ use Cake\ORM\Association\HasOne;
 use Cake\ORM\Query;
 use Cake\ORM\TableRegistry;
 use Cake\Routing\Router;
-use Cake\Utility\Hash;
 use Cake\Utility\Inflector;
 
 /**
@@ -228,11 +226,6 @@ abstract class ResourcesController extends AppController
         $entity = $action(['primaryKey' => $id, 'contain' => $contain]);
 
         if ($this->request->is('delete')) {
-            // If object is not deletable returns 403 FORBIDDEN
-            if ((in_array($this->modelClass, ['Roles', 'Users'])) && ((int)$id === 1)) { // ID 1 not removable
-                throw new ForbiddenException(__d('bedita', 'Could not delete "{0}" 1', $this->modelClass));
-            }
-
             // Delete an entity.
             $action = new DeleteEntityAction(['table' => $this->Table]);
 
@@ -324,10 +317,6 @@ abstract class ResourcesController extends AppController
                 break;
 
             case 'DELETE':
-                $ids = Hash::extract($this->request['data'], '{n}.id');
-                if (in_array($this->modelClass, ['Roles', 'Users']) && ((int)$id === 1) && in_array('1', $ids)) {
-                    throw new ForbiddenException(__d('bedita', 'Could not update relationship "{0}"', $relationship));
-                }
                 $action = new RemoveAssociatedAction(compact('association'));
                 break;
 
--- a/plugins/BEdita/API/src/Controller/ResourcesController.php
+++ b/plugins/BEdita/API/src/Controller/ResourcesController.php
@@ -25,6 +25,7 @@ use BEdita\Core\Model\Action\SetAssociatedAction;
 use Cake\Core\InstanceConfigTrait;
 use Cake\Network\Exception\BadRequestException;
 use Cake\Network\Exception\ConflictException;
+use Cake\Network\Exception\ForbiddenException;
 use Cake\Network\Exception\InternalErrorException;
 use Cake\Network\Exception\NotFoundException;
 use Cake\ORM\Association\BelongsTo;
@@ -32,6 +33,7 @@ use Cake\ORM\Association\HasOne;
 use Cake\ORM\Query;
 use Cake\ORM\TableRegistry;
 use Cake\Routing\Router;
+use Cake\Utility\Hash;
 use Cake\Utility\Inflector;
 
 /**
@@ -226,6 +228,11 @@ abstract class ResourcesController extends AppController
         $entity = $action(['primaryKey' => $id, 'contain' => $contain]);
 
         if ($this->request->is('delete')) {
+            // If object is not deletable returns 403 FORBIDDEN
+            if ((in_array($this->modelClass, ['Roles', 'Users'])) && ((int)$id === 1)) { // ID 1 not removable
+                throw new ForbiddenException(__d('bedita', 'Could not delete "{0}" 1', $this->modelClass));
+            }
+
             // Delete an entity.
             $action = new DeleteEntityAction(['table' => $this->Table]);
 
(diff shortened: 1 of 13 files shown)


plugins/BEdita/API/src/Controller/ResourcesController.php
```diff
@@ -226,6 +228,11 @@ public function resource($id)
         $entity = $action(['primaryKey' => $id, 'contain' => $contain]);
 
         if ($this->request->is('delete')) {
+            // If object is not deletable returns 403 FORBIDDEN
+            if ((in_array($this->modelClass, ['Roles', 'Users'])) && ((int)$id === 1)) { // ID 1 not removable
+                throw new ForbiddenException(__d('bedita', 'Could not delete "{0}" 1', $this->modelClass));
+            }
+
             // Delete an entity.
             $action = new DeleteEntityAction(['table' => $this->Table]);
```
stefanorosanelli: This way it works, but I don't know if this check should be performed inside `DeleteEntityAction` to make sure we have same behavior in other possible use cases

plugins/BEdita/API/src/Controller/ResourcesController.php
```diff
@@ -226,6 +228,11 @@ public function resource($id)
         $entity = $action(['primaryKey' => $id, 'contain' => $contain]);
 
         if ($this->request->is('delete')) {
+            // If object is not deletable returns 403 FORBIDDEN
+            if ((in_array($this->modelClass, ['Roles', 'Users'])) && ((int)$id === 1)) { // ID 1 not removable
```
stefanorosanelli: This way it works, but I don't know if this check should be performed inside DeleteEntityAction to make sure we have same behavior in other possible use cases

Same applies to `RemoveAssociatedAcion`

didoda: Fair enough

stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Action/DeleteEntityAction.php
```diff
@@ -43,6 +44,13 @@ protected function initialize(array $data)
      */
     public function execute(array $data = [])
     {
+        // If object is not deletable returns 403 FORBIDDEN
+        if ((\BEdita\Core\Model\Table\RolesTable::ADMIN_ROLE === $data['entity']['id'] && $data['entity'] instanceof \BEdita\Core\Model\Entity\Role)
+                ||
+            (\BEdita\Core\Model\Table\UsersTable::ADMIN_USER === $data['entity']['id'] && $data['entity'] instanceof \BEdita\Core\Model\Entity\User)) {
+            throw new ForbiddenException(__d('bedita', 'Could not delete "{0}" 1', $data['entity']));
```
fquffio: Actions are supposed to be a bit more agnostic, IMHO.

I'd prefer to see this model-specific logic in `UsersTable::beforeDelete()` and `RolesTable::beforeDelete()`, and raise a generic exception from there. Then, in `UsersController` and `RolesController` catch that exception and turn it into a specific HTTP error by raising the ForbiddenException.

plugins/BEdita/Core/src/Model/Action/RemoveAssociatedAction.php
```diff
@@ -58,10 +59,20 @@ protected function diff(EntityInterface $entity, array $relatedEntities)
...
      * @throws \RuntimeException Throws an exception if an unsupported association is passed.
      */
     protected function update(EntityInterface $entity, $relatedEntities)
     {
+        if (is_array($relatedEntities)) {
+            $ids = array_keys($relatedEntities);
+            if ((in_array(\BEdita\Core\Model\Table\RolesTable::ADMIN_ROLE, $ids) && $entity instanceof \BEdita\Core\Model\Entity\Role)
+                ||
+                (in_array(\BEdita\Core\Model\Table\UsersTable::ADMIN_USER, $ids) && $entity instanceof \BEdita\Core\Model\Entity\User)) {
+                throw new ForbiddenException(__d('bedita', 'Could not update relationship for users/roles for ADMIN_USER and ADMIN_ROLE'));
```
fquffio: Again, I'd prefer this logic to be moved in `RolesUsersTable::beforeDelete()` and raise an exception from there.

Here it is not only because of "coding style", but it's also because not all cases are covered by this check: an invocation to `SetAssociatedAction` would replace all existing links, so by issuing a request on `PATCH /users/1/relationships/roles` with an empty array of data one would still be able to remove user 1 from administrators.

fquffio (changes requested):

plugins/BEdita/Core/src/Model/Action/DeleteEntityAction.php
didoda: Fine for me, but: `beforeDelete` is not called when using `deleteAll`. Is it ok for us or do we use `deleteAll` sometimes?

stefanorosanelli: we could probably disable `deleteAll` or override it with a condition like `id != 1`?

fquffio: `deleteAll` and `updateAll` must be used with extreme care anyway. No lifecycle event is triggered with `deleteAll` or `updateAll` — not only `Model.beforeDelete` — and therefore system corruption may happen in many other ways: cache not invalidated, counters not updated, permissions not checked, …

So far, APIs do not expose ways to delete multiple objects at once — except for relationships with junction tables, whose rows are actually entities, but those are already treated with proper care — and I'd prefer not to do such in any near future. Then, our concerns should be more focused on shell scripts or application-specific logic.

That having been said, `deleteAll` may come in handy at times. We already use it sometimes in tests, where we want a table to be basically `TRUNCATE`d to test empty responses or ensure that the behavior is correct even when data is missing. So I'd prefer to keep it enabled, and give the developers the responsibility of handling their data with proper care. The system is not going to be completely bulletproof anyway, so if somebody _really_ wants to screw up their own database, they will eventually succeed. 🙃

plugins/BEdita/Core/tests/TestCase/Model/Table/RolesTableTest.php
```diff
@@ -155,4 +156,23 @@ public function testFindMine()
...
     }
+
+    /**
+     * Test delete admin role
+     *
+     * @return void
+     *
+     * @covers ::beforeDelete
+     */
+    public function testDeleteAdminRole()
```
fquffio: It is generally better to rely on PHPUnit's expected exceptions:

```php
/**
 * ...
 * @expectedException \BEdita\Core\Exception\ImmutableResourceException
 * @expectedExceptionCode 403
 * @expectedExceptionMessage Could not delete "role" 1
 */
public function testDeleteAdminRole()
{
    $role = $this->Roles->get(RolesTable::ADMIN_ROLE);
    $this->Roles->delete($role);
}
```

This should do the work with slightly less code. 😃

plugins/BEdita/Core/tests/TestCase/Model/Table/RolesTableTest.php
```diff
@@ -155,4 +156,23 @@ public function testFindMine()
...
+    {
+        $roles = $this->Roles->get(\BEdita\Core\Model\Table\RolesTable::ADMIN_ROLE);
```
fquffio: For namespaced classes we usually `use` them at the top of the file to avoid having long class names in the code. `use`s should be in alphabetical order.

For classes in the topmost namespace (i.e. `\Exception`) we accept both solutions, if I remember well.

plugins/BEdita/Core/src/Model/Table/RolesTable.php
```diff
@@ -124,4 +127,19 @@ protected function findMine(Query $query)
...
+     *
+     * @param \Cake\Event\Event $event The beforeSave event that was fired
+     * @param \Cake\Datasource\EntityInterface $entity the entity that is going to be saved
+     * @return void
+     * @throws \BEdita\Core\Exception\ImmutableResourceException if entity is not deletable
+     */
+    public function beforeDelete(Event $event, EntityInterface $entity)
+    {
+        if (static::ADMIN_ROLE === $entity->id) {
+            throw new ImmutableResourceException(__d('bedita', 'Could not delete "{0}" {1}', $entity, $entity->id));
```
fquffio: `$entity` is an object. Is it stringified correctly? 🤔

fquffio (changes requested):

plugins/BEdita/Core/src/Model/Table/RolesTable.php
didoda: Now we don't need dynamic value... In `RolesTable` entity is `Role`.

stefanorosanelli (approved):

fquffio (approved):

PR #1330, opened 2017-08-24
GET & PATCH /auth/user

### LINKED ISSUE #1298: Method to modify user own data
- `PATCH /auth/user` to modify user profile data, excluding `username`, `password` and `email`

- `GET /auth/user` instead of `GET /auth` to read user profile data

stefanorosanelli:
This PR fixes #1298

* `GET /auth/user` is now default endpoint to retrieve authenticated user data - `GET /auth` still working but deprecated

* `PATCH /auth/user` allows modification of user logged data with the exception of: `username`, `password`, `email` - not modifiable by logged user on this endpoint

Since no selection of  `type` or `id` is possible in this case and for consistency with other `LoginController` methods body data request format is plain JSON and not JSON API. 
This can be changed of course...

### DIFF
--- a/plugins/BEdita/API/src/Controller/LoginController.php
+++ b/plugins/BEdita/API/src/Controller/LoginController.php
@@ -15,8 +15,10 @@ namespace BEdita\API\Controller;
 
 use BEdita\Core\Model\Action\ChangeCredentialsAction;
 use BEdita\Core\Model\Action\ChangeCredentialsRequestAction;
+use BEdita\Core\Model\Action\SaveEntityAction;
 use Cake\Controller\Component\AuthComponent;
 use Cake\Core\Configure;
+use Cake\Network\Exception\BadRequestException;
 use Cake\Network\Exception\UnauthorizedException;
 use Cake\ORM\TableRegistry;
 use Cake\Routing\Router;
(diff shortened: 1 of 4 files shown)


plugins/BEdita/API/src/Controller/LoginController.php
```diff
@@ -167,21 +169,59 @@ protected function jwtTokens(array $user)
...
+        $this->request->allowMethod('patch');
+
+        $notAllowed = ['username', 'password', 'password_hash', 'email'];
+        $data = $this->request->getData();
+        if (!empty(array_intersect($notAllowed, array_keys($data)))) {
+            throw new BadRequestException([
+                'title' => __d('bedita', 'Bad input data'),
+                'detail' => sprintf('Fields not allowed: %s', implode(', ', $notAllowed)),
+            ]);
+        }
```
fquffio: Instead of this approach, I'd prefer to use entity's accessible fields. So, `$entity->setAccessible('username', false);` and so on. However, no bad request exception would be thrown in case a user tries to update one of those fields, and their values would be silently discarded. This is consistent with the behaviour in other parts of the application (i.e.: `PATCH /objects/123` with a non-accessible field), but if we want an error to be raised in this case we could move this logic to the `SaveEntityAction`.

We might evaluate this refactor in the future, though.

fquffio (approved):

plugins/BEdita/API/src/Controller/LoginController.php
stefanorosanelli: That's right but this behavior is `enpoint` dependent: from `/users` we should be able to modify those properties, but from `/auth/user` not. 

As whether or not throw an exception: generally I'd prefer the *robustness principle* and simply ignore, but in this case since the request is potentially harmful giving an error response is not a bad thing.

fquffio: Yeah, I meant that those properties could have been made not accessible in the controller on the fly by calling `setAccess()` on the entity returned by `userEntity()`. Therefore those properties would still be accessible in other contexts, but not here.

stefanorosanelli: Ok, we could use entity `setAccess` now and handle exception raise in `SaveEntityAction` in a future refactor?

fquffio: Yes. But if raising an error in this case is a requirement, we might just leave it as it is and take note of a possible future refactor.

stefanorosanelli: It's not a strong requirement actually. I will use `setAccess`

We should probably define different validation behaviors on properties access: some accesses could be ignored, some others could be considered dangerous.

stefanorosanelli: Problem using  `$entity->setAccess(['username', 'password', 'password_hash', 'email'], false);` is that on the response we have `username` and `email` in `meta` instead of `attributes`... :thinking: ...

stefanorosanelli: solved with `$entity = $this->userEntity()` after `$action`

PR #1338, opened 2017-08-30
Add Docker Compose template

### LINKED ISSUE #1312: Docker compose
* [x] docker image for **bedita/manager**
 * [ ] docker compose file

fquffio:
This PR fixes #1312.

A `docker-compose.yml` template is added at the root of the repository. By calling `docker-compose up`, a BEdita environment will be launched. The environment consists of a MySQL 5.7 server container, an API container, and a manager container.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,36 @@
+---
+version: "2.1"
+
+volumes:
+  database:
+  files:
+services:
+  manager:
+    image: "bedita/manager"
+    depends_on:
+      - api
+    ports:
+      - "${APP_PORT:-5000}:5000"
+    environment:
+      BEDITA_API: "http://localhost:${API_PORT:-8089}"
+  api:
+    image: "bedita/bedita:4-cactus"
+    depends_on:
+      - database
+    links:
+      - database
+    ports:
+      - "${API_PORT:-8089}:80"
+    volumes:
+      - "files:/var/www/webroot/files"
+    environment:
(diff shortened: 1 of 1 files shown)


stefanorosanelli:
I get this error
```
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a version of "2" (or "2.0") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```

with:
```
$ docker --version
Docker version 17.05.0-ce, build 89658be
$ docker-compose --version
docker-compose version 1.8.0, build unknown
```

fquffio:
@stefanorosanelli Your Docker version seem to be fairly recent, but your Docker Compose version might be a little outdated.

```
$ docker -v
Docker version 17.06.0-ce, build 02c1d87
$ docker-compose -v
docker-compose version 1.14.0, build c7bdf9e
```

Docker Compose version 2.1 is required in order to provide default values for template's environment variables. For example, this `${API_PORT:-8089}` in version 2.0 is unsupported, and only `${API_PORT}` can be specified. This means that if an environment variable `API_PORT` is not set, the command `docker-compose` would fail instead of falling back on default port 8089.

stefanorosanelli:
Ok, we need to specify minimum `docker` and `docker-compose` versions.

updated to `docker-compose version 1.14.0, build c7bdf9e` it works now

PR #1343, opened 2017-09-02
Write operations on custom properties

### LINKED ISSUE #1340: Write operations on `custom` properties
This issue is part of  #1123 and follows #1327 

 On every POST/PATCH operation on objects:
 * look for object types custom properties in `attributes` - see `/object_types/{id}/properties`
 * those `attribute` must be saved in `objects.custom_props` JSON field

JSON-SCHEMA validation will be done in a future issue

fquffio:
This PR resolves #1340.

Custom properties are now gently moved to `custom_props` when the entity is saved.

A new `ObjectTypeBehavior` was introduced as a place to share common logic for `RelationsBehavior` and `CustomPropertiesBehavior` — and possibly more behaviours to come.

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
@@ -225,6 +225,8 @@ class ObjectsControllerTest extends IntegrationTestCase
                         'lang' => 'eng',
                         'publish_start' => null,
                         'publish_end' => null,
+                        'another_username' => 'synapse', // custom property
+                        'another_email' => 'synapse@example.org', // custom property
                     ],
                     'meta' => [
                         'locked' => false,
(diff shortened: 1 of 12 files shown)


stefanorosanelli (approved):

didoda (approved):

batopa (approved):
Just a consideration. Currently if the object type "profiles" has a custom property "x" saving "users" we can inherit that custom property, i.e.

```php
$users = TableRegistry::get('Users');
$u = $users->newEntity([
    'username' => 'xxx',
    'password' => 'yyy',
    'x' => 'property defined in profiles type',
]);
$users->save($u);
```

In `objects.custom_props` we'll have the `x` property also if it's not directly  associated to `users` object type.

We have to consider if it is a wanted behavior or if we don't want inheritance in custom props.

I'm going to open an issue as a reminder of this situation.

PR #1344, opened 2017-09-04
#963 feat: object reachable check, force skip if necessary

### LINKED ISSUE #963: [API] Make concept of "object reachable" configurable
Currently an object is reachable via API if it's on publication's tree or is related to an object on publication's tree.

We need a way to decide what means "reachable" for a specific frontend API. 
For example in some situation I could want to permit to get or save objects without the need to check those conditions.

didoda:
This PR fixes #963

This gives the opportunity to let objects not in tree be "visible" in `getObjects`.
Force skip reachable check in `Controller`, i.e.:

```
$this->ApiValidator->setReachableCheck(false);
call_user_func_array(array($this, getObjects'), func_get_args());
```

### DIFF
--- a/bedita-app/controllers/components/api_validator.php
+++ b/bedita-app/controllers/components/api_validator.php
@@ -1068,6 +1068,12 @@ class ApiValidatorComponent extends Object {
         return false;
     }
 
+    /**
+     * Set check for reachable object (default true: check it)
+     *
+     * @param boolean $check
+     * @return void
+     */
     public function setReachableCheck($check = true) {
         $this->enableObjectReachableCheck = $check;
     }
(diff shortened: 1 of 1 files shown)


bedita-app/controllers/components/api_validator.php
```diff
@@ -1057,4 +1067,8 @@ public function isMimeTypeValid($mimeType, $objectType) {
 
         return false;
     }
+
+    public function setReachableCheck($check = true) {
```
batopa: Missing doc block

batopa (changes requested):

batopa (approved):

PR #1347, opened 2017-09-18
Bake BE4 plugin task

stefanorosanelli:
This PR adds a new task to `bake` a new BEdita plugin via

```bash
bin/cake bake bedita_plugin MyPlugin
```

This will create an empty plugin structure ready for use as BEdita plugin with some simple customizations of regular cake plugins.

Plugin is not automatically loaded in bootstrap like classic plugins, but needs to be included in `Plugins` config.


src/Shell/Task/BeditaPluginTask.php
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * The BEdita Plugin Task handles creating an empty BEdita4 plugin, ready to be used
+ */
+class BeditaPluginTask extends BakeTask
+{
+    /**
+     * {@inheritDoc}
+     */
+    protected function _modifyBootstrap($plugin, $hasAutoloader)
+    {
```
fquffio, line 15: I'd add a line of comment to explain why this method is being overwritten empty. Something like `// Avoid updating bootstrap.php` would be just fine IMHO.

fquffio (approved):
Ok for me, but wouldn't this template fit better in the DevTools plugin? This is something that's meant for development purposes solely, right?

stefanorosanelli:
Yes, we could move to `DevTools`.
I'll create a new PR in that repo, then I will remove this one

stefanorosanelli:
Close as obsolete
Moved to bedita/dev-tools#4

Issue #1348, opened 2017-09-20
`/model` base path for modeling endpoints


stefanorosanelli:
An endpoint refactor is proposed on the BE4 Model API: all modeling related endpoints will have `/model        ' as base path, we will then have:

-  `/model/object_types`
- `/model/relations`
- `/model/properties`
- `/model/property_types`

This will help giving consistent permissions to operations involving modeling only to `/model` base endpoint

PR #1350, opened 2017-09-21
Introduce /model base endpoint

stefanorosanelli:
This PR solves #1348 

All *modeling* related endpoints are moved to `/model`, namely 
-  `/model/object_types`
- `/model/relations`
- `/model/properties`

Main code changes:
 * new routing URLs, pure resources urls are splitted in two groupd
 * `JsonApiModelTrait` introduced, similar to `JsonApiAdminTrait`
 * resource controller attribute `$_routeNamePrefix` added

Bonus track:
- `/model/property_types` was introduced

### DIFF
--- a/plugins/BEdita/API/src/Controller/AdminController.php
+++ b/plugins/BEdita/API/src/Controller/AdminController.php
@@ -32,6 +32,11 @@ class AdminController extends ResourcesController
     public $modelClass = null;
 
     /**
+     * {@inheritDoc}
+     */
+    protected $_routeNamePrefix = 'api:admin';
+
+    /**
      * Resource name, one of `$this->config('allowedResources')`
      */
     public $resourceName = null;
@@ -76,7 +81,7 @@ class AdminController extends ResourcesController
     {
         return Router::url(
             [
-                '_name' => 'api:admin:resource',
+                '_name' => $this->_routeNamePrefix . ':resource',
                 'item' => $this->resourceName,
                 'id' => $id,
                 'controller' => $this->name,
--- a/plugins/BEdita/API/src/Controller/AdminController.php
+++ b/plugins/BEdita/API/src/Controller/AdminController.php
@@ -34,7 +34,7 @@ class AdminController extends ResourcesController
     /**
      * {@inheritDoc}
      */
-    protected $_routeNamePrefix = 'api:admin';
+    protected $routeNamePrefix = 'api:admin';
 
     /**
      * Resource name, one of `$this->config('allowedResources')`
@@ -81,7 +81,7 @@ class AdminController extends ResourcesController
     {
         return Router::url(
             [
-                '_name' => $this->_routeNamePrefix . ':resource',
+                '_name' => $this->routeNamePrefix . ':resource',
                 'item' => $this->resourceName,
                 'id' => $id,
                 'controller' => $this->name,
--- a/plugins/BEdita/API/src/Controller/ObjectTypesController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectTypesController.php
@@ -29,7 +29,7 @@ class ObjectTypesController extends ResourcesController
     /**
      * {@inheritDoc}
      */
-    protected $_routeNamePrefix = 'api:model';
+    protected $routeNamePrefix = 'api:model';
 
     /**
      * {@inheritDoc}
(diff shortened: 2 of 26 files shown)


plugins/BEdita/API/src/Controller/ObjectTypesController.php
```diff
@@ -13,7 +13,7 @@
 namespace BEdita\API\Controller;
 
 /**
- * Controller for `/object_types` endpoint.
+ * Controller for `/model/model/object_types` endpoint.
```
didoda: should it be instead `/model/object_types`?

didoda (approved):

plugins/BEdita/API/src/Controller/ObjectTypesController.php
stefanorosanelli: of course, it's a typo...

plugins/BEdita/API/src/Controller/AdminController.php
```diff
@@ -32,6 +32,11 @@ class AdminController extends ResourcesController
     public $modelClass = null;
 
     /**
+     * {@inheritDoc}
+     */
+    protected $_routeNamePrefix = 'api:admin';
```
batopa: why not `protected $routeNamePrefix`? In general we avoid to use `_` except for CakePHP things

stefanorosanelli: yes, you're right

stefanorosanelli: done

plugins/BEdita/Core/src/Model/Entity/JsonApiAdminTrait.php
```diff
@@ -27,13 +27,23 @@
...
+     */
+    protected function routeNamePrefix()
+    {
+        return 'api:admin';
+    }
+
+    /**
+     * {@inheritDoc}
      */
     protected function getLinks()
```
fquffio, line 41: Do we still need to override this method?

fquffio (commented):
Seems OK. However, I'd use [routing prefixes](https://book.cakephp.org/3.0/en/development/routing.html#prefix-routing) for the purposes of this PR.

This means that:
- `/model` routes should be connected within a `$routes->prefix('model', ['_namePrefix' => 'model:'], function (RouteBuilder $routes) { /* ... */ });` instruction
- `ObjectTypesController`, `PropertiesController`, ..., should be moved to a `\BEdita\API\Controller\Model` dedicated namespace.

Changing to this other approach should be pretty straightforward. However, since I'd use a similar approach for `admin/` routes as well, we could probably consider making this refactor in a dedicated issue.

stefanorosanelli:
Agreed. We could move both `/admin` and `/model` to dedicated namespaces and routing prefixes in a separate issue or PR.

plugins/BEdita/Core/src/Model/Entity/JsonApiAdminTrait.php
fquffio: Giving answer to my own question: yes, we do.

plugins/BEdita/Core/src/Model/Entity/JsonApiModelTrait.php
```diff
@@ -0,0 +1,37 @@
...
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
+namespace BEdita\Core\Model\Entity;
+
+use Cake\ORM\TableRegistry;
+use Cake\Routing\Router;
```
fquffio: These imports are unused.

plugins/BEdita/Core/src/Model/Entity/JsonApiAdminTrait.php
stefanorosanelli: Yes, given the current routing structure. We could change it in a new PR/issue though

plugins/BEdita/Core/src/Model/Entity/JsonApiModelTrait.php
stefanorosanelli: Right, updated

PR #1351, opened 2017-09-25
CTI: pass _inherited option in beforeSave chain

batopa:
This PR is **low priority** and is a **proposal** to know if `beforeSave` listener is called on main table or on inherited table in CTI scenario.

If you think that is useless feel free to discard it. :sweat_smile:

### DIFF
--- a/plugins/BEdita/Core/tests/TestCase/ORM/Inheritance/InheritanceEventHandlerTest.php
+++ b/plugins/BEdita/Core/tests/TestCase/ORM/Inheritance/InheritanceEventHandlerTest.php
@@ -319,10 +319,12 @@ class InheritanceEventHandlerTest extends TestCase
      */
     public function testBeforeSaveOptions()
     {
+        $eventDispatchedFelines = $eventDispatchedMammals = $eventDispatchedAnimals = 0;
         // Main table
         $this->fakeFelines->eventManager()->on(
             'Model.beforeSave',
-            function (Event $event, EntityInterface $entity, \ArrayObject $options) {
+            function (Event $event, EntityInterface $entity, \ArrayObject $options) use (&$eventDispatchedFelines) {
+                $eventDispatchedFelines++;
                 static::assertArrayNotHasKey('_inherited', $options);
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/tests/TestCase/ORM/Inheritance/InheritanceEventHandlerTest.php
```diff
@@ -309,6 +311,53 @@ public function testBeforeSaveFailure()
...
+        );
+
+        $feline = $this->fakeFelines->newEntity([
+            'name' => 'Gastone',
+            'legs' => 4,
+            'subclass' => 'Lucky pets',
+            'family' => 'Cats',
+        ]);
+        $result = $this->fakeFelines->save($feline);
+    }
```
fquffio: If for some reason some of the events being listened were not dispatched, this test would not fail. Adding something like this would be enough:

```php
$eventDispatchedFelines = $eventDispatchedMammals = /* ... */ = 0;

$this->fakeFelines->eventManager()->on(
    'Model.beforeSave',
    function (Event $event, EntityInterface $entity, \ArrayObject $options) use (&$eventDispatchedFelines) {
        $eventDispatchedFelines++;
        // ...
    }
);

// ...

$result = $this->fakeFelines->save($feline);

static::assertSame(1, $eventDispatchedFelines);
static::assertSame(1, $eventDispatchedMammals);
// ...
```

fquffio (approved):
Seems useful.

plugins/BEdita/Core/tests/TestCase/ORM/Inheritance/InheritanceEventHandlerTest.php
batopa: You are right.

batopa: done

stefanorosanelli (approved):

PR #1354, opened 2017-09-28
Makefile: release, publish and repo split

stefanorosanelli:
This PR introduces a Makefile, simply as a fork of CakePHP Makefile, to create releases, tags, publish dist zip files and to split main repo in two read-only components repos:

 * `bedita/core` for `plugins/BEdita/Core`
 * `bedita/api` for `plugins/BEdita/API`

Repos have not yet been created: are names ok or do we prefer something like `bedita/core-plugin` ?

To do some tests:
 * use a fork, and change owner in `OWNER=####`
 * create empty component repos  `{owner}/core` and `{owner}/api`
 * set env variables `GITHUB_USER` and optionally `GITHUB_TOKEN`
 * run `make release VERSION=4.0.0-beta`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,199 @@
+# The following env variables need to be set:
+# - VERSION
+# - GITHUB_USER
+# - GITHUB_TOKEN (optional if you have two factor authentication in github)
+
+# Use the version number to figure out if the release
+# is a pre-release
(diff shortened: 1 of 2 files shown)


fquffio (approved):

fquffio:
For me `bedita/core` and `bedita/api` are ok. We might want to prefix other optional plugins with `bedita/plugin-` or something like that, but for these two I would prefer the short form. 😃

didoda (approved):

PR #1355, opened 2017-10-06
Multi project urls on same instance using BEDITA_BASE_URL

stefanorosanelli:
This PR adds URL basic multi project capability on BE4 API

Custom environments may be set in optional `config/environment.php` using whatever logic you like.
This file is loaded on bootstrap if found.

A new env var `BEDITA_BASE_URL` is introduced if you want to differentiate projects with base urls.

Usually you will select a project based on some URL rules or via another environment variable on CLI.
Then you may set environment variables accordingly. 
Normally you will need to define:
 * DATABASE_URL
 * CACHE_CAKEMODEL_URL
 * CACHE_BEDITAOBJECTTYPES_URL
 * BEDITA_BASE_URL

Still failing: redirect from `/` to `/home` is not working as `BEDITA_BASE_URL` is set


config/app.default.php
```diff
@@ -395,7 +395,7 @@
      * - `autoload` - boolean - (default: false) Whether or not you want an autoloader registered
      */
     'Plugins' => [
-        'BEdita/DevTools' => ['debugOnly' => true, 'bootstrap' => true],
+      'BEdita/DevTools' => ['debugOnly' => true, 'bootstrap' => true],
```
fquffio: Wrong indentation. 🙃  It used to be correct before.

fquffio (commented):
Ok, but… `config/preprocess.php` is ignored. Should a `config/preprocess.sample.php` be versioned instead?

config/app.default.php
stefanorosanelli: Yes, actual reason is: ti was commented and decommented before commit.

But I think it should be commented as default: if composer `require-dev` section is not loaded, this ends up in actual `config/app.php` but no `BEdita/DevTools` plugin is present that case

stefanorosanelli:
problem with `config/preprocess.sample.php` is that it can vary a lot, I cannot think of a reasonable default.

what about `config/environment.php` instead of `preprocess` ?

we may also have a  `config/environment-cli.php` that can be loaded from `config/bootstrap.php` if file is present and CLI is used

or same `config/environment.php` for CLI also ?

stefanorosanelli:
@fquffio I've changed things a little bit, so please have a look at changed files again... sorry

fquffio (approved):
Ok for me, but if we just want to set environment variables it would probably be slightly better to stick to "dotenv" practice.

If we add `josegonzalez/dotenv` to Composer dependencies and uncomment the block of code in `config/bootstrap.php` just above the lines you added, we should be ready to go. You can have a look at `config/.env.default` file to see how it would look like — pretty much the same as `config/environment.sample.php`, but more BASH-y.

This is something that (AFAIK) is quite common across several programming languages and is supposed to be used in development only — in production env vars should be set by other means.

Of course, this does not allow for dynamic processing that a programming language allows: only basic string interpolation is supported.

stefanorosanelli:
Good points. I did have a loog at a `dotenv` like solution. The code your referring to is actually already commented :wink: in `bootstrap.php`

What we need here is a simple mechanism to set env vars dynamically: this could also be done outside of this code, from a vhst configuration for instance or by some other service. 

So a simple strategy, with a completely custom and unpredictable implementations, seemed to me to invoke an `environment.php` file.

We could have both: `environment.php` default implementation, could be a `dotenv` one (`environment.sample.php`) but you can have a completely different ruleset.

PR #1356, opened 2017-10-17
Object types inheritance

### LINKED ISSUE #908: Inheritance in object types design
Defining new object types we should be able to identify a parent object type, inheriting all properties 
Since this definitions are not subject to change over time we don't need tree-like manipulations structure but a simple `parent_type` like field could be enough

Parent types could be limited to `core` types and not include `user defined` types?

Properties inheritance will be done in #1345 and relations inheritance in #1349 

Outcomes of this issue:
 * ability to set and show object parent type
 * introduce the concept of **abstract** object type, like `objects` and `media` (see #1317)
 * expected behavior of a `/base_type` endpoint: should all subtypes be displayed? Should this happen only for `abstract` types? Should this behavior be _configurable_? Or activated with a filter?

fquffio:
This PR resolves #908 by adding inheritance to object types.

Object types can now be abstract and can inherit from a parent object type. For example, "documents" is now a concrete object type that inherits from abstract object type "objects", which should be the base for all object types. Inheritance can be deep.

Abstract object types cannot be instantiated (i.e. an object cannot be of an abstract object type), but for the purposes of search they return results of all their descendants. Concrete object types can be extended as well (e.g.: "users" extends "profiles"), but search will not cascade to child object types (i.e.: `GET /profiles` and `GET /objects?filter[type]=profiles` will **NOT** return users any more).

This PR includes an important schema migration. Be sure to run `bin/cake migrations migrate -p BEdita/Core` and `bin/cake cache clear_all` before testing.

----------

This functionality is far from being complete. Informations about an object type's parents and descendants should be added to cache, and more tests should be added. However, since many files have already been modified (many are test cases, that had to be updated due to object types IDs being changed) I thought it was better to split it in multiple PRs and issues.


stefanorosanelli:
PR seems fine overall but I'm not sure about differentiating behavior between abstract and concrete types in `GET /objects` like queries.

`GET /locations` should return `venues` and `restaurants` for instance? Probably `locations` should become an abstract type.? Some types may become `abstract` in other projects.

Or we could simply limit inheritance to abstract types - like graphql does with interfaces.

Put it this way we could probably change `profiles`/ `users` inheritance schema: `users` are `objects` not subtypes of `profiles` - they simply share table data. So `GET /profiles` and `GET /users` will never return mixed data.

PS: our use of abstract types actually matches graphql interfaces http://graphql.org/learn/schema/#interfaces - this is a big plus.

stefanorosanelli (approved):

batopa (approved):

PR #1358, opened 2017-10-18
Object types naming rules

stefanorosanelli:
This PR introduces some addtional rules on object types naming:

- a list of reserved names is provided - `plugins/BEdita/Core/config/reserved.php` now very short, expected to grow - that cannot be used as objcet type names
- this list may be used in other use cases - like property names
- plural `name` and `singular` must be different - since plural `name` form is the preferred one as a consequence we may use `media_item` and `news_item` as singular form for `media` and `news`

Differentiating `name` and `singular` enables, among other things, unambiguous different root types in *graphql* for singular and plural forms.
 
Code:
- `NotReserved` and `ObjectTypesValidator` class introduced


plugins/BEdita/API/tests/TestCase/Controller/ObjectTypesControllerTest.php
```diff
@@ -584,7 +585,34 @@ public function testAddInvalid()
 
         $this->configRequestHeaders('POST', $this->getUserAuthHeader());
         $this->post('/model/object_types', json_encode(compact('data')));
+        $this->assertResponseCode(400);
+        $this->assertContentType('application/vnd.api+json');
+        $this->assertEquals($count, TableRegistry::get('ObjectTypes')->find()->count());
 
+        // add reserved word failrure
```
batopa: typo: failure

plugins/BEdita/Core/src/Model/Validation/ObjectTypesValidator.php
```diff
@@ -0,0 +1,83 @@
...
+            ->notEmpty('singular')
+            ->add('singular', 'notSameAs', [
+                'rule' => function ($value, $context) {
+                    if (empty($context['data']['name'])) {
+                        return true;
+                    }
+
+                    return ($context['data']['name'] !== $value);
+                },
+                'message' => 'Name and singular fields must be different'
```
batopa: Should we use `__d('bedita', 'Name and singular fields must be different')` for string translations?

plugins/BEdita/Core/tests/TestCase/Model/Validation/NotReservedTest.php
```diff
@@ -0,0 +1,67 @@
...
+                false,
+            ],
+        ];
+    }
+
+    /**
+     * Test reserved rule.
+     *
+     * @param string|null $expected Expected alias, or `null`.
+     * @param string $alias Alias to search for.
```
batopa: the `@param`s are wrong

batopa (changes requested):

batopa (approved):

plugins/BEdita/API/tests/TestCase/Controller/ObjectTypesControllerTest.php
```diff
@@ -584,7 +585,34 @@ public function testAddInvalid()
 
         $this->configRequestHeaders('POST', $this->getUserAuthHeader());
         $this->post('/model/object_types', json_encode(compact('data')));
+        $this->assertResponseCode(400);
+        $this->assertContentType('application/vnd.api+json');
+        $this->assertEquals($count, TableRegistry::get('ObjectTypes')->find()->count());
 
+        // add reserved word failure
```
fquffio: I'd prefer to have this test split into its own test case, to better distinguish possible failures.

fquffio (changes requested):

fquffio (approved):

PR #1361, opened 2017-10-18
Add migrations for `videos`, `audio` and `files`

### LINKED ISSUE #1317: Core media object types
Implement core media object types like images, videos, audio and files.

Is `media` an abstract base type like `objects`?

stefanorosanelli:
This PR fixes #1317 

Core media types are simply introduced through migrations.

Just like that: no fixtures and tests were added, I'm too lazy today :wink: 
But seriously, do we need them here?


fquffio (approved):

fquffio:
Fixture update would probably be helpful, but tons of tests would need to be updated in that case... so I’d add them in the future if and when we’ll really need those — YAGNI.

PR #1362, opened 2017-10-19
Basic analytics middleware

### LINKED ISSUE #1315: Basic analytics
Implement basic analytics data save on persistent storage (initially also simple log files)

stefanorosanelli:
This PR fixes #1315

Basic analytics system via middleware has been implemented.

Analytics data are simply stored as default in `logs/analytics.log` using a JSON encoded format.
Destination may be changed using `Log::config('analytics')` settings.

Created JSON provides this structure:
```json
{
    "r" : "request timestamp",
    "a" : "application id or null",
    "usr" :  "logged user id or null",
    "m" :  "HTTP method",
    "url" : "requeste URL",
    "q" : "query string",
    "s" : "HTTP response status code",
    "c" : "application error code or null",
    "x" : "custom data via events",
    "e" : "elapsed time"
}
``` 

Custom data may be added from plugins implementing event listeners for **'Analytics.custom'** event.
Event method expected signature is `(Event $e, ServerRequestInterface $request, ResponseInterface $response)` 
Returned data will then be added to `"x"` array.

### DIFF
--- a/plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
+++ b/plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * BEdita, API-first content management framework
- * Copyright 2016 ChannelWeb Srl, Chialab Srl
+ * Copyright 2017 ChannelWeb Srl, Chialab Srl
  *
  * This file is part of BEdita: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
@@ -12,12 +12,14 @@
  */
 namespace BEdita\API\Middleware;
 
+use BEdita\Core\State\CurrentApplication;
+use BEdita\Core\Utility\LoggedUser;
+use Cake\Core\Configure;
 use Cake\Http\Response;
 use Cake\Log\Log;
-use Cake\Log\LogTrait;
-use Cake\Network\CorsBuilder;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use Throwable;
 
 /**
  * Middleware to trace analytics data
@@ -26,14 +28,26 @@ use Psr\Http\Message\ServerRequestInterface;
  */
 class AnalyticsMiddleware
 {
-    use LogTrait;
-
     /**
      * Request start time
      *
      * @var float
      */
     protected $startTime = null;
+
+    /**
+     * Analytics data
+     *
+     * @var array
+     */
+    protected $data = [];
+
+    /**
+     * Registered callbacks fo analytics extensions
+     *
+     * @var array
+     */
+    protected static $callbacks = [];
 
     /**
      * Configure analytics logger if not configured yet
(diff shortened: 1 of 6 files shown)


plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Analytics data
+     *
+     * @var array
+     */
+    protected $data = [];
+
+    /**
+     * Registered callbacks fo analytics extensions
```
batopa: typo: for

plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
```diff
@@ -0,0 +1,170 @@
...
+namespace BEdita\API\Middleware;
+
+use BEdita\Core\State\CurrentApplication;
+use BEdita\Core\Utility\LoggedUser;
+use Cake\Core\Configure;
+use Cake\Http\Response;
+use Cake\Log\Log;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Throwable;
```
batopa: `Throwable` has been introduced in PHP 7 but at the moment we support 5.6 too

plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
```diff
@@ -0,0 +1,170 @@
...
+     * @param callable $next The next middleware to call.
+     * @return \Psr\Http\Message\ResponseInterface A response.
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next)
+    {
+        $response = $next($request, $response);
+
+        /**
+         * TODO: add custom application error code
+         */
```
batopa: I suggest to get the application error reading the response body just in case of HTTP status code `>= 400`

batopa (changes requested):

plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
fquffio: That's the ghetto way, bro! 😜

plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
```diff
@@ -0,0 +1,170 @@
...
+        $this->data = [
+            'r' => $request->getEnv('REQUEST_TIME'),
+            'a' => CurrentApplication::getApplicationId(),
+            'usr' => LoggedUser::id(),
+            'm' => $request->getMethod(),
+            'url' => $request->getUri()->getPath(),
+            'q' => $request->getUri()->getQuery(),
+            's' => $response->getStatusCode(),
+            'x' => $this->readCallBackData($request, $response),
+            'e' => round(microtime(true) - $this->startTime, 4, PHP_ROUND_HALF_EVEN),
```
fquffio: If this delta-time computation is made here, it is not completely clear (at least to me) whether execution times from things like `$this->readCallBackData()` are taken into account — it is an implementation detail the order in which the interpreter processes the instructions in lines 155-163.

To clear out clouds I'd suggest to either compute the total execution time _before_ (at line 153), or _after_ (at line 165) then adding it to `$this->data`.

-------------------

By the way, the interpreter seems to process instructions in the most intuitive order:

```php
<?php

function a()
{
    echo 'A' . PHP_EOL;

    return 'A';
}

function b()
{
    echo 'B' . PHP_EOL;

    return 'B';
}

$data = [
    'a' => a(),
    'b' => b(),
];

var_dump($data);

/****** OUTPUT ******
A
B
array(2) {
  ["a"]=>
  string(1) "A"
  ["b"]=>
  string(1) "B"
}
********************/
```

However, I'd prefer to make it as clear as possible.

plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /**
+     * Get analytics custom data from registered callbacks
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Message\ResponseInterface $response The response.
+     * @return array
+     */
+    protected function readCallbackData(ServerRequestInterface $request, ResponseInterface $response)
```
fquffio: Why not dispatching an event instead? By using `EventDispatcherInterface` and `EventDispatcherTrait` we should be able to obtain the same result using a pattern that's already common in the whole application.

Components that are willing to add their own callback could just register themselves as listeners for the event being dispatched and modifying event result.

plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
stefanorosanelli: Agreed, let's make it more clear

plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
stefanorosanelli: Right, I'll give a try

stefanorosanelli:
@fquffio @batopa please have a look again, did some changes after your comments... 
Also PR description was changed with events for custom data.

fquffio (approved):

plugins/BEdita/API/src/Middleware/AnalyticsMiddleware.php
```diff
@@ -0,0 +1,153 @@
...
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Message\ResponseInterface $response The response.
+     * @return array
+     */
+    protected function readCustomData(ServerRequestInterface $request, ResponseInterface $response)
+    {
+        $event = $this->dispatchEvent('Analytics.custom', [$request, $response]);
+        if (empty($event->result)) {
+            return [];
```
fquffio, line 99: Just a side note: empty arrays are always encoded by `json_encode` as `[]`. For some parsers this can be troublesome if they expect an object. If we want empty value to be encoded as `{}` we might return `new \stdClass()` instead of the empty array, to be sure that `"x"` is always an object.

batopa (approved):

Issue #1353, opened 2017-09-27
Properties list of an object type


stefanorosanelli:
[This is needed in `bedita/graphql` but it will probably useful in #1345 and in modeling API]

Get a list of properties with their types of an object type.
Structure similar to `/model/properties` or even better like:
 ```php
'description' => [
     'type' => 'string',
     'required' => false,
    ....
]
```
This list should be cached somehow.

**Comment by stefanorosanelli** [comment]
Much of the work has been done in #1363 
Currently you may use a cached version of properties with something like:
```php
        $properties = TableRegistry::get('Properties')->find('objectType', [$name])
            ->cache(sprintf('id_%s_props', $id), '_bedita_object_types_')
            ->toArray();
``` 
where `$name` is the object type name and `$id` its id

Another possibility will be the use of virtual property `ObjectType::properties` that will also provide a cached list of properties.

@batopa @fquffio can we close this?

batopa:
Ok, closing

PR #1363, opened 2017-10-24
Properties inheritance

### LINKED ISSUE #1345: Inheritance in custom properties
As described in https://github.com/bedita/bedita/pull/1343#pullrequestreview-60596659 currently a object type that extends another type inherit custom properties.

For example given `users` extension of `profiles` extension of `objects` we can define a property on `profiles` type and save it via `users`.

```php
$users = TableRegistry::get('Users');
$u = $users->newEntity([
    'username' => 'xxx',
    'password' => 'yyy',
    'profiles_property' => 'property defined in profiles type',
]);
$users->save($u);
```

In `objects.custom_props` we'll have `profiles_property` also if it's not directly associated to users object type.

Is that the expected behavior or we should isolate custom properties by object type?

fquffio:
This PR fixes #1345 and partially solves #1353.

### Description of changes

Custom properties are inherited using **object type inheritance** introduced in #1356. It has also been implemented a basic description of static properties, that are reflected from current database schema and makes an attempt to represent **class table inheritance** in terms of object type inheritance.

This operates under the assumption that, if an object type B extends an object type A, at some point in the CTI inheritance chain of B, there will be the table(s) in the CTI inheritance chain of A. In other words:

> B inherits from A as object types implies
> Table(A) ∈ { T | Table(B) inherits from T as database tables }

This is a totally reasonable assumption, as far as I can see, however _no enforcement is made_ to ensure this is always true. If it weren't true, bad things could happen, just like Simone and Giuliano teaming up together, `@chialab/dna` becoming a fork of `react`, or Gustavo getting out of the toilet.

**Bonus track**: in order to give static columns a consistent ID that does not conflict with custom properties IDs, a generator of [UUID version 5](https://tools.ietf.org/html/rfc4122#section-4.1.3) has been introduced. UUID version 5 are like UUID version 4, but instead of being generated randomly, they are generated based on a namespace (that must be a valid UUID — we're using the "nil" UUID) and a name (column name in `table.column` notation) using SHA1 hashes. Algorithm has been derived from [this library](https://github.com/lootils/uuid), and a proper Copyright notice has been placed where appropriate.

### API changes

1. Two new filters have been added to `/properties` endpoint that can be used:
    - `filter[object_type]`: this filter accepts an object type name/ID and it will filter properties by object type taking inheritance into account.
    - `filter[type]`: this filter accepts either `static`, `dynamic` or `both` as its only option and it will filter the list of available properties by their type.
2. `CustomPropertiesBehavior` has been updated to use `objectType` filter that takes inheritance into account. This means that a custom property set on object type "Objects" will be available to any object type.

### Caveats

At the present stage, static properties don't have a correct `property_type_id`. Instead, `property_type_id` will be populated with the first ID found in `property_types` table — and yes, if `property_types` is empty an error will occur. A proper detection and mapping of SQL column types to BEdita property types will be implemented in another PR, as this alone will be quite a deal.


didoda (approved):

stefanorosanelli:
Seems ok, good job!
Just one note: difference between `static` and `dynamic` is a low level implementation detail that not all developers/users should be aware of. Wouldn't it be possible to display all properties as default and have `filter[static]` or `filter[dyamic]` to get corresponding subset?

fquffio:
@stefanorosanelli It is possible, actually, but it could only be enabled on all endpoints that show a list of properties (including `GET|POST|PATCH|DELETE /.../relationships/properties`), and I feel like it could cause a bit of confusion.

Let's consider the endpoint `GET /model/object_types/documents/relationships/properties`: this would return a list of both static and dynamic properties including those set on "objects" (i.e.: not directly set on "documents"). However, by calling `PATCH /model/object_types/documents/relationships/properties` with an empty body, only the _dynamic_ properties that are _directly_ set on "documents" would be erased, and the client would be rather surprised to find out that its request to empty the fields for "documents" has only erased a small subset of the resources it could see before.

I find it could be clearer to state that the endpoint `GET /model/properties` is a place where this implementation detail is exposed — and, in general, all `/model/*` endpoints will expose implementation details, one way or another, whether we like it or not.

Higher-level representation of an object type properties can still be added in places where we can safely provide the client with an enriched set of data without causing confusion. Speaking of properties, I'd probably add a brand new endpoint `GET /schema/:objectType` that returns a JSON schema for the requested object type: here properties of all kinds that are applicable to the object type would be flattened out, thus giving to the client a mean to obtain the high-level information it needs.

stefanorosanelli:
@fquffio I don't see any particular problems in having a set of immutable properties that cannot be changed or removed from an object type - some of them can only be hidden using `object_types.hidden` but we can find a better way to do that.
To me the expected behavior is exactly the one you suggested:  `/model/object_types/documents/relationships/properties` returns a list of resources including static and dynamic properties. We probably need a virtual property to make it more clear - static properties cannot be removed or changed, like you cannot remove `user 1` or `role 1` or modify their relationship.

@bedita/core what do you think?

Having a read-only `schema` endpoint seems a very good idea: I'd go with `GET /model/schema/:type` where `:type` may be an object or a resource. This will help us a lot in `bedita/manager`, in the SDK and in other future projects,

batopa:
I would expect that `GET /model/properties` returns all properties (static + dynamic). Via `meta` we can inform the client if the property is `static` (we could use `unchangeable` as for roles) and `hidden`.

Every write operation on `static` properties will return a `403 Forbidden`.

At the same way `GET /model/object_types/documents/relationships/properties` would return all properties and calling `PATCH /model/object_types/documents/relationships/properties` with an empty body will return `403 Forbidden` with appropriate message (and error code?).

With specific filter the client could take only `static`, `dynamic`, `hidden` and `visible`.

PR #1364, opened 2017-10-27
Composer: code sniffer proper setup and launch

stefanorosanelli:
This PR introduces some minor changes to `composer,json`

- new `cs-setup` script to set CakePHP codesniffer rules as default
- new `update-dev` script to launch `composer update` followed by `cs-setup`
- `cs-check` and `cs-fix` have been updated to use same rules as in Travis/CI

To try them simply invoke:
 - `composer run-script cs-setup`
 - `composer run-script update-dev`
and so on...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "bedita/bedita",
     "type": "project",
     "description": "BEdita API-first content management server app",
-    "homepage": "http://bedita.com",
+    "homepage": "https://www.bedita.com",
     "license": "LGPL-3.0",
     "support": {
         "source": "https://github.com/bedita/bedita",
@@ -17,7 +17,7 @@
         {
             "name": "ChannelWeb s.r.l.",
             "email": "info@channelweb.it",
-            "homepage": "http://www.channelweb.it"
+            "homepage": "https://www.channelweb.it"
         }
     ],
     "require": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -53,11 +53,21 @@
...
-        "test": "phpunit --colors=always"
+        "cs-check": " vendor/bin/phpcs --colors -n -p --extensions=php --ignore=/Migrations/,/Seeds/ ./config ./src ./tests ./plugins/*/*/config ./plugins/*/*/src ./plugins/*/*/tests",
+        "cs-fix": "vendor/bin/phpcbf --colors --extensions=php --ignore=/Migrations/,/Seeds/ ./config ./src ./tests ./plugins/*/*/config ./plugins/*/*/src ./plugins/*/*/tests",
+        "cs-setup": [
+            "vendor/bin/phpcs --config-set installed_paths vendor/cakephp/cakephp-codesniffer",
+            "vendor/bin/phpcs --config-set default_standard CakePHP",
+            "vendor/bin/phpcs --config-set colors 1"
+        ],
+        "test": "vendor/bin/phpunit --colors=always",
+        "update-dev": [
```
fquffio, line 67: Could we add `@cs-setup` to `post-install` instead of adding an additional setup command?

composer.json
```diff
@@ -53,11 +53,21 @@
...
-        "cs-fix": "phpcbf --colors --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests",
-        "test": "phpunit --colors=always"
+        "cs-check": " vendor/bin/phpcs --colors -n -p --extensions=php --ignore=/Migrations/,/Seeds/ ./config ./src ./tests ./plugins/*/*/config ./plugins/*/*/src ./plugins/*/*/tests",
+        "cs-fix": "vendor/bin/phpcbf --colors --extensions=php --ignore=/Migrations/,/Seeds/ ./config ./src ./tests ./plugins/*/*/config ./plugins/*/*/src ./plugins/*/*/tests",
+        "cs-setup": [
+            "vendor/bin/phpcs --config-set installed_paths vendor/cakephp/cakephp-codesniffer",
+            "vendor/bin/phpcs --config-set default_standard CakePHP",
+            "vendor/bin/phpcs --config-set colors 1"
+        ],
+        "test": "vendor/bin/phpunit --colors=always",
```
fquffio, line 66: I'm not sure about this, but I think things should work even without `vendor/bin/` prefix…

fquffio (approved):

composer.json
batopa: Uhm... @fquffio  since `phpcs` is in `require-dev` what would happen if the `install` command was called with `--no-dev` option? Is there a way to exec `post-install` only in dev context?

fquffio: Ugh, you're right. I don't know if there is. One could maybe test for `phpcs` command to be available, but it would be more complicated than it's needed, probably. Ignore my comment above. 🙃

batopa: Anyway I think it was a good idea, maybe `post-update-cmd` would be the best place to have phpcs always well configured.
We could try to manage it in `BEdita\App\Console\Installer::postInstall`.
Maybe in another issue :smile:

batopa (approved):

composer.json
stefanorosanelli: Just to make sure vendor version of `phpcs` and `phpunit` is used and to avoid a global configuration change for `phpcs`

composer.json
stefanorosanelli: Initially I wanted to put `@cs-setup` in `post-updated-cmd` but, apart from the `--no-dev` problem mentioned before it's something you need only as a developer.

My proposal here for devs is to use `composer run-script update-dev` instead of `composer update`

didoda (approved):

PR #1365, opened 2017-10-28
Default application and remove/disable controls

stefanorosanelli:
This PR introduces a default application with id 1 via migrations.

Some new constraints are introduced:
- default application and current application in use cannot be removed
- default application and current application cannot be disabled

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/ApplicationsTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/ApplicationsTable.php
@@ -96,7 +96,7 @@ class ApplicationsTable extends Table
     /**
      * Generate the api key on application creation.
      *
-     * If applications is DEFAULT_APPLICATION or current invoking application and `enabled` to false
+     * If applications is DEFAULT_APPLICATION or current invoking application and `enabled` is `false`
      * raise an ImmutableResourceException
      *
      * @param \Cake\Event\Event $event The event dispatched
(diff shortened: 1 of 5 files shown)


fquffio (approved):

plugins/BEdita/Core/src/Model/Table/ApplicationsTable.php
```diff
@@ -85,19 +94,25 @@ public function buildRules(RulesChecker $rules)
...
-     * Generate the api key on application creation
+     * Generate the api key on application creation.
+     *
+     * If applications is DEFAULT_APPLICATION or current invoking application and `enabled` is `false`
+     * raise an ImmutableResourceException
      *
      * @param \Cake\Event\Event $event The event dispatched
      * @param \Cake\Datasource\EntityInterface $entity The entity to save
      * @return void
      */
```
didoda: Some doc about exception, i.e.: `@throws \BEdita\Core\Exception\ImmutableResourceException if entity is immutable`

didoda (approved):

plugins/BEdita/Core/src/Model/Table/ApplicationsTable.php
stefanorosanelli: rigth, done

batopa (approved):

PR #1368, opened 2017-10-31
Simple bash deploy script

stefanorosanelli:
This is a super simple bash script to deploy an instance using git.

Suitable in dev/test environments but **absolutely not** a substitute of more sophisticated strategies necessary in distributed/multi-instance scenarios.

Script may be invoked like `bin/simple_git_deploy.sh` on BEdita root folder, but can be launched from any other location.

Usage: `bin/simple_git_deploy.sh [deploy_user] [app_user] [plugins_list]`

- `deploy_user`: user responsible for git/deploy operations, typically file owner
- `app_user`: user launching API actions, typically web server user
- `plugins_list`: comma separated list of active plugins to update via git together with core

Examples:
- `bin/simple_deploy.sh www-data`
- `bin/simple_deploy.sh bedita www-data BEdita/GraphQL,MyPlugin`

### DIFF
--- a/bin/simple_git_deploy.sh
+++ b/bin/simple_git_deploy.sh
@@ -10,10 +10,12 @@
 #  * `plugins_list`: comma separated list of active plugins to update together with core
 #
 # Notes:
-#  - when no arguments are passed no deploy or app user is set and all plugins found in `plugins/` are updated
+#  - when no arguments are passed no deploy or app user is set and no plugins are updated
 #  - on single argument only `deploy_user` is set, and used as `app_user`` too
-#  - `plugins_list` use relative path inside `plugins/` for names, example: `BEdita/DevTools,MyPlugin`
-#     on empty `plugins list` all plugins found in `plugins/` are updated
+#  - `plugins_list`
+#       * use relative path inside `plugins/` for names, example: `BEdita/DevTools,MyPlugin`
+#       * using `--all-plugins` all plugins found in `plugins/` are updated
+#       * on empty `plugins list` no plugins are updated
 #
 ################################################################################
 
@@ -22,7 +24,7 @@ DEPLOY_PREFIX_CMD=''
 APP_USER=''
 APP_PREFIX_CMD=''
 
-BE4_DIR=$PWD
+BE4_DIR=$(dirname $(cd $(dirname "$0") && pwd))
 if [ $# -eq 0 ]; then
     CURR_USR=`whoami`
     echo "Using current user ($CURR_USR) as deploy and app user"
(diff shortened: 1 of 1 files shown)


bin/simple_git_deploy.sh
```diff
@@ -0,0 +1,113 @@
...
+#     on empty `plugins list` all plugins found in `plugins/` are updated
+#
+################################################################################
+
+DEPLOY_USER=''
+DEPLOY_PREFIX_CMD=''
+APP_USER=''
+APP_PREFIX_CMD=''
+
+BE4_DIR=$PWD
```
fquffio:
```bash
BE4_DIR=$(dirname $(cd $(dirname "$0") && pwd))
```

This apparently works, and removes the requirement of the working directory.

fquffio:
```bash
BE4_DIR=$(dirname $(cd $(dirname "$0") && pwd))
```

This apparently works, and removes the requirement of the working directory.

bin/simple_git_deploy.sh
```diff
@@ -0,0 +1,113 @@
...
+$DEPLOY_PREFIX_CMD git pull
+
+for PLUGIN in ${PLUGINS_DIR[*]};
+do
+    echo "$DEPLOY_PREFIX_CMD git -C $PLUGIN pull"
+    $DEPLOY_PREFIX_CMD git -C $PLUGIN pull
+done
+
+# 3. run composer update
+echo "$DEPLOY_PREFIX_CMD composer update --no-interaction"
```
fquffio: Should we run `composer install` instead?

bin/simple_git_deploy.sh
```diff
@@ -0,0 +1,113 @@
...
+        if [[ -d $plug_dir/.git ]]; then
+            echo "plugin found: $dir"
+            PLUGINS_DIR+=("$plug_dir")
+        else
+            echo "plugin NOT found: $plug_dir"
+        fi
+    done
+else
+    echo "Look for plugins in plugins/ folder"
+    for dir in ./plugins/*/ ; do
```
fquffio: I'm not quite sure that updating _all_ plugins is a desirable behaviour when nothing is specified… 🤔 In development this is probably true, but would it cause problems with current working copy? What happens if a user has changed files in one or more plugin, and they launch this script?

fquffio (approved):

bin/simple_git_deploy.sh
stefanorosanelli: That's right, we could change strategy:
* if no argument is passed no plugin is updated
* if a plugin list is specified those plugins are updated
* using a special argument like `--all-plugins` all plugins are updated

bin/simple_git_deploy.sh
stefanorosanelli: However if you're not on the root folder other commands will fail....

fquffio: Uhm… Right. `git` and `composer` should include the `-C $BE4_DIR` option, plugins auto-list should be aware of the working dir, and `bin/cake` should become `$BE4_DIR/bin/cake`.

I haven't tried, but I think it should work… However this is not essential — 99% of the times the working directory will be the correct one.

stefanorosanelli: Done, now the script is callable from anywhere

Issue #1366, opened 2017-10-30
Object types operations rules


stefanorosanelli:
Some rules are needed on `/model/object_types` write actions:

1. ~~**delete** an abstract object_type with sub-types should be **forbidden**~~ - done in #1374  
1. ~~**delete** an object type with at least 1 object should be **forbidden**~~ - done in #1374
1. setting `is_abstract` _true_ on types with an existing object should be inhibited, and also setting _false_ if a subtype exists
1. `parent_name` should be a valid abstract object type, not changeable if objects of this type exist
1. introduction of `core_type` flag, marking core and custom types: a `core` type cannot be deleted
1. introduction of `enabled` flag: if false no new object of this type can be created, setting `enabled` false is forbidden if objects of this type exist
1. `plugin` and `model` attributes are hidden, only `ŧable` property will be exposed in the current form `{plugin}.{model}` - corresponding `{plugin}` and `{model}` must be valid and available.
1. there should be some consistency between `parent_name` and `table`: using `BEdita/Core.Objects` as model and `media` as parent type name should be **forbidden**
1. `associations` and `hidden` should be validated as arrays with meaningful items
1. `name` and `singular` should not be numbers to avoid conflicts with ids

@bedita/core any missing or bad rules?

As a side-note also `created` and `modified` fields should be added to `object_types`.

PR #1370, opened 2017-10-31
Parent name in object types

### LINKED ISSUE #1369: Object type parent on creation
After #1356 on object type creation a parent type should be set via `parent_name` attribute.

Default parent must be set to `objects` if not specified.

stefanorosanelli:
This PR fixes #1369 

- a virtual property `parent_name` was introduced as parent object type name
- on object type creation a default `parent_id = 1` is set
- nested set model tree updated calling `Tree::recover()` - there's probably a better way to do that

There are no checks on `parent_name` value at this moment, this will be addressed in #1366 

Bonus tracks: 
 - a new integration test with object type and object creation, view and removal
 - `users` type now extends `objects` and not `profiles`


plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -186,13 +195,29 @@ public function get($primaryKey, $options = [])
...
+    /**
      * Invalidate cache after saving an object type.
+     * Recover Nested Set Model tree structure (tree_left, tree_right)
      *
      * @return void
      */
     public function afterSave()
     {
         Cache::clear(false, self::CACHE_CONFIG);
+        $this->behaviors()->get('Tree')->recover();
```
fquffio: Is this required? Since Tree behavior is attached, any operation that invokes callbacks (i.e.: `save()`, `saveMany()`, `delete()`, …) should already result in a valid tree. However, `updateAll()` and `deleteAll()` still require this… but those do not fire the `afterSave` event either. 🤔

plugins/BEdita/API/tests/IntegrationTest/NewObjectTypesTest.php
```diff
@@ -0,0 +1,142 @@
...
+     * Test new object type creation with object implementation
+     *
+     * @param $typeData array New object type intpu data
+     * @param $attributes array New object attributes
+     * @dataProvider newTypesProvider
+     * @coversNothing
+     */
+    public function testNewTypes($typeData, $attributes)
+    {
+        $lastId = $this->lastObjectId();
```
fquffio, line 73: This test case covers many operations. If one fails the whole test fails, making it difficult to spot the real cause of the error at first sight, and possibly hiding other failures. I'd prefer to split this test case in multiple test cases.

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
```diff
@@ -196,4 +201,36 @@ protected static function listAssociations(Table $Table, array $hidden = [])
...
+     * Setter for virtual property `parent_name`.
+     *
+     * @param string $parentName Parent object type name.
+     * @return string
+     */
+    protected function _setParentName($parentName)
+    {
+        try {
+            $objectType = TableRegistry::get('ObjectTypes')->get($parentName);
+            $this->parent_id = $objectType->id;
```
fquffio: Maybe we can also set `$this->parent = $objectType`.

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
```diff
@@ -492,4 +499,36 @@ public function testFindAll()
...
+        static::assertTrue((bool)$success);
+
+        $this->ObjectTypes->delete($objectType);
+        $data['parent_name'] = 'objects';
+        $this->ObjectTypes->patchEntity($objectType, $data);
+
+        $success = $this->ObjectTypes->save($objectType);
+        static::assertTrue((bool)$success);
+
+        $this->ObjectTypes->delete($objectType);
```
fquffio: This instruction is not needed: fixtures already take care of cleaning up things.

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
```diff
@@ -492,4 +499,36 @@ public function testFindAll()
...
+            'name' => 'foos',
+            'plugin' => 'BEdita/Core',
+            'model' => 'Objects',
+        ];
+
+        $objectType = $this->ObjectTypes->newEntity();
+        $this->ObjectTypes->patchEntity($objectType, $data);
+
+        $success = $this->ObjectTypes->save($objectType);
+        static::assertTrue((bool)$success);
```
fquffio: Should we `static::assertSame(ObjectTypesTable::DEFAULT_PARENT_ID, $success->parent_id)` too?

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
```diff
@@ -492,4 +499,36 @@ public function testFindAll()
...
+
+        $this->ObjectTypes->delete($objectType);
```
fquffio: Again, I'd split this "second part" into its own test case… Or maybe we can use a data provider instead, that attempts a save with multiple `$data` sets.

fquffio (changes requested):

plugins/BEdita/API/tests/IntegrationTest/NewObjectTypesTest.php
stefanorosanelli: That's true, but since it's an integration test (with zero coverage) not a unit test I don't see particular problems here. 
Testing this sequence of API calls is the actual goal.

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
stefanorosanelli: Honestly I don't know, but if I comment this line `tree_left` and `tree_right` are not correctly updated.
I'm sure there's a better way to do it.... but at least it works

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
stefanorosanelli: I'll add a provider.

fquffio (approved):

PR #1371, opened 2017-10-31
Refactor BEdita shell + solve problematic behaviour with SQLite

### LINKED ISSUE #1178: Shell test BeditaShellTest: nasty behaviour with sqlite, in following shell test
Using branch `issue/v4/1096-manage-applications-endpoints-and-endpoints-permissions-via-shell`, there's an issue with class `BeditaShellTest`, using sqlite connection (i.e. in `config/app.php`: `['Datasources' => ['test' => ['url' => 'sqlite:///tmp/bedita4.sqlt']]]`).

- Launching phpunit separately on `plugins/BEdita/Core/tests/TestCase/Shell/BeditaShellTest.php` and `plugins/BEdita/Core/tests/TestCase/Shell/EndpointPermissionsShellTest.php` works fine.
- Launching phpunit on `plugins/BEdita/Core/tests/TestCase/Shell/` folder returns error on `EndpointPermissionsShellTest` (first test using fixtures `roles`, `object_types`, `endpoints`, `applications`, `endpoint_permissions`). Failing nature: `SQLSTATE[23000]: Integrity constraint violation: 19 FOREIGN KEY constraint` / during `DELETE FROM roles`.

Something goes wrong in `BeditaShellTest`, when db connection alias is set (`ConnectionManager::alias('test', 'default')`).

Temporarily file `BeditaShellTest` has been renamed to `ZZZBeditaShellTest`, to skip travis errors.

fquffio:
This PR fixes #1178.

The fix includes a major refactor of BEdita shell: each step in the `bin/cake bedita setup` command has been split into its own Task, that is unit tested separately. This helps with the problem that we used to experience, since we have more control on each step tests, and we're able to better handle the mess those tasks do with connections and database tables. Also, tasks are now more thoroughly tested.


fquffio:
There's something weird with code coverage… For instance, [SetupConnectionTask](https://codecov.io/gh/bedita/bedita/pull/1371/tree/plugins/BEdita/Core/src/Shell/Task) seem to be not covered at all, while tests are present and `@covers` annotations have been set up properly, as far as I can tell… 🤔

stefanorosanelli:
Strange behavior indeed.

With `vendor/bin/phpunit --coverage-html ...` we get 0% coverage on [SetupConnectionTask](https://codecov.io/gh/bedita/bedita/pull/1371/tree/plugins/BEdita/Core/src/Shell/Task) instead with `vendor/bin/phpunit --coverage-html ... plugins/BEdita/Core/tests/TestCase/Shell/Task/` we get:

- 88.78% coverage on MySQL 
- 74.49% coverage on SQlite (less lines are covered in this case) but 2 failures...

plugins/BEdita/Core/src/Shell/Task/CheckApiKeyTask.php
```diff
@@ -0,0 +1,75 @@
...
+
+        try {
+            $this->verbose('=====> Loading default application... ', 0);
+            $application = $this->Applications->get(ApplicationsTable::DEFAULT_APPLICATION);
+            $this->verbose('<info>DONE</info>');
+        } catch (RecordNotFoundException $e) {
+            $this->verbose('<error>FAIL</error>');
+            $this->abort('Default application is missing, please check your installation');
+
+            return false;
```
stefanorosanelli: unreachable statement after `$this->abort()`

plugins/BEdita/Core/src/Shell/Task/CheckFilesystemTask.php
```diff
@@ -0,0 +1,160 @@
...
+     */
+    public function main(...$paths)
+    {
+        // Load paths to be checked.
+        $paths = array_unique(array_filter($paths) ?: [TMP, LOGS]);
+
+        // Detect HTTP daemon user.
+        $httpdUser = $this->getHttpdUser();
+        if (empty($httpdUser)) {
+            $this->out('=====> <warning>Unable to detect webserver user</warning>');
```
stefanorosanelli, line 64: missing coverage of empty `$httpdUser`, but we can live without :wink:

fquffio: Actually, this part is covered by `CheckFilesystemTaskTest::testExecuteAutodetectOk()` and `CheckFilesystemTaskTest::testExecuteAutodetectFail()`. However, one of this test is always skipped (if a webserver is running, fail is skipped, if a webserver is _not_ running, ok is skipped). In local environment it is easy to test both cases by running tests then turning off Apache or whatever, and re-running them again.

stefanorosanelli: Ok, I only looked at the coverage results...

stefanorosanelli:
Code coverage problem is somehow still related to test execution order: renaming `BEditaShellTest` to `ZBEditaShellTest` increases coverage of `SetupConnectionTask ` from 0% to 88.78%

fquffio:
@stefanorosanelli I'm trying to user more specific `@covers` annotations, to see if it solves the problem. As far as I can tell there's nothing wrong in our configuration, however… 🤔

fquffio:
Ok, we ran into what looks to me like a PHPUnit bug, but might also be a kind of bad practice I'm not aware of. `BeditaShellTest` used to reference two constants defined in `SetupConnectionTaskTest`, thus that test case was autoloaded before being executed, and PHPUnit silently skipped it. c28153c should have worked arount this, and hopefully PR coverage will become a bit higher.

plugins/BEdita/Core/src/Shell/Task/SetupConnectionTask.php
```diff
@@ -0,0 +1,315 @@
...
+     * Configure connection.
+     *
+     * @return void
+     */
+    public function main()
+    {
+        $connectionName = $this->param('connection');
+        $connection = ConnectionManager::get($connectionName);
+        if (!($connection instanceof Connection)) {
+            $this->abort('Invalid connection object');
```
stefanorosanelli, line 99: [minor] is this check really necessary? 
we have it in many places, and is also a never tested case...

stefanorosanelli (approved):

plugins/BEdita/Core/src/Shell/Task/SetupConnectionTask.php
fquffio: Mmh… Despite not strictly necessary right now, it might become useful in the future. `ConnectionManager::get()` is guaranteed to return a `ConnectionInterface` instance. Database connections are just one possible implementation of `ConnectionInterface`. For example, ElasticSearch connections will be different implementations of the same interface.

This is why I would leave it in a place where we are explicitly managing a database connection: it might cause issues in the future if a user runs `bin/cake bedita setup --connection elasticsearch`. I have added a test case to test this condition.

PR #1374, opened 2017-11-06
Object types removal controls

stefanorosanelli:
This PR introduces some new validation rules on object type removal:

- Prevent delete if type is abstract and at least a subtype exists
- Don't allow delete if at least an object of this type exists
- a **RESTRICT** constraint is introduced in `objects.object_type_id`: before that a delete on `object_types` table caused removal of all objects with this type

Some changes has also been introduced in tests.
There are stille some problems in our tests sequence as we've seen in #1178 #1371. 
The `fixtureManager->shutDown()` trick introduced in `ObjectsHandlerTest` is necessary to avoid failure on sqlite on the global test sequence - don't ask me why :smiley_cat:


plugins/BEdita/API/tests/TestCase/Controller/ObjectTypesControllerTest.php
```diff
@@ -513,6 +513,7 @@ public function testEmpty()
         ];
 
         TableRegistry::get('Properties')->deleteAll([]);
+        TableRegistry::get('Objects')->deleteAll([]);
```
batopa, line 565: Why is it needed?

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -196,16 +197,25 @@ public function get($primaryKey, $options = [])
...
-     * @return void
+     * @throws \Cake\Network\Exception\ForbiddenException if operation on entity is not allowed
      */
     public function beforeRules(Event $event, EntityInterface $entity)
     {
         if ($entity->isNew() && empty($entity->parent_id)) {
             $entity->parent_id = self::DEFAULT_PARENT_ID;
         }
+        $eventData = $event->getData();
+        if (!empty($eventData['operation']) && $eventData['operation'] === 'delete'
```
batopa: I missed the point. Why isn't this logic in `beforeDelete()`?

plugins/BEdita/API/tests/TestCase/Controller/ObjectTypesControllerTest.php
batopa: Ok, I understood

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
batopa: Ok, I didn't read the comment... sorry :smile:

plugins/BEdita/Core/config/Migrations/20171104185428_ObjectTypesKey.php
```diff
@@ -0,0 +1,42 @@
...
+                ]
+            )
+            ->update();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function down()
+    {
```
fquffio: This method shouldn't be empty… it should revert changes made in `up()`. In this case it is not _strictly_ necessary since the migration is idempotent, but I'd prefer to have it anyways — and in any case it would probably be better to at least leave a comment to explain why it is being left empty.

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -220,6 +230,21 @@ public function afterSave()
...
+     * Don't allow delete actions if at least an object of this type exists.
+     *
+     * @param \Cake\Event\Event $event The beforeSave event that was fired
+     * @param \Cake\Datasource\EntityInterface $entity the entity that is going to be saved
+     * @return void
+     * @throws \Cake\Network\Exception\ForbiddenException if entity is not deletable
+     */
+    public function beforeDelete(Event $event, EntityInterface $entity)
+    {
+        if (TableRegistry::get('Objects')->find()->where(['object_type_id' => $entity->id])->count() > 0) {
```
fquffio: `count()` triggers a count of how many rows of that kind exist. It should be fast, but we're not interested in the actual amount of rows exist — it's enough to check if one exist.

We might use `TableRegistry::get('Objects')->exists([/* ... */])` instead. It should be faster since it leverages the load on database, and possibly be even more self-explanatory.

tests/bootstrap.php
```diff
@@ -47,3 +47,5 @@
     ],
 ]);
 Configure::write('debug', true);
+
+Cache::clearAll();
```
fquffio: Is this necessary? Cache should be already disabled for all tests, and explicitly enabled for test cases that perform assertions on cache. Assertions on cache, however, should use an in-memory cache engine. If it's not the case, I'd solve this issue instead of clearing the whole cache.

plugins/BEdita/Core/tests/TestCase/Utility/ObjectsHandlerTest.php
```diff
@@ -56,6 +57,13 @@ public function setUp()
      */
     public function tearDown()
     {
+        // seems not necessary, but without `fixtureManager->shutDown`
+        // an integrity constraint error on foreign key is raised on SQLitr
+        $info = Database::basicInfo();
+        if ($info['vendor'] === 'sqlite') {
+            $this->fixtureManager->shutDown();
```
fquffio: If it's necessary, then I'd simply trigger the `shutDown()` for all database vendors, with an explanatory comment — it would be invoked for all vendors, but it's a test, so we don't care.

I'd like to investigate it a bit, though…

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
```diff
@@ -549,4 +551,47 @@ public function testParent($data)
...
+     * @return void
+     * @dataProvider beforeDeleteProvider
+     * @covers ::beforeDelete()
+     * @covers ::beforeRules()
+     */
+    public function testBeforeDelete($typeName, $expected)
+    {
+        if ($expected instanceof \Exception) {
+            $this->expectException(get_class($expected));
+            static::expectExceptionMessage($expected->getMessage());
```
fquffio: This method _should_ be called dynamically (i.e.: `$this->expectExceptionMessage()`).

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
```diff
@@ -549,4 +551,47 @@ public function testParent($data)
...
+                'documents',
+                new ForbiddenException('Objects of this type exist'),
+            ],
+        ];
+    }
+
+    /**
+     * Test `beforeDelete`
+     *
+     * @return void
```
fquffio: Missing `@param` annotations.

fquffio (changes requested):

plugins/BEdita/Core/config/Migrations/20171104185428_ObjectTypesKey.php
stefanorosanelli: :+1:  for an explanatory comment

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
stefanorosanelli: Right...

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
stefanorosanelli: Good to know...

plugins/BEdita/Core/tests/TestCase/Utility/ObjectsHandlerTest.php
stefanorosanelli: Problem is: with MySQL and postgres invoking `shutDown()` here causes a failure....nice, isn't it :smiley: ?

I tried to find out why, without success: I think this has to do with #1178 #1371

Launching this unit test alone is fine, problem is the test sequence.

tests/bootstrap.php
stefanorosanelli: Cache seems to be disabled only for `_bedita_object_types_`.
`_cake_core_` and `_cake_model_` instead are written and used on testsuite launch
one solution is to clear only those two - but this could be done of course in a different PR/issue

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
fquffio: Ok… so the point is that performing a query `SELECT ... FROM object_types WHERE parent_id = 123` in `ObjectTypesTable::beforeDelete()` always returns an empty set because the relevant nodes have already been deleted by `TreeBehavior::beforeDelete()`, right?

I'd suggest to apply these changes instead:
```patch
--- plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php	(revision 5a0b7fab88bc12e610c0362eba2a8091da86fa06)
+++ plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php	(revision )
@@ -43,6 +43,8 @@
  * @method \BEdita\Core\Model\Entity\ObjectType[] patchEntities($entities, array $data, array $options = [])
  * @method \BEdita\Core\Model\Entity\ObjectType findOrCreate($search, callable $callback = null, $options = [])
  *
+ * @mixin \Cake\ORM\Behavior\TreeBehavior
+ *
  * @since 4.0.0
  */
 class ObjectTypesTable extends Table
@@ -209,12 +211,7 @@
     public function beforeRules(Event $event, EntityInterface $entity)
     {
         if ($entity->isNew() && empty($entity->parent_id)) {
+            $entity->set('parent_id', self::DEFAULT_PARENT_ID);
-            $entity->parent_id = self::DEFAULT_PARENT_ID;
-        }
-        $eventData = $event->getData();
-        if (!empty($eventData['operation']) && $eventData['operation'] === 'delete'
-            && $entity->get('is_abstract') && $this->find()->where(['parent_id' => $entity->id])->first()) {
-            throw new ForbiddenException(__d('bedita', 'Abstract type with existing subtypes'));
         }
     }
 
@@ -239,6 +236,9 @@
      */
     public function beforeDelete(Event $event, EntityInterface $entity)
     {
+        if ($entity->get('is_abstract') && $this->childCount($entity) > 0) {
+            throw new ForbiddenException(__d('bedita', 'Abstract type with existing subtypes'));
+        }
         if (TableRegistry::get('Objects')->exists(['object_type_id' => $entity->id])) {
             throw new ForbiddenException(__d('bedita', 'Objects of this type exist'));
         }
```

This way we rely on `TreeBehavior::childCount($entity)` to know how many descendants the current node has. This _is not_ achieved through a query unless explicitly asked passing a second argument as `true`, but rather uses NSM left and right indexes to know how many nodes are between those values — basically `(right - left - 1) / 2`.

Raising an exception should hopefully roll back the transaction, so data integrity should be safe. 😱

plugins/BEdita/Core/tests/TestCase/Utility/ObjectsHandlerTest.php
fquffio: I tried merging the source branch of PR #1371 into this source branch, and the problem seems to have vanished… 🤔 

We might leave a `// TODO` comment to remind us that this part could be removed if/when both PRs are merged.

fquffio (approved):

plugins/BEdita/Core/tests/TestCase/Utility/ObjectsHandlerTest.php
stefanorosanelli: Better: we can merge #1371 and then remove `shutDown()`

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
stefanorosanelli: This will work, but honestly I'd prefer not to rely on a rollback for these kind of checks.
But using `childCount()` instead of `find()->first()` is better...

fquffio: Yes, you're right.

Also, in that case a lot of rows would be deleted by tables like `objects`, `object_relations`, `object_properties`, … due to foreign key constraints, so we would drop tons of rows and _then_ trigger a rollback. We better avoid.

PR #1382, opened 2017-11-13
Accept `false` and `true` as boolen values in filters

stefanorosanelli:
This PR enables query field filters like:
-  `/model/object_types?filter[is_abstract]=true`
- `/objects?filter[locked]=false`

that currently return a `500 - Internal Error` (!!!), also with wrong values like  `/objects?filter[locked]=value`

A custom database `BoolType` has been added. 
There are other possible (and better) solutions of course...

### DIFF
--- a/plugins/BEdita/Core/src/Database/Type/BoolType.php
+++ b/plugins/BEdita/Core/src/Database/Type/BoolType.php
@@ -22,7 +22,6 @@ use InvalidArgumentException;
  */
 class BoolType extends CakeBoolType
 {
-
     /**
      * Convert bool data into the database format.
      * `true` and `false` as strings are accepted, other strings are discarded
@@ -33,22 +32,13 @@ class BoolType extends CakeBoolType
      */
     public function toDatabase($value, Driver $driver)
     {
-        if ($value === true || $value === false || $value === null) {
-            return $value;
-        }
-
-        if (in_array($value, [1, 0, '1', '0'], true)) {
-            return (bool)$value;
-        }
-        if (is_string($value)) {
-            if (strtolower($value) === 'true') {
-                return true;
-            } elseif (strtolower($value) === 'false') {
-                return false;
-            } else {
-                return null;
+        try {
+            return parent::toDatabase($value, $driver);
+        } catch (InvalidArgumentException $e) {
+            if (is_string($value)) {
+                return filter_var($value, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE);
             }
+            throw $e;
         }
-        throw new InvalidArgumentException('Cannot convert value to bool');
     }
 }
--- a/plugins/BEdita/Core/src/Database/Type/BoolType.php
+++ b/plugins/BEdita/Core/src/Database/Type/BoolType.php
@@ -24,7 +24,7 @@ class BoolType extends CakeBoolType
 {
     /**
      * Convert bool data into the database format.
-     * `true` and `false` as strings are accepted, other strings are discarded
+     * `true` and `false` as strings are accepted, other strings will cause an `InvalidArgumentException`
      *
      * @param mixed $value The value to convert.
      * @param \Cake\Database\Driver $driver The driver instance to convert with.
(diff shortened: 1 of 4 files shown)


plugins/BEdita/Core/tests/TestCase/Database/Type/BoolTypeTest.php
```diff
@@ -0,0 +1,103 @@
...
+                new InvalidArgumentException('Cannot convert value to bool'),
+            ]
+        ];
+    }
+
+    /**
+     * Test `toDatabase` method
+     *
+     * @param mixed $expected Expected result
+     * @param mixed $input Input data to be marshaled.
```
didoda: are parameters in wrong order? first `$input` then `$expected`

fquffio: Usually we put expected result first, but it's not a rule nor a "common practice", as far as I can tell: it's just how data provider documentation is written in PHPUnit official docs.

stefanorosanelli: `@param` order fixed

didoda (approved):

plugins/BEdita/Core/src/Database/Type/BoolType.php
```diff
@@ -0,0 +1,54 @@
...
+    public function toDatabase($value, Driver $driver)
+    {
+        if ($value === true || $value === false || $value === null) {
+            return $value;
+        }
+
+        if (in_array($value, [1, 0, '1', '0'], true)) {
+            return (bool)$value;
+        }
+        if (is_string($value)) {
```
fquffio: `return filter_var($value, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE);` yields quite much the same result with no additional `if / elseif / else` structures. 🙃

plugins/BEdita/Core/src/Database/Type/BoolType.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Convert bool data into the database format.
+     * `true` and `false` as strings are accepted, other strings are discarded
+     *
+     * @param mixed $value The value to convert.
+     * @param \Cake\Database\Driver $driver The driver instance to convert with.
+     * @return bool|null
+     */
+    public function toDatabase($value, Driver $driver)
+    {
```
fquffio, line 34: To avoid duplicating logic, I'd prefer something like this:

```php
public function toDatabase($value, Driver $driver)
{
    try {
        return parent::toDatabase($value, $driver);
    } catch (InvalidArgumentException $e) {
        if (is_string($value)) {
            // ...
        }
        throw $e;
    }
}

fquffio (commented):
I'm not super-excited about overwriting such a "core" database type as `BoolType`.

However, the only other (simple) approach I can think of is to add a check in `QueryFilterTrait` that properly converts the value if the field is detected to be a boolean. That alternative doesn't look very clean to me, so 👍  for this approach.

plugins/BEdita/Core/src/Database/Type/BoolType.php
stefanorosanelli: seems better indeed...

stefanorosanelli: done

plugins/BEdita/Core/src/Database/Type/BoolType.php
stefanorosanelli: done

fquffio (approved):

fquffio:
👍  Thanks!

I'm just not 100% sure that a random string (e.g.: `gustavo`) should be interpreted as `null`… 😕

stefanorosanelli:
I'm not sure either, but interpreting a random string as `true` or `false` seems even worse... :smiley:

fquffio:
Yeah, but I think that in that case the `InvalidArgumentException` raised by CakePHP's default `BoolType` is quite relevant, and it should be left.

Since if uncaught, that exception would result in a 500 error, it's responsibility of a higher layer to intercept (or prevent) that exception, and "convert" it to a 400 error.

stefanorosanelli:
With last commit only strings like `true` and `false` are accepted
We should solve input filter validation in general to avoid **500** errors

PR #1383, opened 2017-11-13
Object types rules

### LINKED ISSUE #1366: Object types operations rules

stefanorosanelli:
This PR fixes #1366 

- new fields added to `object_types`, namely:
  * `created`, `modified` like many other tables/resources
  * `core_type` flag indicating whether this a core type (i.e. types created via migration, always present)
  * `enabled` to deactivate/activate a type
- `core_type` rules: 
  * a core type cannot be deleted
  * its value cannot be changed - it will always be `false` for custom types
- `enabled` rules: 
  * if `false` no new object of this type can be created
  * setting `enabled` to `false` is forbidden if objects of this type exist and if subtypes exist
  * if `false` it cannot be used as parent type
- `is_abstract` rules:
  * setting `true` is forbidden if objects of this type exist
  * setting `false` forbidden if a subtype exists
  * must be `true` to be used as `parent` type
  * if `true` with existing subtypes delete is forbidden
- `parent_name` should be an `enabled` abstract object type, not changeable if objects of this type exist
- `plugin` and `model` attributes are hidden, only `table` property is displayed in the form `{plugin}.{model}` - corresponding table model must exist in order to be used
- `name` and `singular` should include at least a letter `a-zA-Z` - should we force lower snake case?

Tests: `news` type in `ObjectTypesFixture` is created with `enabled` and `core_type` false in order to perform meaningful tests.

---

Rules not implemented, addressed in future issues 
- consistency between `parent_name` and `table`: e.g. using `BEdita/Core.Objects` as model and `media` as parent type name should be forbidden
- `associations` and `hidden` should be validated as arrays with meaningful items


plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -64,6 +66,20 @@ class ObjectTypesTable extends Table
     const DEFAULT_PARENT_ID = 1;
 
     /**
+     * Default `plugin` if not specified.
+     *
+     * @var int
```
didoda: `@var string`

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -64,6 +66,20 @@ class ObjectTypesTable extends Table
...
+     */
+    const DEFAULT_PLUGIN = 'BEdita/Core';
+
+    /**
+     * Default `model` if not specified.
+     *
+     * @var int
```
didoda: as above, `@var string`

plugins/BEdita/Core/src/Model/Validation/ObjectTypesValidator.php
```diff
@@ -42,12 +42,14 @@ public function __construct()
         $this
             ->requirePresence('name', 'create')
             ->notEmpty('name')
+            ->regex('name', '/^([a-zA-Z]+)/')
```
didoda: it's a simple regex, btw an inline doc could be useful, i.e.: `// valid name starts with alpha (not numeric) char`

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
```diff
@@ -596,4 +626,135 @@ public function testBeforeDelete($typeName, $expected)
...
+                ],
+                new BadRequestException('"Missing/Plugin.NotFound" is not a valid model table name'),
+            ],
+        ];
+    }
+
+    /**
+     * Test `beforeSave`
+     *
+     * @param string $typeName Object type name to save
```
didoda: wrong param? `@param array $data ...`

didoda (changes requested):
minor changes in phpDoc

plugins/BEdita/Core/config/Migrations/20171110095938_AddCoreEnabledCreatedModified.php
```diff
@@ -0,0 +1,96 @@
...
+                    'enabled',
+                ],
+                [
+                    'name' => 'objecttypes_enabled_idx',
+                ]
+            )
+            ->update();
+
+        // UPDATE created, modified
+        $this->query(sprintf("UPDATE object_types SET created = '%s', modified = '%s'", date('Y-m-d H:i:s'), date('Y-m-d H:i:s')));
```
fquffio, line 60: Could this be achieved by setting `created` and `modified` column with `CURRENT_TIMESTAMP` as their default value?

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -211,11 +230,27 @@ public function get($primaryKey, $options = [])
      */
     public function beforeRules(Event $event, EntityInterface $entity)
     {
-        if ($entity->isNew() && empty($entity->parent_id)) {
-            $entity->set('parent_id', self::DEFAULT_PARENT_ID);
+        if ($entity->isNew()) {
+            if (empty($entity->get('parent_id'))) {
+                $entity->set('parent_id', self::DEFAULT_PARENT_ID);
+            }
+            if (empty($entity->get('plugin'))) {
```
fquffio: This might cause problems with models that are in application and _not_ in any plugin! `plugin` should be allowed to be `null`.

We can replace these checks with a single `if (empty($entity->get('table'))`, that sets table to `BEdita/Core.Objects` by default.

I know we're not planning to support models outside of plugin scope, but dealing with `table` property also provides a simpler approach, and I think it's safer than setting `plugin` and `model` to their default values in parallel.

fquffio (changes requested):

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectTypesTableTest.php
stefanorosanelli: yes, typo
fixed

plugins/BEdita/Core/src/Model/Validation/ObjectTypesValidator.php
stefanorosanelli: done

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
stefanorosanelli: done

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
stefanorosanelli: done

plugins/BEdita/Core/config/Migrations/20171110095938_AddCoreEnabledCreatedModified.php
stefanorosanelli: done

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
stefanorosanelli: ok, done

plugins/BEdita/Core/config/Migrations/20171110095938_AddCoreEnabledCreatedModified.php
stefanorosanelli: tried, but somehow unit tests are failing on sqlite on `BEditaShell` - mysteriously as usual :wink:

didoda (approved):

fquffio (approved):

PR #1386, opened 2017-11-15
Add garbage collection for refresh_token

### LINKED ISSUE #1385: [API] Configure refresh_token expired time and clean db from obsolete entries
Currently `refresh_token` is saved in `hash_jobs` table with expired time fixed to `2050-01-01`. To avoid the risk that `hash_jobs` table quickly grows we need to delete entries that aren't used for some time.

We need to configure an expired time then every time a user accesses to a refresh token we should update the `hash_job.expired` field and delete expired `refresh_jobs` with some probability implementing some forms of garbage collection.

batopa:
This PR fixes #1385 

After this PR the `refresh_token` expire time is configurable via `api.auth.refreshTokenExpiresIn`. Default is set to `1 month`.
Every time a renew is performed the `hash_jobs.expired` field related to `refresh_token` will be updated.

In order to clean `hash_jobs` table from expired/unused `refresh_token` a garbage collection method `ApiAuthComponent::gc()` has been introduced and called every time an `access_token` is generated i.e. when user tries to authenticate or to renew an expired `access_token`.

### DIFF
--- a/bedita-app/controllers/components/api_auth.php
+++ b/bedita-app/controllers/components/api_auth.php
@@ -61,20 +61,26 @@ class ApiAuthComponent extends Object implements ApiAuthInterface {
     protected $token = null;
 
     /**
-     * The payload used token generation
+     * The payload used for token generation
      *
      * @var array
      */
     protected $payload = array();
 
     /**
-     * Configuration used to customize token generation
+     * Configuration used to customize token generation.
+     * - `JWT` contains configurations to generate the `access_token`
+     * - `refreshTokenExpiresIn` is the `refresh_token` expires time from now.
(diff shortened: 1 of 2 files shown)


bedita-app/controllers/components/api_auth.php
```diff
@@ -210,6 +222,26 @@ public function renewToken($refreshToken) {
...
 
     /**
+     * Garbage collection for `refresh_token`.
+     * Delete from `hash_jobs` table the reference to expired `refresh_token`
+     * with some probability
+     *
+     * @return void
+     */
+    protected function gc() {
+        if (mt_rand(1, 100) > 5) {
```
fquffio: 95% ? 😳

didoda (approved):

PR #1390, opened 2017-11-22
Use `application/schema+json` content type

stefanorosanelli:
This PR introduces the expected content type `application/schema+json`  for `JSON Schema` on `/model/schema` endpoint

There is still no content type negotiation check on `Accept` headers (all types are accepted), but use of `Accept: application/schema+json` is recommended.

Tests on content type have been added.


plugins/BEdita/API/tests/TestCase/Controller/Model/SchemaControllerTest.php
```diff
@@ -34,28 +34,54 @@ class SchemaControllerTest extends IntegrationTestCase
         'plugin.BEdita/Core.properties',
     ];
 
+    /**
+     * Data provider for `testJsonSchema`
+     *
+     * @return void
```
fquffio: [minor] `@return array`

plugins/BEdita/API/src/Controller/Model/SchemaController.php
```diff
@@ -26,14 +26,26 @@
...
      */
     public function initialize()
     {
         parent::initialize();
-        $this->components()->unload('JsonApi');
-        $this->viewBuilder()->setClassName('Json');
+        if ($this->components()->get('JsonApi')) {
+            $this->components()->unload('JsonApi');
+        }
+        $this->viewBuilder()->setClassName('BEdita\API\View\JsonSchemaView');
```
fquffio: Frankly I don't like having a custom view class just for the sake of this scope… It looks quite much a bit overkill to me.

I'd use a different and simpler approach, even though it is less reusable:
```diff
--- plugins/BEdita/API/src/Controller/Model/SchemaController.php	(revision 128981534193332d8b07aec4cbec29da075e4483)
+++ plugins/BEdita/API/src/Controller/Model/SchemaController.php	(revision )
@@ -16,7 +16,6 @@
 use BEdita\API\Controller\AppController;
 use BEdita\Core\Model\Schema\JsonSchema;
 use Cake\Event\Event;
-use Cake\Routing\Router;
 
 /**
  * Controller for `/model/schema/{type}` endpoint.
@@ -39,13 +38,11 @@
     public function initialize()
     {
         parent::initialize();
+
         if ($this->components()->get('JsonApi')) {
             $this->components()->unload('JsonApi');
         }
-        $this->viewBuilder()->setClassName('BEdita\API\View\JsonSchemaView');
-        $this->response->type([
-            'jsonschema' => self::CONTENT_TYPE,
-        ]);
+        $this->viewBuilder()->setClassName('Json');
     }
 
     /**
@@ -64,7 +61,7 @@
      * Get JSON-SCHEMA of a type.
      *
      * @param string $typeName Name of an object type or of a resource type.
-     * @return void
+     * @return \Cake\Http\Response
      */
     public function jsonSchema($typeName)
     {
@@ -73,5 +70,7 @@
         $url = (string)$this->request->getUri();
         $this->set(JsonSchema::generate($typeName, $url));
         $this->set('_serialize', true);
+
+        return $this->render()->withType(static::CONTENT_TYPE);
     }
 }
```

fquffio (commented):
If we need to serve JSON Schema with its specific content type only here, I'd get rid of `JsonSchemaView`, if possible.

plugins/BEdita/API/src/Controller/Model/SchemaController.php
```diff
@@ -26,14 +26,26 @@
...
+
     /**
      * {@inheritDoc}
      */
     public function initialize()
     {
         parent::initialize();
-        $this->components()->unload('JsonApi');
-        $this->viewBuilder()->setClassName('Json');
+        if ($this->components()->get('JsonApi')) {
```
fquffio: `$this->components()->has('JsonApi')` would be fine in this case.

plugins/BEdita/API/src/Controller/Model/SchemaController.php
stefanorosanelli: That's right, I quickly tried similar solutions but didn't find this one.

fquffio (approved):

PR #1397, opened 2017-12-20
Schema for static properties

### LINKED ISSUE #1379: Default property types
- define core `property_types` to be inserted via migrations like: `string`, `int`, `float`, `bool`, `date`
- implement correct type association for static properties

fquffio:
This PR closes #1379.

Default property types have been introduced thanks to a schema migration. Furthermore, a refactor of the logic that automatically detects the most appropriate property type for a schema column has been made.


plugins/BEdita/Core/config/Migrations/20171109094654_AddDefaultPropertyTypes.php
```diff
@@ -0,0 +1,106 @@
...
+                [
+                    'name' => 'url',
+                    'params' => '{"type":"string","format":"uri"}',
+                ],
+                [
+                    'name' => 'date',
+                    'params' => '{"type":"string","format":"date-time"}',
+                ],
+                [
+                    'name' => 'number',
```
stefanorosanelli, line 58: what about adding also `integer` as core type `{"type":"number", "format":"integer"}`  ?

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
```diff
@@ -247,4 +248,40 @@ protected function _setParentName($parentName)
...
+        foreach ($allProperties as $property) {
+            $accessMode = null;
+            if (!$entity->isAccessible($property->name)) {
+                $accessMode = 'readOnly';
+            } elseif (in_array($property->name, $hiddenProperties)) {
+                $accessMode = 'writeOnly';
+            }
+            $properties[$property->name] = $property->getSchema($accessMode);
+
+            if ($property->required && $accessMode !== 'readOnly') {
```
stefanorosanelli: does this mean that (not nullable) hidden properties could be `required` ? :thinking:

plugins/BEdita/Core/src/Model/Table/PropertyTypesTable.php
```diff
@@ -84,4 +102,102 @@ public function buildRules(RulesChecker $rules)
...
+
+        // Try to infer a generic type from column definition.
+        $type = $table->getSchema()->getColumnType($name);
+        if (isset($propertyTypes[$type])) {
+            return $propertyTypes[$type];
+        }
+        switch ($type) {
+            // Try to convert specific types to more generic ones.
+            case 'integer':
+                $type = 'number';
```
stefanorosanelli: as said before: what about an `integer` type?

plugins/BEdita/Core/src/Utility/JsonSchema.php
```diff
@@ -0,0 +1,135 @@
...
+
+            $accessMode = null;
+            if (!$entity->isAccessible($name)) {
+                $accessMode = 'readOnly';
+            } elseif (in_array($name, $hiddenProperties)) {
+                $accessMode = 'writeOnly';
+            }
+            $property = new StaticProperty(compact('name', 'table'));
+            $properties[$name] = $property->getSchema($accessMode);
+            if ($property->required && $accessMode !== 'readOnly') {
```
stefanorosanelli: `writeOnly` hidden properties could be required ? (same note in `Model/Entity/ObjectType.php`)

stefanorosanelli (approved):

plugins/BEdita/Core/config/Migrations/20171109094654_AddDefaultPropertyTypes.php
batopa: It seems a good idea. What about specs? Anyway I will eventually open another issue for that-

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
batopa: If I understood `writeOnly` properties can be traced as `required` but `writeOnly` props never appears in API responses.
Maybe should be 
```
if ($property->required && $accessMode === null)
```
or maybe I missed something :smile_cat:

plugins/BEdita/Core/src/Utility/JsonSchema.php
batopa: Same question as above also for me :recycle:

plugins/BEdita/Core/config/Migrations/20171109094654_AddDefaultPropertyTypes.php
stefanorosanelli: in draft6 and 7 there's no `integer` primitive type but only `number`
http://json-schema.org/latest/json-schema-core.html#rfc.section.4.2.1
but `integer` is one of the most obvious custom type

fquffio: `integer` is an allowed `type`. However, it might be difficult to validate since some programming languages just don't distinguish between numbers with and without a fractional part.

We might add `{"type": "integer", "multipleOf": 1}` to try to achieve the maximum portability… 🤔

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
fquffio: In theory (...) I think that a `writeOnly` property could be required. For instance, even if it is not our case, one shall not create a User without a password, or an Application without an API key, …. Those would be examples of required fields that are writeOnly.

For readOnly fields the concept of required is IMHO less meaningful. However I agree with @batopa that requiring `$accessMode === null` might avoid this confusion.

plugins/BEdita/Core/src/Model/Table/PropertyTypesTable.php
fquffio: In that case we might just drop this `case` branch.

plugins/BEdita/Core/src/Utility/JsonSchema.php
fquffio: Right now, yes. But I see that it might be confusing at times… 🤔

plugins/BEdita/Core/config/Migrations/20171109094654_AddDefaultPropertyTypes.php
fquffio: @stefanorosanelli According to [Draft-7 JSON Schema Validation](http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.1.1) "integer" is an allowed keyword for `type`…

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
stefanorosanelli: ok @batopa proposal seems fine

plugins/BEdita/Core/config/Migrations/20171109094654_AddDefaultPropertyTypes.php
stefanorosanelli: let's use simply `{"type": "integer"}` for now then

batopa (approved):

Issue #1402, opened 2018-01-31
Folders model + API


stefanorosanelli:
* introduce `folders` model, each folder must have a unique reference in `trees` table (nested set model)
* `folders` with empty `parent_id` should have its id as `root_id` => root folders
* each object contained in a folder must contain paths in meta response - `GET /documents/{{id}}` should return in `"meta"` 
```
"paths" : [
  "/3/4",
  "/35",
]
```
* `folders` types MUST have only one item in `"paths"` or no items if root folders
* each folder or leaf object must have `parent` tree relationship
* `folders` must have also `children` tree relationship
* descendents will be availabe via filters
  - `GET /objects?filter[ancestor]={{id}}` objects having `{{id}}` as ancestor
  - `GET /documents?filter[ancestor]={{id}}` same for documents

* setting objects inside `folders` will work as usual `JSON API` relationships using `children`/`parent` as names

PR #1408, opened 2018-02-01
Folders model

stefanorosanelli:
This PR is a first step of  #1402 

A migration to add `folders` core type has been added
Unit tests have been updated

@batopa feel free to merge or work in the same branch

---

**UPDATE**

The Folders model part should be done.
Creating a `folder` without `parent` save it as root folder.

`parent_id` and `parent` are virtual properties for `Folder` entity and setting it will be set `parents` property needed for save.

```php
// save root folder
$folders = TableRegistry::get('folders');
$f = $folders->newEntity(['title' => 'root']);
$f->type = 'folders';
$folders->save($f);
```

```php
// save folder as child
$folders = TableRegistry::get('folders');
$f = $folders->newEntity(['title' => 'folder']);
$f->type = 'folders';
$f->parent_id = 15;
// or equivalent
// $f->parent = $folders->get(15);
$folders->save($f);
```

```php
// folder becomes root
$folders = TableRegistry::get('folders');
$f = $folders->get(16);
$f->parent_id = null;
// or equivalent
// $f->parent = null;
$folders->save($f);
```

The API parts is partially implemented and will be completed in another PR.
Moreover some issues about missing `relationhips` links for inherited associations are solved too.

Now for every object should be present all links to associations inherited (for example `parents` that is defined in `ObjectsTable` is visible in every object) and **BEdita relations** defined for inherited abstract object types.


plugins/BEdita/API/tests/TestCase/Controller/RolesControllerTest.php
```diff
@@ -506,6 +506,12 @@ public function testRelated()
                                 'self' => 'http://api.example.com/users/1/relationships/roles',
                             ],
                         ],
+                        'parents' => [
```
stefanorosanelli, line 509: can `roles` have folder parents? they shouldn't since they are not objects I guess....

plugins/BEdita/Core/config/Migrations/20180208104214_AddTreesParentNodeId.php
```diff
@@ -0,0 +1,123 @@
...
+    public function up()
+    {
+        $this->table('trees')
+            ->removeIndexByName('trees_rootleft_idx')
+            ->removeIndexByName('trees_rootright_idx')
+            ->update();
+
+        $this->table('trees')
+            ->addColumn('parent_node_id', 'integer', [
+                'comment' => 'parent node id',
```
stefanorosanelli, line 28: we could say: `parent node, points to trees.id` - this is a reference to trees.id not to objects.id right?

plugins/BEdita/Core/src/Model/Entity/Folder.php
```diff
@@ -0,0 +1,126 @@
...
+use Cake\ORM\Entity;
+use Cake\ORM\Table;
+use Cake\ORM\TableRegistry;
+use Cake\Utility\Hash;
+
+/**
+ * Folder Entity
+ *
+ * @property int $parent_id
+ *
```
stefanorosanelli: `@property \BEdita\Core\Model\Entity\Folder|null $parent`
missing?

plugins/BEdita/Core/src/Model/Entity/Tree.php
```diff
@@ -1,38 +1,116 @@
...
+     * Setter for `parent_id`.
+     * It also set `root_id` and `parent_node_id`.
+     *
+     * When `$parent_id` is null it means that the node is a root (first level on tree) so
+     * - the `root_id` is set equal to `object_id`
+     * - the parent_node_id is set to `null`
+     *
+     * When `$parent_id` is not null the parent node is retrieved and
+     * - the `root_id` is set equal to parent node `root_id`
+     * - the `parent_node_id` is set to pthe parent node `id`
```
stefanorosanelli: typo `pthe`

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Custom rule for checking that entity has at most one parent.
+     * The check is done on `parents` property
+     *
+     * @param EntityInterface $entity The entity to check
+     * @return bool
+     */
+    public function hasAtMostOneParent(EntityInterface $entity)
+    {
+        if (empty($entity->parents)) {
```
stefanorosanelli: Typical `Scrutinizer` complaint:
should we use `$entity->get('parents')` here instead?


```
Accessing parents on the interface Cake\Datasource\EntityInterface suggest that you code against a concrete implementation. How about adding an instanceof check?```

File: plugins/BEdita/Core/src/Model/Table/FoldersTable.php [hunk 6]
```diff
@@ -0,0 +1,153 @@
...
+            return true;
+        }
+
+        $rule = new ValidCount('parents');
+
+        return $rule($entity, ['operator' => '==', 'count' => 1]) && !empty($entity->parent->id);
```
stefanorosanelli, line 99: same as above

plugins/BEdita/API/tests/TestCase/Controller/RolesControllerTest.php
batopa: It refers to `users`

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
batopa: done

stefanorosanelli:
Ok for me.
[can't **approve** since I'm the original author...]

plugins/BEdita/Core/config/Migrations/20180131171902_FoldersType.php
```diff
@@ -0,0 +1,78 @@
...
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
+use Cake\ORM\Table;
+use Migrations\AbstractMigration;
+
+/**
+ * Add `fodlers` core media type.
```
fquffio: [minor] Typo: `folders`.

plugins/BEdita/Core/config/Migrations/20180208104214_AddTreesParentNodeId.php
fquffio: 👍 Only, we should be careful using commas and other characters such as parentheses in column comments, as they make SQLite columns unchangeable… 😞 

(This was in response to @stefanorosanelli comment.)

plugins/BEdita/Core/src/Model/Entity/Folder.php
```diff
@@ -0,0 +1,127 @@
...
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
+namespace BEdita\Core\Model\Entity;
+
+use Cake\ORM\Entity;
```
fquffio: Unused import.

plugins/BEdita/Core/src/Model/Entity/Tree.php
```diff
@@ -1,38 +1,116 @@
...
+     * ensuring the recover of the tree nested set model.
+     *
+     * @param \BEdita\Core\Model\Entity\Folder|null $folder The folder entity to set as parent
+     * @return \BEdita\Core\Model\Entity\Folder|null
+     */
+    protected function _setParentObject(Folder $folder = null)
+    {
+        $parentId = ($folder === null) ? null : $folder->id;
+
+        if ($this->parent_id !== $parentId) {
```
fquffio: This check is already performed internally by `parent_id` setter. I'd avoid this duplication…

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
```diff
@@ -0,0 +1,153 @@
...
+        $node = $trees->find()
+            ->where(['object_id' => $entity->id])
+            ->firstOrFail();
+
+        // parent unchanged
+        if ($entity->parent_id === $node->parent_id) {
+            return;
+        }
+
+        $node->parent_id = $entity->parent_id;
```
fquffio, line 150: Maybe (AKA: maybe not), we could lower number of conditions by doing something nasty like this:

```php
public function afterSave($event, $entity)
{
    $trees = TableRegistry::get('Trees');
    if ($entity->isNew()) {
        // (as above)
    }

    $node = $trees->find()->where([/*...*/])->firstOrFail();
    $node->parent_id = $entity->parent_id; // Same parent_id checks already happens internally.
    if (!$entity->isDirty('parent_node_id')) {
        return; // Tree unchanged.
    }

    $trees->saveOrFail($node);
}
```

This might not be as straightforward as your approach is, but it might save one condition, and perhaps `Folder::isParentSet()` could be removed too.

plugins/BEdita/Core/src/Model/Table/TreesTable.php
```diff
@@ -99,9 +104,76 @@ public function validationDefault(Validator $validator)
...
+
+        $table = TableRegistry::get('Objects');
+        $isParentFolder = $table->find()
+            ->where([
+                $table->aliasField('id') => $entity->parent_id,
+            ])
+            ->innerJoinWith('ObjectTypes', function (Query $query) {
+                return $query->where(['ObjectTypes.name' => 'folders']);
+            })
+            ->count();
```
fquffio: Since we're using a strict condition on the primary key, it is fine to `COUNT(*)`. However, in these cases it is generally more efficient to have a query built this way:

```sql
SELECT 1 FROM mytable WHERE myfield = 'something' LIMIT 1;
```

This prevents SQL from loading more rows than needed, as we actually only want to know if a row that respects such conditions exists — whether it is one or a thousand is not relevant.

CakePHP's ORM has a `Table::exist($conditions)` method that could be used. In this case:

```php
return TableRegistry::get('Objects')->exists([
    'id' => $entity->parent_id,
    'object_type_id' => TableRegistry::get('ObjectTypes')->get('folders')->id,
]);
```

plugins/BEdita/Core/src/Model/Table/TreesTable.php
```diff
@@ -99,9 +104,76 @@ public function validationDefault(Validator $validator)
...
+    /**
+     * Check that `parent_id` property of the entity corresponds to a folder
+     *
+     * @param EntityInterface $entity The entity to validate
+     * @return bool
+     */
+    public function isParentValid(EntityInterface $entity)
+    {
+        if ($entity->parent_id === null) {
+            return true;
```
fquffio: `parent_id` can be `null` if and only if `$entity->object_id` is an ID of a folder. Is it worth checking?

plugins/BEdita/Core/src/ORM/Inheritance/AssociationCollection.php
```diff
@@ -56,11 +57,34 @@ public function __construct(Table $table)
...
+     * Get the inherited associations collection.
+     *
+     * The clone of `self::innerCollection` is cleaned by `RelatedTo` associations
+     * that involve concrete objects.
+     *
+     * @return \Cake\ORM\AssociationCollection
+     */
+    protected function inheritedAssociations()
+    {
+        $innerCollection = clone $this->innerCollection;
```
fquffio, line 76: Mindfuck question: what happens if `$this->innerCollection` is itself a BEdita `AssociationCollection`? It is cloned, but `$this->innerCollection->innerCollection` and `$innerCollection->innerCollection` still point to the same instance. 😱 So if an association removal cascades, it is actually removed from the actual AssociationCollection, right? 🤔(Don't ask me to read it again, because I wouldn't understand what I wrote myself.)

Maybe we should add something like this:
```php
public function __clone()
{
    $this->innerCollection = clone $this->innerCollection;
}
```

plugins/BEdita/Core/tests/TestCase/Model/Action/ListObjectsActionTest.php
```diff
@@ -14,6 +14,7 @@
 namespace BEdita\Core\Test\TestCase\Model\Action;
 
 use BEdita\Core\Model\Action\ListObjectsAction;
+use BEdita\Core\Test\Fixture\ObjectsFixture;
```
fquffio: Unused import.

plugins/BEdita/Core/tests/TestCase/Model/Entity/TreeTest.php
```diff
@@ -0,0 +1,109 @@
...
+        'plugin.BEdita/Core.object_relations',
+        'plugin.BEdita/Core.profiles',
+        'plugin.BEdita/Core.users',
+        'plugin.BEdita/Core.trees',
+    ];
+
+    /**
+     * Trees table
+     *
+     * @var \BEdita\Core\Model\Table\TreeTable
```
fquffio: [minor] Typo: `TreesTable`

plugins/BEdita/Core/tests/TestCase/Model/Table/FoldersTableTest.php
```diff
@@ -0,0 +1,273 @@
...
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
+namespace BEdita\Core\Test\TestCase\Model\Table;
+
+use BEdita\Core\Model\Table\FoldersTable;
```
fquffio: Unused import.

plugins/BEdita/Core/tests/TestCase/Model/Table/TreesTableTest.php
```diff
@@ -1,11 +1,30 @@
...
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
 namespace BEdita\Core\Test\TestCase\Model\Table;
 
+use BEdita\Core\Model\Table\TreesTable;
```
fquffio: Unused import.

plugins/BEdita/Core/tests/TestCase/Model/Table/TreesTableTest.php
```diff
@@ -56,29 +82,134 @@ public function tearDown()
...
+        $entity = $this->Trees->newEntity();
+        $entity->parent_id = $parentId;
+        static::assertEquals($expected, $this->Trees->isParentValid($entity));
     }
 
     /**
-     * Test buildRules method
+     * Data provider for `testChangeRoot()`
      *
      * @return void
```
fquffio: [minor] `@return array`

plugins/BEdita/Core/tests/TestCase/Model/Table/TreesTableTest.php
```diff
@@ -56,29 +82,134 @@ public function tearDown()
...
+        $entity->parent = $Folders->get(12);
+
+        $Folders->save($entity);
+
+        $parentNode = $this->Trees
+            ->find()
+            ->where(['object_id' => $entity->parent->id])
+            ->first();
+
+        $node = $this->Trees
```
fquffio: This value is not used anywhere… maybe an assertion is missing?

plugins/BEdita/Core/tests/TestCase/ORM/Association/RelatedToTest.php
```diff
@@ -126,4 +126,63 @@ public function testGetSubQueryForMatching(array $expected, $table, $association
 
         static::assertEquals($expected, $result);
     }
+
+    /**
+     * Data provider for `testIsSourceAbstract()`
+     *
+     * @return boolean
```
fquffio: [minor] `@return array`

fquffio (changes requested):
Great work!

Great work! I requested some changes, but they're not strictly required… we might merge right now and address some of these later. I might take care of typos and some other trivial changes, if you wish.

plugins/BEdita/Core/src/Model/Entity/Tree.php
```diff
@@ -1,38 +1,116 @@
...
+     * Set the related `parent_id` if changed.
+     * In this way `parent_node_id` will be marked as dirty (see `self::_setParentId()`)
+     * ensuring the recover of the tree nested set model.
+     *
+     * @param \BEdita\Core\Model\Entity\Folder|null $folder The folder entity to set as parent
+     * @return \BEdita\Core\Model\Entity\Folder|null
+     */
+    protected function _setParentObject(Folder $folder = null)
+    {
+        $parentId = ($folder === null) ? null : $folder->id;
```
fquffio: I'm not a super fan of ternary operators… they make line coverage less relevant. I'd prefer:

```php
$parentId = null;
if ($folder !== null) {
    $parentId = $folder->id;
}
```

plugins/BEdita/Core/src/ORM/Inheritance/AssociationCollection.php
batopa: :scream: I think you are right! I'll try your approach and make a unit test on it.

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
batopa: Ok, not too nasty mate, I like it ;)
But...
The first `if` was there to avoid a SELECT on `trees` if no `parent` was set on `Folder` entity. What is better? 
If it's not important I would go with your solution.

plugins/BEdita/Core/src/Model/Table/TreesTable.php
batopa: Good point!

plugins/BEdita/Core/src/Model/Table/TreesTable.php
batopa: :thinking: Without that check the following `find()` (that will become `exists()`) see https://github.com/bedita/bedita/pull/1408#discussion_r175834784) will return `false`, but a `parent_id === null` is valid value for root folder.

plugins/BEdita/Core/tests/TestCase/Model/Table/TreesTableTest.php
batopa: Naaaa... not needed, I remove it

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
fquffio: Uhm… yeah, you're right, I hadn't thought about that. I don't know… we would be saving one `SELECT` only for root folders, which is not a huge optimisation, maybe. Furthermore, if for some reason `$entity->_properties['parent']` has not been loaded, it would fall into that `if` even if it shouldn't.

I 'm not sure which is the better approach…

plugins/BEdita/Core/src/Model/Table/TreesTable.php
fquffio: No, I'd leave this check, but instead of returning `true` straight away, we should probably do something like:
```php
if ($entity->parent_id === null) {
    return $Objects->exists(['id' => $entity->object_id, 'object_type_id' => $foldersOTID]);
}
```

Or is this check only performed for folder nodes?

batopa: Ahhhh ok, I understand. It's not only for folder node, is used as applcation rule for every node so you are right :+1:

batopa:
@fquffio all requested changes should be done except https://github.com/bedita/bedita/pull/1408#discussion_r175832435

If all is ok we might merge and think about it later. Do you agree?

fquffio (approved):

Issue #1405, opened 2018-01-31
JSON Schema cache


stefanorosanelli:
Cache JSON Schema response containing revision after #1404

This cache should be invalidated when:
* object type is modified
* migration script is launched
* related properties are changed/added/removed

stefanorosanelli:
Done in #1409

PR #1409, opened 2018-02-02
Add schema revision and cache

### LINKED ISSUE #1404: Add revision to JSON Schema
Add a revision hash/number to json schema representation of types.

Simpler way seems:
* custom `revision` property in `/model/schema/{type}` response - we should have our own `$schema` for that, but we are not forced to (we'll do it later fi necessary)
* md5 hash of the schema data array as revision

stefanorosanelli:
This PR fixes #1404 and #1405 

* custom property `revision` added to schema with a `crc32` hash of schema data
* type cache added using `_bedita_object_types_` config with same invalidation rules

### DIFF
--- a/plugins/BEdita/Core/src/Utility/JsonSchema.php
+++ b/plugins/BEdita/Core/src/Utility/JsonSchema.php
@@ -15,6 +15,8 @@ namespace BEdita\Core\Utility;
 
 use BEdita\Core\Model\Entity\ObjectType;
 use BEdita\Core\Model\Entity\StaticProperty;
+use BEdita\Core\Model\Table\ObjectTypesTable;
+use Cake\Cache\Cache;
 use Cake\Datasource\Exception\RecordNotFoundException;
 use Cake\Network\Exception\NotFoundException;
 use Cake\ORM\TableRegistry;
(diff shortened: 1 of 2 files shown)


fquffio:
Code is 👍 .

I just have a doubt… I don't quite like adding a "revision" field to the resulting schema… I'd rather serve the schema with an `ETag` header, and include the revision in the resources' response payload.

E.g.:

```http
GET /model/schema/users
Content-Type: application/vnd.schema+json
ETag: W/"abcd"
```

```http
GET /users/1

{
  "data": { ... },
  "meta": {
    "schema-revisions": {
      "users": "abcd"
    }
  }
}
```

This would have the benefit of HTTP caching, and the schema would remain "clean".

plugins/BEdita/Core/src/Utility/JsonSchema.php
```diff
@@ -86,6 +95,22 @@ public static function typeSchema($typeName)
...
+     *
+     * @param array|bool $schema Schema array or `false`
+     * @return array|bool Schema with `revision` or `false`
+     */
+    protected static function addRevision($schema)
+    {
+        if (!is_array($schema)) {
+            return $schema;
+        }
+        $schema['revision'] = sprintf("%u", crc32(json_encode($schema)));
```
fquffio, line 109: Why CRC32? This is useful if we want to convert this value to a number to do some math with it, because it produces a 32-bit hash that can be safely converted to an integer on 64-bit machines, but if it has to be an opaque string we might use SHA1…

stefanorosanelli:
Having `revision`in response body may be easier to read and handle for a client in both cases, `GET /model/schema/users` and `GET /users/1`

Having an `ETag` header is of course a good idea - I would use both
This way a client can decide to use HTTP or custom cache

plugins/BEdita/Core/src/Utility/JsonSchema.php
stefanorosanelli: Just to have a short, simple and fast hash. 
Might use other algorithms of course.

didoda (approved):

PR #1411, opened 2018-02-09
Create/modify not allowed on abstract endpoints

stefanorosanelli:
This PR fixes object types endpoints **allowed methods**: abstract types may not be created (POST) or modified (PATCH)


fquffio (approved):
I'm not quite sure that `POST` and `PATCH` are actually inhibited. We can't `POST` an object with `"type": "objects"`, but we can `POST` an image on `/objects`. 🤔

stefanorosanelli:
On `POST /objects` with `"type": "objects"` we get a 403.
On `POST /objects` with `"type": "images"` we get a 409.
Both seem correct. Or am I missing something?

fquffio:
Ok, sorry, you're right. I called the wrong API with the wrong data. 😞

PR #1413, opened 2018-02-10
`meta.schema` info on JSON API responses (#1406)

### LINKED ISSUE #1406: Schema info in response
Every response containing types in `"data"` should contain some sort of schema information.

Proposal example:

```json
"meta": {
  "schema": {
    "images": {
       "href": "https://api.example.com/model/schema/images",
       "revision": "sfe234fs"
    },
    "streams": {
       "href": "https://api.example.com/model/schema/streams",
       "revision": "th67rv24"
    },
  }
}

```

stefanorosanelli:
This PR fixes #1406

On every JSON API response schema information on represented types is added, where URL and revision of schema is displayed.

A section is added to main `meta` reponse like this:
```json
"meta": {
  "schema": {
    "users": {
      "$id": "http://api.example.com/model/schema/users"
      "revision": "2112132969"
    }
  }
}
``` 
Schema info of all types included in a response will be presented, 
 
This information is added only for types having a `JSON Schema` representation, some internal resources like `properties`, `applications` and `object_types` are not supported - see `/model` and `/admin` endpoints.

### DIFF
--- a/plugins/BEdita/API/src/TestSuite/IntegrationTestCase.php
+++ b/plugins/BEdita/API/src/TestSuite/IntegrationTestCase.php
@@ -99,7 +99,6 @@ abstract class IntegrationTestCase extends CakeIntegrationTestCase
             'className' => 'Debug'
         ]);
 
-        JsonApi::resetSchemaInfo();
         EventManager::instance()->on(new CommonEventHandler());
     }
 
--- a/plugins/BEdita/API/src/TestSuite/IntegrationTestCase.php
+++ b/plugins/BEdita/API/src/TestSuite/IntegrationTestCase.php
@@ -14,7 +14,6 @@
 namespace BEdita\API\TestSuite;
 
 use BEdita\API\Event\CommonEventHandler;
-use BEdita\API\Utility\JsonApi;
 use BEdita\Core\State\CurrentApplication;
 use BEdita\Core\Utility\LoggedUser;
 use Cake\Event\EventManager;
--- a/plugins/BEdita/API/src/TestSuite/IntegrationTestCase.php
+++ b/plugins/BEdita/API/src/TestSuite/IntegrationTestCase.php
@@ -14,6 +14,7 @@
 namespace BEdita\API\TestSuite;
 
 use BEdita\API\Event\CommonEventHandler;
+use BEdita\API\Utility\JsonApi;
 use BEdita\Core\State\CurrentApplication;
 use BEdita\Core\Utility\LoggedUser;
 use Cake\Event\EventManager;
@@ -49,6 +50,8 @@ abstract class IntegrationTestCase extends CakeIntegrationTestCase
         'plugin.BEdita/Core.external_auth',
         'plugin.BEdita/Core.object_types',
         'plugin.BEdita/Core.objects',
+        'plugin.BEdita/Core.locations',
+        'plugin.BEdita/Core.media',
         'plugin.BEdita/Core.profiles',
         'plugin.BEdita/Core.users',
         'plugin.BEdita/Core.roles',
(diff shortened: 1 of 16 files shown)


plugins/BEdita/API/src/TestSuite/IntegrationTestCase.php
```diff
@@ -96,6 +99,7 @@ public function setUp()
             'className' => 'Debug'
         ]);
 
+        JsonApi::resetSchemaInfo();
```
fquffio: Wouldn't it be better to call this in `JsonApiView` instead?

plugins/BEdita/API/src/Utility/JsonApi.php
```diff
@@ -24,6 +26,30 @@
...
+     * @var array
+     */
+    protected static $schema = [];
+
+    /**
+     * Resource types not having a JSON Schema.
+     *
+     * @var array
+     */
+    protected static $noSchema = [
```
fquffio: There is nothing hardcoded in this class, but this. Are we sure this is the only suitable option? Couldn't we use `JsonApiTrait`s instead to expose schema detail (e.g.: `getJsonSchema(): ?array`), and infer types without a JSON Schema from that?

plugins/BEdita/API/src/Utility/JsonApi.php
```diff
@@ -65,13 +91,81 @@ public static function formatData($items, $options = 0, array $fields = [], arra
...
+     * @param string $type Type name
+     * @return array
+     */
+    public static function schemaInfo($type)
+    {
+        if (!empty(static::$schema[$type])) {
+            return static::$schema[$type];
+        }
+
+        static::$schema[$type] = [
```
fquffio: This static class does not store any runtime information but this. I'd avoid if possible.

Furthermore, in a real case scenario this kind of runtime cache is completely useless, since in line 109 types are already made unique by means of `array_unique()`, thus this would likely never result in a cache hit.

I'd remove the static property `$schema` and simply return the requested result when asked. This wouldn't affect performance, and it would let us remove `resetSchemaInfo()` method, thus making this class' APIs cleaner, and more straightforward to test.

plugins/BEdita/API/tests/TestCase/Utility/JsonApiTest.php
```diff
@@ -536,11 +642,40 @@ public function testFallbackLinks()
...
         static::assertEquals($expected, $result);
     }
+
+    /**
+     * Test `schemaInfo` method
+     *
+     * @return void
+     * @covers ::schemaInfo
+     * @covers ::resetSchemaInfo
+     */
```
didoda: add `@covers ::schemaRevision`? maybe not necessary: method is already covered by `testSchemaRevision`

didoda (approved):

fquffio (changes requested):

plugins/BEdita/API/src/Utility/JsonApi.php
stefanorosanelli: It was a quick and dirty solution. We can do better :wink:  
I'll try something like your proposal.

plugins/BEdita/API/src/Utility/JsonApi.php
stefanorosanelli: You're right, we could probably avoid this and have a cleaner solution.

plugins/BEdita/API/src/TestSuite/IntegrationTestCase.php
stefanorosanelli: Unfortunately this is needed in tests not involving `JsonApiView` but if we get rid of `resetSchemaInfo()` (see comments below) we will drop this one too.

plugins/BEdita/API/tests/TestCase/Utility/JsonApiTest.php
stefanorosanelli: you mean `JsonSchema::schemaRevision` ? 
it's already covered in its own test case class, don't needed here

fquffio (approved):

PR #1414, opened 2018-02-15
ETag header on json schema

stefanorosanelli:
This PR adds ETag header to `/model/schema/{type}` response, using `revision` as its value.

A 304 Not Modified empty response is provided on  a `'If-None-Match'` matching request.

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Controller/Model/SchemaControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/Model/SchemaControllerTest.php
@@ -14,6 +14,7 @@
 namespace BEdita\API\Test\TestCase\Controller\Model;
 
 use BEdita\API\TestSuite\IntegrationTestCase;
+use BEdita\API\Test\TestConstants;
 
 /**
  * {@see \BEdita\API\Controller\Model\SchemaController} Test Case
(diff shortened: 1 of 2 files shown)


plugins/BEdita/API/tests/TestCase/Controller/Model/SchemaControllerTest.php
```diff
@@ -103,4 +104,26 @@ public function testJsonSchemaAbstractType()
...
         static::assertFalse($result);
     }
+
+    /**
+     * Test ETag response header and Not Modified response.
+     *
+     * @return void
+     *
+     * @covers ::jsonSchema()
+     */
```
didoda, line 114: `@covers ::generate()` too?

didoda (approved):

plugins/BEdita/API/tests/TestCase/Controller/Model/SchemaControllerTest.php
stefanorosanelli: Since it's a controller test case we better test only `SchemaController` methods.
`JsonSchema::generate()` is tested in its own `JsonSchemaTest` class.

fquffio (approved):

PR #1419, opened 2018-02-20
Bug saving empty date

### LINKED ISSUE #1418: Error saving empty date - CTI problem
Saving an empty date on `users` or `profles` causes a 500 error.
Example:
```http
    PATCH /users/123
    {
    "data": {
        "id": "123",
        "type": "users",
        "attributes": {
            "publish_start": ""
        }
    }
    }
```
Will return
```json
    "error": "500",
    "title": "SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: \u0027\u0027 for column \u0027publish_start\u0027 at row 1",
```
Instead same action on types having only `objects` table works fine as expected:   same `PATCH /documents/123` will give a `200 OK`

Using CTI it seems only last table in the chain (`users` in this example) is considered in marshalling values to entity field columns, other table fields are saved directly with no type marshalling.

batopa:
This PR fixes #1418 

A `Marshaller` class for `\BEdita\Core\ORM\Inheritance` was added. This class takes care to prepare the right property map of the inheritance chain used to marshall entity from data.

### DIFF
--- a/plugins/BEdita/API/tests/IntegrationTest/MarshallInheritedPropertiesTest.php
+++ b/plugins/BEdita/API/tests/IntegrationTest/MarshallInheritedPropertiesTest.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * BEdita, API-first content management framework
- * Copyright 2017 ChannelWeb Srl, Chialab Srl
+ * Copyright 2018 ChannelWeb Srl, Chialab Srl
  *
  * This file is part of BEdita: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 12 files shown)


stefanorosanelli (approved):

plugins/BEdita/API/tests/IntegrationTest/MarshallInheritedPropertiesTest.php
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
+namespace BEdita\API\Test\IntegrationTest;
+
+use BEdita\API\TestSuite\IntegrationTestCase;
+
+/**
+ * Test that the inherited propertiesare correctly marshalled into main entity.
```
didoda: typo: `properties are` [...]

didoda (approved):

PR #1421, opened 2018-02-22
Enable filter on multiple associated items by id

stefanorosanelli:
This PR enables filters on associated items using lists.

`GET /users?filter[roles]=1` lists users having associated role with id=1, already works.

With this PR it will also be possible:
* `GET /users?filter[roles][]=1&filter[roles][]=2` lists users having associated role with id=1 or id=2

This will work for every relation, suppose we have `locations` **described_by** `documents`

 * `GET /locations?filter[described_by][]=11&filter[described_by][]=12` will list locations having a `documents` with id=11 or id=12 related with `described_by`

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/ListEntitiesAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/ListEntitiesAction.php
@@ -104,11 +104,11 @@ class ListEntitiesAction extends BaseAction
             if ($this->Table->associations()->has($camelizedKey)) {
                 // Associated match (primary key only).
                 $target = $this->Table->association($camelizedKey)->getTarget();
-                $targetPrimaryKey = array_map(
-                    [$target, 'aliasField'],
-                    (array)$target->getPrimaryKey()
-                );
-                $conditions = array_combine($targetPrimaryKey, (array)$value);
+                $targetPrimaryKey = $target->aliasField($target->getPrimaryKey());
+                if (is_array($value)) {
+                    $targetPrimaryKey .= ' IN';
+                }
+                $conditions = [$targetPrimaryKey => $value];
 
                 $query = $query
                     ->distinct(array_map(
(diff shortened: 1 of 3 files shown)


plugins/BEdita/Core/src/Model/Action/ListEntitiesAction.php
```diff
@@ -104,11 +104,11 @@ protected function buildFilter(Query $query, array $filter)
             if ($this->Table->associations()->has($camelizedKey)) {
                 // Associated match (primary key only).
                 $target = $this->Table->association($camelizedKey)->getTarget();
-                $targetPrimaryKey = array_map(
-                    [$target, 'aliasField'],
-                    (array)$target->getPrimaryKey()
-                );
-                $conditions = array_combine($targetPrimaryKey, (array)$value);
+                $targetPrimaryKey = $target->aliasField($target->getPrimaryKey());
```
batopa, line 107: :thinking:  I'm not sure that is the right solution. I think the previous code was done for supporting multiple primary keys.

Maybe we would build `$targetPrimaryKey` as previous and set the `IN` condition just if  `count($targetPrimaryKey)` is equal to `1` otherwise I think the `array_combine` should be used.

@fquffio am I wrong?

fquffio: You are right. However, right now we don't use composite primary keys, except for some junction tables that are not "exposed" as standalone resources. Therefore, it should be (quite) safe.

Also, there is another problem here: the `INNER JOIN` strategy was definitely not the best choice in the first place: it makes SQL return multiple times the same row for "to many" associations, and forces us to use `DISTINCT` (see line 114). A better approach would likely involve `EXISTS` (i.e.: `SELECT <fields> FROM <table> WHERE EXISTS (<subquery>)`), but in this case we should build a subquery first — and that subquery would be different according to the type of association.

batopa: OK! I think we can live with it :smile_cat:

batopa (approved):

fquffio (approved):

PR #1424, opened 2018-03-01
Avoid `NotFoundException` on SQlite 

stefanorosanelli:
This PR fixes an error on SQlite

If an *endpoint* is not in `endpoints` table a `NotFoundException` is thrown on every call

A test case has been added in `EndpointAuthorizeTest.php`

### DIFF
--- a/plugins/BEdita/API/src/Auth/EndpointAuthorize.php
+++ b/plugins/BEdita/API/src/Auth/EndpointAuthorize.php
@@ -218,7 +218,7 @@ class EndpointAuthorize extends BaseAuthorize
             $this->endpoint = $Endpoints->newEntity(
                 [
                     'name' => $endpointName,
-                    'enabled' => true,
+                    'enabled' => 1,
                 ],
                 ['validate' => false]
             );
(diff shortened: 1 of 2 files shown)


didoda (approved):

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
```diff
@@ -218,7 +218,7 @@ protected function getEndpoint()
             $this->endpoint = $Endpoints->newEntity(
                 [
                     'name' => $endpointName,
-                    'enabled' => true,
+                    'enabled' => 1,
```
fquffio, line 221: Why? If `true` does not work it means there's a problem.

fquffio (commented):
I don't see how this solves the problem described… 🤔

plugins/BEdita/API/src/Auth/EndpointAuthorize.php
stefanorosanelli: My mistake, got confused testing on a bad sqlite db where `enabled` was `integer` and not `boolean`
Removing this PR

PR #1433, opened 2018-03-20
Chore: LocalAdapter path + test

stefanorosanelli:
This PR fixes default local adapter path configuration using `webroot/_files` as default and avoiding double slashes `//`

Bonus track: temporary folders created in testcase are removed after test execution.

### DIFF
--- a/plugins/BEdita/Core/src/Filesystem/Adapter/LocalAdapter.php
+++ b/plugins/BEdita/Core/src/Filesystem/Adapter/LocalAdapter.php
@@ -30,7 +30,7 @@ class LocalAdapter extends FilesystemAdapter
      */
     protected $_defaultConfig = [
         'baseUrl' => null,
-        'path' => WWW_ROOT . DS . 'files',
+        'path' => WWW_ROOT . '_files',
         'writeFlags' => LOCK_EX,
         'linkHandling' => Local::DISALLOW_LINKS,
         'permissions' => [],
@@ -44,7 +44,7 @@ class LocalAdapter extends FilesystemAdapter
     {
         $success = parent::initialize($config);
 
-        if (empty($this->_config['baseUrl']) && strpos($this->getConfig('path'), WWW_ROOT . DS) === 0) {
+        if (empty($this->_config['baseUrl']) && strpos($this->getConfig('path'), WWW_ROOT) === 0) {
             // Files are stored within the document root, so base URL can be automatically detected if not already set.
             $path = str_replace(DS, '/', substr($this->getConfig('path'), strlen(WWW_ROOT)));
             $this->setConfig('baseUrl', Router::url($path, true));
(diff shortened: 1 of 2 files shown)


didoda (approved):

plugins/BEdita/Core/src/Filesystem/Adapter/LocalAdapter.php
```diff
@@ -44,7 +44,7 @@ public function initialize(array $config)
     {
         $success = parent::initialize($config);
 
-        if (empty($this->_config['baseUrl']) && strpos($this->getConfig('path'), WWW_ROOT . DS) === 0) {
```
fquffio, line 47: The trailing directory separator was there to avoid errors in case there's another directory whose path begins with `BEDITA_ROOT/webroot`. For instance, if i want to store my static files under `BEDITA_ROOT/webroot2` the automatic detection of base URL should fail. However this is quite an edge case, and we may want to ignore it.

plugins/BEdita/Core/tests/TestCase/Filesystem/Adapter/LocalAdapterTest.php
```diff
@@ -23,6 +23,24 @@
  */
 class LocalAdapterTest extends TestCase
 {
+    /**
+     * Temporary test files path, removed after test
+     * @var string
+     */
+    const FILES_PATH = WWW_ROOT . 'static-files' . DS . 'subdir';
```
fquffio, line 30: In `tests/bootstrap.php` we configure `LocalAdapter` to use `plugins/BEdita/Core/tests/uploads` as its root. That configuration is not used in this case, but we might use the same path…

fquffio (approved):

plugins/BEdita/Core/src/Filesystem/Adapter/LocalAdapter.php
```diff
@@ -44,7 +44,7 @@ public function initialize(array $config)
...
+        if (empty($this->_config['baseUrl']) && strpos($this->getConfig('path'), WWW_ROOT) === 0) {
             // Files are stored within the document root, so base URL can be automatically detected if not already set.
             $path = str_replace(DS, '/', substr($this->getConfig('path'), strlen(WWW_ROOT)));
```
fquffio, line 49: ...instead, I'd put `strlen(WWW_ROOT) + 1` or `strlen(WWW_ROOT . DS)` here.

plugins/BEdita/Core/src/Filesystem/Adapter/LocalAdapter.php
stefanorosanelli: You're right but `WWW_ROOT` is defined as `ROOT . DS . 'webroot' . DS` so it already contains dir separator the end.

plugins/BEdita/Core/src/Filesystem/Adapter/LocalAdapter.php
stefanorosanelli: also not necessary since `WWW_ROOT` contains last dir separator

plugins/BEdita/Core/src/Filesystem/Adapter/LocalAdapter.php
fquffio: Oh, right. Sorry, I hadn't noticed that.

PR #1436, opened 2018-03-22
Thumbnails

### LINKED ISSUE #1403: Thumbnail generation
Thumb will be generatend in sync/async mode using
```
GET /media/thumbs/{{id}}?width=100&height=100

GET /media/thumbs/{{id}}?preset=small

GET /media/thumbs?id={{id1}},{{id2}},{{id3}}&preset=medium

```
Expected response
```php
{
    "meta": {
       "thumbs" : [
          {
            "id": "123",
            "uuid": "f47c0592-4cb3-4ab1-a7b6-eb9a84fd4e5e",
            "url": "https://path/to/thumb.png",
            "ready": true
          }
       ]
    }
}
```

if `ready` **true** image thumb is ready, if **false** in processing...

---

Configuration example

```php
 [
   'Thumbs' => [
       // if `true` any format is allowed
       // if `false` only predefined presets are accepted
       'allowAny' => false,
      // available Thumb generators
      'generators' => [
       'default ' => 'BEdita/Core.Simple',
       'async' => 'BEdita/Core.Async',      
     ]
      'presets' => [
        'small' => [
             'width' => 100,
             'height' => 100,
              // optional custom generator for this preset
             //'generator' => 'async', 
        ],
     ]
   ]  
```

fquffio:
This PR closes #1403 by implementing thumbnails generation, and the `GET /media/thumbs` endpoint.

## Usage

Thumbnails can be generated using the `Thumbnail` class:
```php
$stream = TableRegistry::get('Streams')->get($myStreamUuid);

// Provide custom options…
Thumbnail::get($stream, [
  'generator' => 'async',
  'w' => 100,
  'h' => 100,
]);
// …or use one of the available presets.
Thumbnail::get($stream, 'my_preset');
```

Keep in mind that the generator might be asynchronous, so the thumbnail might not be available right away.

## API

Thumbnails can be requested by the client using the `GET /media/thumbs/:id` or `GET /media/thumbs?ids=<comma-separated-ids>` endpoints. A preset can be specified using the `?preset=<preset-name>` query parameter, or (if explicitly allowed via configuration) a whole set of options — width, height, format, …

```http
GET /media/thumbs?ids=1,2,3&preset=async
```

```http
HTTP/1.1 200 OK

{
  // ...
  "meta": {
    "thumbnails": [
      {
        "id": 1,
        "uuid": "some-uuid-1",
        "ready": false,
        "url": "http://example.com/something.jpg"
      },
      {
        "id": 2,
        "uuid": "some-uuid-2",
        "ready": true,
        "url": "http://example.com/something-else.jpg"
      },
      {
        "id": 3,
        "uuid": "some-uuid-3",
        "ready": false,
        "acceptable": false,
        "url": "http://example.com/something.txt"
      }
    ]
  }
}
```

## Generators

This PR implements two thumbnail generators: **Glide** and **Async**.

- **Glide** generator synchronously generates a thumbnail using [Glide](http://glide.thephpleague.com).
- **Async** generator enqueues a job to be processed by `ThumbnailService` that will eventually invoke another generator to actually generate the thumbnail. 

## Questions

- Are the options in the format `w` (width), `h` (height), `fm` (format), ok? If not, we must "translate" these options inside Glide generator. Be aware that every generator might accept a different set of options, so we might not be able to declare an exhaustive vocabulary at all.
- Right now the capability of actually generating a thumbnail for a Stream isn't checked until we actually try to generate it. This means that asynchronous generators don't return `acceptable: false` for, say, plain text files, while synchronous generators correctly do. We might require the generators to tell if they are capable to generate a thumbnail, of course, but that would make asynchronous generators not-so-asynchronous… can we live with that?

## Bonus tracks

- Tests now use a fixed date and time. This has been implemented calling `Time::setTestNow()` in `tests/bootstrap.php`, and only two tests had to be updated — because they were calling `time()` directly. This makes timestamps generated in tests predictable.
- `PATCH /streams/:uuid/relationships/object` now accepts all subtypes of "media". Before, only "images" was allowed, and "files" wasn't. A regression test has been added.
- `config/app.default.php` has beed more extensively documented, and a couple of pieces of configuration can now be set using environment variables.

## Why so many files changed?

Because I had to add one media to `ObjectsFixture`, and well… you know how these things end.


stefanorosanelli:
* options format: I think we can use `Glide` options as default and translate them to other generator formats once we implement one
* probably `acceptable: false` could be returned directly without invoking async generators but not a big problem anyway... we can prooably change that later

config/app.default.php
```diff
@@ -427,13 +427,61 @@
...
+     */
     'Filesystem' => [
         'default' => [
             'className' => 'BEdita/Core.Local',
             'path' => WWW_ROOT . DS . '_files',
+            'url' => env('FILESYSTEM_DEFAULT_URL', null),
+        ],
+        'thumbnails' => [
+            'className' => 'BEdita/Core.Local',
+            'path' => WWW_ROOT . DS . '_files' . DS . 'thumbs',
```
stefanorosanelli: [minor] this should be `WWW_ROOT . '_files' . DS . 'thumbs'`, same applies to `default.path`

plugins/BEdita/API/src/Controller/MediaController.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Get and validate IDs in request URL.
+     *
+     * @return int[]
+     */
+    protected function getIds()
+    {
+        $id = $this->request->getParam('id');
+        $ids = $this->request->getQuery('ids');
```
stefanorosanelli: do we have a limit on max number of `ids` passed? 
should this be somehow pagination limit?

plugins/BEdita/Core/src/Filesystem/Thumbnail/GlideGenerator.php
```diff
@@ -0,0 +1,164 @@
...
+    public function generate(Stream $stream, array $options = [])
+    {
+        $path = $this->getFilename($stream, $options);
+
+        try {
+            $thumbnail = $this->makeThumbnail($stream, $options);
+
+            FilesystemRegistry::getMountManager()->put($path, $thumbnail);
+        } catch (NotReadableException $e) {
+            throw new InvalidStreamException(__('Unable to generate thumbnail for stream {0}', $stream->uuid), null, $e);
```
stefanorosanelli: should be `[$stream->uuid]`

config/app.default.php
stefanorosanelli: [minor] this should be `WWW_ROOT . '_files' . DS . 'thumbs'`, also `default.path` should be changed accrodingly

plugins/BEdita/API/src/Controller/MediaController.php
stefanorosanelli, line 37: is there a limit on `'ids'` number? 
should not exceed pagination limit...

plugins/BEdita/Core/src/Job/Service/ThumbnailService.php
```diff
@@ -0,0 +1,55 @@
...
+            /** @var \BEdita\Core\Model\Table\StreamsTable $table */
+            $table = TableRegistry::get('Streams');
+            $stream = $table->get(Hash::get($payload, 'uuid'));
+
+            $generator = Thumbnail::getGenerator(Hash::get($payload, 'generator', 'default'));
+            $generator->generate($stream, (array)Hash::get($payload, 'options'));
+
+            return true;
+        } catch (RecordNotFoundException $e) {
+            // Stream not found, mark job as complete anyway.
```
stefanorosanelli, line 46: should we `Log` something or is this a normal flow?

stefanorosanelli (changes requested):
Great job!!
Just a couple of minor problems or clarifications

plugins/BEdita/Core/src/Filesystem/Thumbnail/GlideGenerator.php
fquffio: According to [`__()` function implementation](https://github.com/cakephp/cakephp/blob/master/src/I18n/functions.php#L22) it should be equivalent. Do you prefer to stick to this convention? I'm quite sure there are examples of both syntaxes in our code.

**Multiple arguments**: https://github.com/bedita/bedita/blob/4-cactus/plugins/BEdita/Core/src/Model/Action/RemoveAssociatedAction.php#L82
**Array**: https://github.com/bedita/bedita/blob/4-cactus/plugins/BEdita/Core/src/Model/Action/ChangeCredentialsAction.php#L95

fquffio: (BTW it should become `__d('bedita', ...)` anyways, at least)

plugins/BEdita/API/src/Controller/MediaController.php
fquffio: Right, I'm adding this check.

plugins/BEdita/Core/src/Job/Service/ThumbnailService.php
fquffio: This happens when you enqueue a thumbnail-ization job for a given stream, and then delete that stream _before_ the job is processed. Quite unusual, to be fair, but the job should not be retried, because it is guaranteed to fail again. We still may want to log something — if it happens too often there almost certainly is a problem — but I wouldn't treat it as an error… maybe something like this?

```php
Log::debug(sprintf('Thumbnail service could not find stream %s', $stream->uuid));
```

config/app.default.php
fquffio: Yep, thanks!

plugins/BEdita/API/src/Controller/MediaController.php
fquffio: Actually there's the option of paginating the results.

E.g.: `GET /media/thumbs?ids=1,2,3,4&page_size=2` would return two pages of results, assuming each of those Media has one associated Stream. Query parameters for pagination would be the same as usual.

Which do you prefer? Paginated results or **400 Bad Request** if `count($ids) > Configure::read('Pagination.maxLimit')`?

didoda: I'd say `400 Bad Request`

didoda (approved):

plugins/BEdita/API/src/Controller/MediaController.php
stefanorosanelli: +1 on  `400 Bad Request`

plugins/BEdita/Core/src/Job/Service/ThumbnailService.php
stefanorosanelli: Seems ok, I'd use `Log::warn`since it can be related to an error but also `Log::info` is acceptable here...

plugins/BEdita/Core/src/Filesystem/Thumbnail/GlideGenerator.php
stefanorosanelli: Yes, both should work. 
Scrutinizer is suggesting the array format. We may want to choose one form anyway... it's ok

stefanorosanelli (approved):

Issue #1428, opened 2018-03-16
Password and login policy


stefanorosanelli:
Implement basic password and login policy.

In BE3 we had this configuration defaults:
```php
$config['loginPolicy'] = array (
	"maxLoginAttempts" => 10,
	"maxNumDaysInactivity" => 180,
	"maxNumDaysValidity" => 60,
	"passwordRule" => "/\w{3,}/", // regexp to match for valid passwords (empty => no regexp)
	"passwordErrorMessage" => "Password must contain at least 3 valid alphanumeric characters", // error message for passwrds not matching given regexp
);

```
Similar settings could go in an `Auth` configuration key

Those settings should be available to client applications in order to perform some client validation.

stefanorosanelli:
Password policy has been implemented, other policies like `maxLoginAttempts`, `maxNumDaysInactivity` or `maxNumDaysValidity` will be implemented in a new issue

PR #1437, opened 2018-03-26
Passwd policy

stefanorosanelli:
This PR implements first part of #1428 

Simple password format validation rule was added via new `Auth` key config,


batopa:
Just an observation:

I would prefer something as
```
'Auth' => [
     'passwordPolicy' => [
          'rule' => // anonymous function or class name to use for validation or regexp...,
           'message' => '',
     ]
]
```

Then we could use it in `\BEdita\Core\Model\Validation\UsersValidator` to setup a custom validation rule.

But I think we can live with it for now.

stefanorosanelli:
Ok for the new config, but in this particular case we need to have a check before validation rules: at validation step password is already encrypted.
I had to use `beforeMarshal()` for that reason...

batopa:
Uhm... validation rules (not application rules) should be applied before data is converted into entity. Do we have already hashed password?

Go with the config anyway.

batopa (approved):

didoda (approved):

batopa:
ok I merge and open an issue to move check at data validation level.

PR #1439, opened 2018-03-26
Roles on signup

### LINKED ISSUE #1438: Signup with given roles
Upon `/signup` request we may want to assign some roles to new user.

This roles MUST be available in `Signup.roles` configuration
[this config will normally be defined at application level]

stefanorosanelli:
This PR fixes #1438 

* on `/signup` request using a `roles` array it is now possible to set roles on a newly created user
* requested roles MUST be in `Signup.roles` config and cannot contain `admin` role

Note: `SignupUserActionTest` class needs a refactor since there are many methods and only global  `@covers` at class level  => we should cover as much as possible single methods...

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
@@ -91,9 +91,10 @@ class SignupUserAction extends BaseAction implements EventListenerInterface
 
         // operations are not in transaction because AsyncJobs could use a different connection
         $user = $this->createUser($data['data']);
-        // add roles to user, with validity check
-        $this->addRoles($user, $data['data']);
         try {
+            // add roles to user, with validity check
+            $this->addRoles($user, $data['data']);
+
             $job = $this->createSignupJob($user);
             $activationUrl = $this->getActivationUrl($job, $data['data']);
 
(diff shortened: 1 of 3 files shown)


plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -173,6 +184,45 @@ protected function createUser(array $data)
...
+     * Load requested roles entities with validation
+     *
+     * @param array $roles Requested role names
+     * @return \BEdita\Core\Model\Entity\Role[] requested role entities
+     * @throws \Cake\Network\Exception\BadRequestException When role validation fails
+     */
+    protected function loadRoles(array $roles)
+    {
+        $entities = [];
+        $allowed = (array)Configure::read('Signup.roles');
```
batopa, line 215: I think a better place to check roles should be an application rule. Since `SignupUserAction` implements `EventListenerInterface` it should be easy as

```
/**
 * {@inheritdoc}
 */
public function implementedEvents()
{
    return [
        'Auth.signup' => 'sendMail',
        'Model.buildRules' => 'isRolesValid',
    ];
}
```

and implement `isRolesValid` method.

batopa (changes requested):
Overall it's ok for me. I just require a change on roles check but only if it come easy.

batopa (approved):
Ok for now

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
stefanorosanelli: we'll do that after merge, if not a problem

PR #1446, opened 2018-04-05
Allow the use of uname instead of id for objects API

batopa:
This PR allows the use of `uname` instead of `id` for API requests on objects endpoints.

```http
GET /documents/doc-uname
GET /documents/doc-uname/:relationship
GET /documents/doc-uname/relationships/:relationship
```

### DIFF
--- a/plugins/BEdita/API/src/Controller/ObjectsController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectsController.php
@@ -185,6 +185,7 @@ class ObjectsController extends ResourcesController
     {
         $this->request->allowMethod(['get', 'patch', 'delete']);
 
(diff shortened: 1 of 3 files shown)


fquffio (approved):

stefanorosanelli (approved):

fquffio:
In the future I'd consider returning 302 instead of 200 in such cases, as it sounds a little more REST-compliant to me. So:

```http
GET /objects/my-unique-name
```

```http
HTTP/1.1 302 Moved Temporarily
Location: /objects/42
```

Same would apply to object types, roles, …

stefanorosanelli:
agreed, same strategy on `resource` types using `name` instead of `uname`

PR #1448, opened 2018-04-09
Add default `title` for profiles and users

stefanorosanelli:
This PR adds a default `title` like a `display name` or `fullname` field for `Profiles` and `Users` if no `title` is set.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
@@ -110,7 +110,7 @@ class ProfilesTable extends Table
         if (empty($entity->get('email'))) {
             $entity->set('email', null);
         }
-        if (empty($entity->get('title')) && (!empty($entity->get('name')) || !empty($entity->get('surname')))) {
+        if (!$entity->has('title') && ($entity->has('name') || $entity->has('surname'))) {
             $title = sprintf('%s %s', (string)Hash::get($entity, 'name', ''), (string)Hash::get($entity, 'surname', ''));
             $entity->set('title', $title);
         }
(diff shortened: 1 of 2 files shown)


batopa (approved):

plugins/BEdita/Core/tests/TestCase/Model/Table/ProfilesTableTest.php
```diff
@@ -249,5 +249,6 @@ public function testBeforeSave()
         $success = $this->Profiles->save($profile);
         static::assertTrue((bool)$success);
         static::assertNull($success->get('email'));
+        static::assertEquals('Gustavo Supporto', $success->get('title'));
```
batopa, line 252: Maybe I would add another test with `title` field populated.

plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
```diff
@@ -105,5 +110,9 @@ public function beforeSave(Event $event, EntityInterface $entity)
         if (empty($entity->get('email'))) {
             $entity->set('email', null);
         }
+        if (empty($entity->get('title')) && (!empty($entity->get('name')) || !empty($entity->get('surname')))) {
```
fquffio: If possible I'd use `!$entity->has('title')` instead of `empty($entity->get('title'))`, and same with "name" and "surname". It's not exactly the same meaning, though, and it would behave differently with empty strings, but it would allow a forcibly empty title.

plugins/BEdita/Core/tests/TestCase/Model/Table/ProfilesTableTest.php
stefanorosanelli: ok, I'll do it

plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
stefanorosanelli: done

plugins/BEdita/Core/tests/TestCase/Model/Table/ProfilesTableTest.php
stefanorosanelli: done

PR #1450, opened 2018-04-12
 Avilable types for folders relationships #1449

### LINKED ISSUE #1449: "available" link on `children/parent/parents` relations
Response of calls like `GET /documents/9/parents` or `GET /folders/2/childrens` contains a `links` section with `"available": null` -> it should contains a link to available object types for those relations.

That is: `folders` for `parent` and `parents` and `objects` for `children`

stefanorosanelli:
This PR fixes #1449 

`"available"` types link URL is resolved with a simple controller config.
Can be done better in many ways I guess....


plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -352,29 +360,41 @@ protected function getAvailableUrl($relationship)
             return $available;
         }
 
+        $types = $this->getAvailableTypes($relationship);
```
fquffio, line 361: I'd only add a check:
```php
if (empty($types)) {
    return null;
}
```

fquffio (approved):

plugins/BEdita/API/src/Controller/ObjectsController.php
stefanorosanelli: :thinking: since it's  an object relationship is it valid and/or safe to have no related types?

should this case raise an error?

stefanorosanelli: change committed, without a test case

fquffio: It's probably something that should never happen. I personally can't think of a case where this list is empty. Yet, I'd prefer to have an explicit `null` instead of a malformed URL.

batopa (approved):

Issue #1431, opened 2018-03-19
Folder API write operations


stefanorosanelli:
After  #1402 and #1408 some API operations are still missing:

 * place an object inside a folder, or more folders (?)
 * remove an object from a folder
 * move an object from a folder to another

batopa:
In addition to writing operations also

```http
GET /folders/:id/parent
```
doesn't work responding with `404`

```json
{
    "error": {
        "status": "404",
        "title": "Relationship \"parent\" does not exist"
    }
}
```

Moreover

```http
GET /folders/:id/parents
```
should return `404 Not found` but it returns `200` with an array of one parent.

PR #1451, opened 2018-04-13
#1431 GET /folders/:id/parent is here

batopa:
This PR partially cover  #1431 implementing `GET /folders/:id/parent` 

The write operation is almost done but missing a key point.

### DONE

1. allow method for `/folders/:id/relationship/parent` is limited to `GET` and `PATCH` (you can only modify an existent parent)
2. The `PATCH` works... but

### TODO

The `PATCH` should be update the current row on `trees` table but it first remove the current row and then insert a new row. Deleting the row all children are lost :disappointed: 

-----

#### Side note

I expressly marked as **incomplete** `FoldersControllerTest::testDelete()` because I don't know what policy we want to apply deleting a folder. There are different possibilities to treat in different PR if you agree.

1. a folder can be deleted only if it doesn't have folders as children.
2. a folder can be always deleted and the children of type `folders` become orphans.
3. other...

Another thing is about JSON API specs. For root folder `GET /folders/:id/parent` return (response simplified)

```json
{
    "data": []
}
```

From specs I read
> Consider, for example, a request to fetch a to-one related resource link. This request would respond with null when the relationship is empty (such that the link is corresponding to no resources) but with the single related resource’s resource object otherwise.

In our case `parent` is `to-one` related resource so the correct response should be

```json
{
    "data": null
}
```

Maybe we need to consider to fix it (I think in ` BEdita\API\Utility\JsonApi::formatData()`).

### DIFF
--- a/plugins/BEdita/API/config/routes.php
+++ b/plugins/BEdita/API/config/routes.php
@@ -27,6 +27,7 @@ Router::plugin(
             'streams',
             'users',
             'media',
+            'folders',
         ];
         $adminControllers = [
             'applications',
--- a/plugins/BEdita/API/src/Controller/FoldersController.php
+++ b/plugins/BEdita/API/src/Controller/FoldersController.php
@@ -1,0 +1,97 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2018 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+namespace BEdita\API\Controller;
+
+use BEdita\Core\Model\Action\ListRelatedFoldersAction;
+use Cake\Network\Exception\NotFoundException;
+use Cake\ORM\Association;
+
+/**
+ * Controller for `/folders` endpoint.
+ *
+ * The main aim is to bridge `parent` API relationship to `parents` BTM association.
+ *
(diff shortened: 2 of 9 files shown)


didoda (approved):

stefanorosanelli (approved):

stefanorosanelli:
We should fix the `PATCH` problem in another issue. Of course in case of `folder move` children structure MUST be kept.

Speaking of delete action, this is how I see it:

- a folder can always be removed, it can't only if `locked` or on missing permissions
- subfolders are automatically deleted, they don't become orphans
- other contained `leaf objects` not contained in other folders: can become orphan or may also be deleted with a special parameter/option

fquffio:
> The PATCH should be update the current row on trees table but it first remove the current row and then insert a new row. Deleting the row all children are lost 😞

How about replacing `Objects->Folders` association with `Objects->Trees` in `*AssociatedAction` when dealing with such cases?

For what concerns nodes deletion, I agree that deleting a folder should delete all its subfolders. This may apply to both soft- and hard-delete, or (IMHO better) to hard-delete only — if it was applied to soft-delete as well, restoring a folder should also restore all its subfolders, but this might lead to unexpected results if one of its subfolders had been soft-deleted previously: soft-delete + restore would result in a different situation from the starting point. I wouldn't cascade delete on all children objects, though, as it seems a bit dangerous to me.

fquffio (approved):

batopa:
> How about replacing Objects->Folders association with Objects->Trees in *AssociatedAction when dealing with such cases?

Yes, good point! However since the action uses [`\Cake\ORM\Association\BelongsToMany::replaceLinks()`](https://api.cakephp.org/3.5/class-Cake.ORM.Association.BelongsToMany.html#_replaceLinks) I would investigate first why join rows already present aren't updated but deleted and inserted.

fquffio:
Agree. If that's something that can be done later in time we can merge and investigate.

PR #1452, opened 2018-04-13
fix: card always reachable

didoda:
This forces api validator `isReachable` to true when object type is a card.
Object validation in api apply to relations too.
When a card is used in a relation, api validator check card id too.
Card is not an object in tree, so it's not reachable itself.
Card can (or cannot) have related data on tree.
However, IMHO, `isReachable` should be ignored for cards.

### DIFF
--- a/bedita-app/controllers/components/api_validator.php
+++ b/bedita-app/controllers/components/api_validator.php
@@ -376,17 +376,8 @@ class ApiValidatorComponent extends Object {
         }
         // User card always reachable
         $user = $this->controller->ApiAuth->getUser();
-        if (!empty($user['id'])) {
-            $card = ClassRegistry::init('ObjectUsers')->find('first', array(
-                'fields' => array('object_id'),
-                'conditions' => array(
-                    'user_id' => $user['id'],
-                    'switch' => 'card',
-                ),
-            ));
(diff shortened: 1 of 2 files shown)


fquffio (changes requested):
I don't quite like the idea that _all_ cards are always reachable.

Instead, I'd make reachable only the card of the currently logged-in user.

It should be possible to do something like this:
```php
$user = $user = $this->controller->ApiAuth->getUser();
if (!empty($user['id'])) {
    $card = ClassRegistry::init('ObjectUsers')
        ->find('first', [
            'conditions' => [
                'user_id' => $user['id'],
                'switch' => 'card',
            ],
        ]);
    if ($card && $card['object_id'] === $objectId) {
        return true;
    }
}
```

fquffio (approved):

bedita-app/controllers/components/api_validator.php
```diff
@@ -374,6 +374,20 @@ public function isObjectReachable($objectId, $checkPermissions = true) {
         if (!$this->enableObjectReachableCheck) {
             return true;
         }
+        // User card always reachable
```
batopa, line 377: To avoid repeating the query every time the method is called maybe we could move the Card load in `ApiAuthComponent::findUser()` but I think in `$user` var you should already have data about `ObjectUser`.

batopa (changes requested):

batopa (approved):

PR #1456, opened 2018-04-18
Routing rules change

stefanorosanelli:
This PR changes routing rules this way:

 * route load order was changed in order to allow other plugins to use resources (not objects)  properly in JSON API 
 * rule names are now set in models to have better control
 *  fix: `controller` names in rules are now `camelized`

### DIFF
--- a/plugins/BEdita/API/src/Controller/ResourcesController.php
+++ b/plugins/BEdita/API/src/Controller/ResourcesController.php
@@ -24,6 +24,7 @@ use BEdita\Core\Model\Action\SaveEntityAction;
 use BEdita\Core\Model\Action\SetAssociatedAction;
 use BEdita\Core\Utility\JsonApiSerializable;
 use Cake\Core\InstanceConfigTrait;
+use Cake\Datasource\EntityInterface;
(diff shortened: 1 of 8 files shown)


plugins/BEdita/API/src/Controller/ResourcesController.php
```diff
@@ -200,16 +201,22 @@ public function index()
...
+    protected function resourceUrl(EntityInterface $entity, $primaryKey)
     {
+        $prefix = 'api:resources';
+        if ($entity instanceof JsonApiSerializable) {
+            $prefix = $entity->routeNamePrefix();
+        }
+
         return Router::url(
             [
-                '_name' => $this->routeNamePrefix . ':resource',
```
fquffio, line 210: Can `$routeNamePrefix` class member (see line 74) be removed now?

batopa (approved):

plugins/BEdita/API/src/Controller/ResourcesController.php
stefanorosanelli: Of course! Done

PR #1458, opened 2018-04-24
Don't list deleted objects in `children` and `parents

stefanorosanelli:
This PR fixes a bug in `children`, `parents` and `parent` relations: deleted objects MUST not be listed.

Problem: `ListEntitiesAction` was used instead of `ListObjectsAction` as list association.

Deleted `children` and `parents` test cases have been added.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
@@ -13,7 +13,9 @@
 
 namespace BEdita\Core\Model\Action;
 
+use BEdita\Core\Model\Table\ObjectsTable;
 use BEdita\Core\ORM\Association\RelatedTo;
+use BEdita\Core\ORM\Inheritance\Table;
 use Cake\ORM\Association;
 use Cake\ORM\TableRegistry;
 
@@ -44,7 +46,8 @@ class ListRelatedObjectsAction extends ListAssociatedAction
                 $objectType = current($objectTypes);
             }
             $this->ListAction = new ListObjectsAction(compact('table', 'objectType'));
-        } elseif ($this->Association->getName() === 'Parents') {
+        } elseif ($this->Association->getTarget() instanceof ObjectsTable
+                || $this->Association->getTarget() instanceof Table) {
             $table = $this->Association->getTarget();
             $this->ListAction = new ListObjectsAction(compact('table'));
         }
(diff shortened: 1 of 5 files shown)


plugins/BEdita/Core/tests/TestCase/Model/Action/ListRelatedFoldersActionTest.php
```diff
@@ -48,6 +48,7 @@ class ListRelatedFoldersActionTest extends TestCase
      * @return void
      *
      * @covers ::execute()
+     * @covers \BEdita\Core\Model\Action\ListRelatedObjectsAction::initialize()
```
fquffio, line 51: Uhm… we're not actually covering that method, testing its functionality or whatever, are we? We're just "using" it. There's the [`@uses`](http://phpunit.readthedocs.io/en/7.1/annotations.html#uses) tag for these cases, but we're not using it anywhere because it would be too verbose.

plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
```diff
@@ -44,6 +44,9 @@ protected function initialize(array $config)
                 $objectType = current($objectTypes);
             }
             $this->ListAction = new ListObjectsAction(compact('table', 'objectType'));
+        } elseif ($this->Association->getName() === 'Parents') {
```
fquffio: This is totally fine. I'm just wondering if we can perform some kind of a more flexible check, like `$this->Association->getTarget() instanceof ObjectsTable` or something like that… @batopa ?

(Same would apply to `ListRelatedFoldersAction::initialize()`)

fquffio (approved):

plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
batopa: A more flexible approach should be better for sure but  for `Parents` association `$this->Association->getTarget()` would be `FoldersTable` while for `Children` association would be `ObjectsTable`. 

Maybe we could check that `$this->Association->getTarget() instanceof ObjectsTable || $this->Association->getTarget() instanceof \BEdita\Core\ORM\Inheritance\Table `. 
In this way we could keep this check only in `ListRelatedObjectsAction` and remove the check from `ListRelatedFoldersAction`.

stefanorosanelli: Done with check suggested by @batopa

fquffio (approved):

PR #1459, opened 2018-04-24
Fix `?include=children,parent` query string

stefanorosanelli:
This PR fixes errors on `?include=children`, `?include=parent` query string on `/folders` and `?include=parents` on other object endpoints.

### DIFF
--- a/plugins/BEdita/Core/tests/TestCase/Model/Entity/ObjectEntityTest.php
+++ b/plugins/BEdita/Core/tests/TestCase/Model/Entity/ObjectEntityTest.php
@@ -507,6 +507,26 @@ class ObjectEntityTest extends TestCase
      */
     public function testGetRelationshipsIncluded()
     {
+        $entity = TableRegistry::get('Documents')->get(2, ['contain' => ['Test']]);
+        $entity = $entity->jsonApiSerialize();
+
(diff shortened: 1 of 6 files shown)


plugins/BEdita/Core/tests/TestCase/Model/Entity/ObjectEntityTest.php
```diff
@@ -506,14 +507,14 @@ public function testGetRelationshipsDeleted()
      */
     public function testGetRelationshipsIncluded()
     {
-        $entity = TableRegistry::get('Documents')->get(2, ['contain' => ['Test']]);
+        $entity = TableRegistry::get('Folders')->get(12, ['contain' => ['Parents']]);
```
fquffio: [very minor] Instead of editing this test case, I'd prefer adding a new one.

fquffio (approved):

plugins/BEdita/Core/tests/TestCase/Model/Entity/ObjectEntityTest.php
stefanorosanelli: ok, done!

batopa (approved):

PR #1462, opened 2018-04-26
Fix SQL injection vulnerability with relations parameters

fquffio:
This PR fixes a (potentially serious) vulnerability. When saving a relation with parameters, since the encoded JSON was not being properly escaped, it was possible to inject custom SQL by simply adding an apostrophe (`'`) to one relation parameter.

For instance, when linking a Document and an Image together, it would have been possible to give a maliciously-crafted "label" to expose arbitrary data from any table.

With this PR the logic has been heavily refactored to make use of CakePHP's ORM, so that we're sure that any user-provided data is escaped as it should.

### DIFF
--- a/bedita-app/models/object_relation.php
+++ b/bedita-app/models/object_relation.php
@@ -28,6 +28,26 @@ App::import('Sanitize');
 class ObjectRelation extends BEAppModel
 {
 
+    /**
+     * Check if a relation exists.
+     *
+     * @return bool
+     */
+    public function exists() {
+        $name = $this->name;
+        if (!isset($this->data[$name]['id']) || !isset($this->data[$name]['object_id']) || !isset($this->data[$name]['switch'])) {
+            return false;
+        }
+
+        $conditions = array(
+            $this->alias . '.id' => $this->data[$name]['id'],
+            $this->alias . '.object_id' => $this->data[$name]['object_id'],
+            $this->alias . '.switch' => $this->data[$name]['switch'],
+        );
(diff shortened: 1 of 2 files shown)


didoda (approved):

bedita-app/models/object_relation.php
```diff
@@ -28,6 +28,26 @@
 class ObjectRelation extends BEAppModel
 {
 
+    /**
+     * Check if a relation exists.
+     *
+     * @return bool
+     */
+    public function exists() {
```
batopa, line 36: This method doesn't seem used. Is it correct?

bedita-app/models/object_relation.php
```diff
@@ -28,6 +28,26 @@
 class ObjectRelation extends BEAppModel
 {
 
+    /**
+     * Check if a relation exists.
```
batopa, line 32: I would add that the check is performed on `Model::$data`.

batopa (commented):
It seems correct! Just a few notes to clarify

bedita-app/models/object_relation.php
fquffio: Actually this method overrides `Model::exists()` (the default CakePHP's method). It is used internally during the `Model::save()` operation to check if a new record should be created, or an existing should be updated.

It is required to override this method because by default it would only check `id`, but since we're using a composite primary key we must perform a different check.

batopa (approved):

fquffio:
Closing this PR as it turned out to have bad effects. #1608 provides a rougher, although less safe, solution that does not involve a heavy refactor of the save logic.

PR #1463, opened 2018-04-26
`parent`, `ancestor` and `roots` filter

### LINKED ISSUE #1460: `parent` and `ancestor` filters
General filter on `GET` objects list like:

* `filter[parent]={id|uname}` objects having a certain `folder` as **parent**
   * might be handy something like `/folders?filter[parent]=null` to get root folders
* `filter[ancestor]={id|uname}` objects having a certain `folder` as **ancestor** => apply to `/folders` to get a complete tree

stefanorosanelli:
This PR fixes #1460 

Three new filters have been introduced:

* `filter[parent]={id|uname}` to get children of a certain `folder`
* `filter[ancestor]={id|uname}` to get all descendants of a certain folder
* `/folders?filter[roots]` to get root folders

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/FoldersTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/FoldersTable.php
@@ -161,17 +161,16 @@ class FoldersTable extends ObjectsTable
      */
     protected function findRoots(Query $query)
     {
-        $subquery = TableRegistry::get('Trees')->find('list')->where(['parent_id IS NULL']);
-
         $query->join([
             'table' => 'trees',
             'alias' => 'Trees',
             'type' => 'INNER',
             'conditions' => 'Trees.object_id = ' . $this->aliasField('id'),
         ])
-        ->where(function (QueryExpression $exp) use ($subquery) {
-            return $exp->in('Trees.id', $subquery);
-        });
+        ->where(function (QueryExpression $exp) {
+            return $exp->isNull('Trees.parent_id');
+        })
+        ->order('Trees.tree_left');
 
         return $query;
     }
--- a/plugins/BEdita/Core/src/Model/Table/FoldersTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/FoldersTable.php
@@ -62,6 +62,11 @@ class FoldersTable extends ObjectsTable
                 'Trees.tree_left' => 'asc',
             ],
         ]);
+
+        $this->hasMany('TreeParentNodes', [
+            'className' => 'Trees',
+            'foreignKey' => 'parent_id',
+        ]);
     }
 
     /**
@@ -161,17 +166,12 @@ class FoldersTable extends ObjectsTable
      */
     protected function findRoots(Query $query)
     {
-        $query->join([
-            'table' => 'trees',
-            'alias' => 'Trees',
-            'type' => 'INNER',
-            'conditions' => 'Trees.object_id = ' . $this->aliasField('id'),
-        ])
-        ->where(function (QueryExpression $exp) {
-            return $exp->isNull('Trees.parent_id');
-        })
-        ->order('Trees.tree_left');
-
-        return $query;
+        return $query
+            ->innerJoinWith('TreeNodes', function (Query $query) {
+                return $query->where(function (QueryExpression $exp) {
+                    return $exp->isNull($this->TreeNodes->aliasField('parent_id'));
+                });
+            })
+            ->order('TreeNodes.tree_left');
     }
 }
--- a/plugins/BEdita/Core/src/Model/Table/FoldersTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/FoldersTable.php
@@ -14,8 +14,10 @@
 namespace BEdita\Core\Model\Table;
 
 use BEdita\Core\Model\Entity\Folder;
+use Cake\Database\Expression\QueryExpression;
 use Cake\Datasource\EntityInterface;
 use Cake\Event\Event;
+use Cake\ORM\Query;
 use Cake\ORM\RulesChecker;
 use Cake\ORM\Rule\ValidCount;
 use Cake\ORM\TableRegistry;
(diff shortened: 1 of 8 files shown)


plugins/BEdita/Core/src/Model/Table/FoldersTable.php
```diff
@@ -150,4 +152,27 @@ public function afterSave(Event $event, EntityInterface $entity)
...
+    {
+        $subquery = TableRegistry::get('Trees')->find('list')->where(['parent_id IS NULL']);
+
+        $query->join([
+            'table' => 'trees',
+            'alias' => 'Trees',
+            'type' => 'INNER',
+            'conditions' => 'Trees.object_id = ' . $this->aliasField('id'),
+        ])
+        ->where(function (QueryExpression $exp) use ($subquery) {
```
batopa: I think it can be simplified without using subquery as

```php
$query->join([
            'table' => 'trees',
            'alias' => 'Trees',
            'type' => 'INNER',
            'conditions' => 'Trees.object_id = ' . $this->aliasField('id'),
        ])
            return $exp->isNull('Trees.parent_id');
        });
```

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -275,4 +277,62 @@ public function getId($uname)
...
+            ])
+            ->find('list');
+
+        $query->join([
+            'table' => 'trees',
+            'alias' => 'Trees',
+            'type' => 'INNER',
+            'conditions' => 'Trees.object_id = ' . $this->aliasField('id'),
+        ])
+        ->where(function (QueryExpression $exp) use ($subquery) {
```
batopa: We could avoid the subquery using

```php
if ($direct) {
    return $query->where(['Trees.parent_id' => $id])
}

$node = TableRegistry::get('Trees')->find()
    ->where(['Trees.object_id' => $this->getId($id)])
    ->first();

return $query->where([
    'Trees.tree_left >' => $node->tree_left,
     'Trees.right_left <' => $node->tree_left,
]);
```

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -275,4 +277,61 @@ public function getId($uname)
...
+     * @return \Cake\ORM\Query
+     */
+    protected function treeDescendants(Query $query, $id, $direct = true)
+    {
+        $query = $query->join([
+            'table' => 'trees',
+            'alias' => 'Trees',
+            'type' => 'INNER',
+            'conditions' => 'Trees.object_id = ' . $this->aliasField('id'),
+        ]);
```
batopa: [super minor]: The `order()` could be moved here since is always the same

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
```diff
@@ -150,4 +152,26 @@ public function afterSave(Event $event, EntityInterface $entity)
...
+        $query->join([
+            'table' => 'trees',
+            'alias' => 'Trees',
+            'type' => 'INNER',
+            'conditions' => 'Trees.object_id = ' . $this->aliasField('id'),
+        ])
+        ->where(function (QueryExpression $exp) {
+            return $exp->isNull('Trees.parent_id');
+        })
+        ->order('Trees.tree_left');
```
fquffio: Instead of building the join this way, I'd by far prefer to use associations.

We might add two associations:
```php
# ObjectsTable
$this->hasMany('TreeNodes', [
    'className' => 'Trees',
    'foreignKey' => 'object_id',
]);

# FoldersTable
$this->hasMany('TreeParentNodes', [
    'className' => 'Trees',
    'foreignKey' => 'parent_id',
]);
```

Doing so, this finder could be rewritten as follows:
```php
return $query
    ->innerJoinWith('TreeNodes', function (Query $query) {
        return $query->where(function (QueryExpression $exp) {
            return $exp->isNull($this->TreeNodes->aliasField('parent_id'));
        });
    })
    ->order('TreeNodes.tree_left');
```

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -275,4 +277,61 @@ public function getId($uname)
...
+    /**
+     * Finder for objects having a certain `ancestor` on the tree.
+     *
+     * @param Query $query  Query object instance.
+     * @param array $options Id or unique name of ancestor
+     * @return \Cake\ORM\Query
+     */
+    protected function findAncestor(Query $query, array $options)
+    {
+        return $this->treeDescendants($query, Hash::get($options, '0'), false);
```
fquffio: If we add the associations as of the other comment, this finder could be rewritten as follows:
```php
$parentId = Hash::get($options, '0'); // I'm not a big fan of `Hash::get($arr, 0)`. I'd prefer an explicit check in this case.
$parentNode = $this->TreeNodes->find()
    ->where([
        $this->TreeNodes->aliasField('object_id') => $parentId,
    ])
    ->firstOrFail();

return $query
    ->innerJoinWith('TreeNodes', function (Query $query) use ($parentNode) {
        return $query->where(function (QueryExpression $exp) use ($parentNode) {
            return $exp
                ->gt($this->TreeNodes->aliasField('tree_left'), $parentNode->get('tree_left'))
                ->lt($this->TreeNodes->aliasField('tree_right'), $parentNode->get('tree_right'));
        });
    })
    ->order($this->TreeNodes->aliasField('tree_left')); // <-- Does this make sense?
```

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -275,4 +277,61 @@ public function getId($uname)
...
+    /**
+     * Finder for objects having a certain `parent` on the tree.
+     *
+     * @param Query $query Query object instance.
+     * @param array $options Id or unique name of ancestor
+     * @return \Cake\ORM\Query
+     */
+    protected function findParent(Query $query, array $options)
+    {
+        return $this->treeDescendants($query, Hash::get($options, '0'), true);
```
fquffio: If we add the associations as of the other comment, this finder could be rewritten as follows:
```php
$parentId = Hash::get($options, '0');

return $query
    ->innerJoinWith('TreeNodes', function (Query $query) use ($parentId) {
        return $query->where([
            $this->TreeNodes->aliasField('parent_id') => $parentId,
        ]);
    })
    ->order($this->TreeNodes->aliasField('tree_left'));
```

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
stefanorosanelli: no more necessary, after new refactor

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
stefanorosanelli: done

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
stefanorosanelli: done

fquffio (approved):
👍 Travis build is failing because we should add `tree_nodes` and `tree_parent_nodes` to `ObjectEntity::$_hidden` and `Folder::$_hidden` respectively.

stefanorosanelli:
Noticed the failure, shoudl be fixed now

Issue #1461, opened 2018-04-26
Path info in `meta`


stefanorosanelli:
For every `object`, or only on `folders` to start, display in `meta` tree path information.

A single path will be available for `folders`, a list of `paths` for other object types.

PR #1464, opened 2018-04-27
#1461 path in meta for folders

batopa:
This PR add in `path` in `meta` for `folders` object  as described in #1461

### DIFF
--- a/plugins/BEdita/Core/tests/TestCase/Model/Entity/FolderTest.php
+++ b/plugins/BEdita/Core/tests/TestCase/Model/Entity/FolderTest.php
@@ -169,4 +169,40 @@ class FolderTest extends TestCase
         $folder = $folder->jsonApiSerialize();
         static::assertArrayHasKey('parent', $folder['relationships']);
     }
+
+    /**
+     * Data provider for `testGetPath()`
+     *
+     * @return array
+     */
+    public function getPathProvider()
+    {
+        return [
+            'root' => [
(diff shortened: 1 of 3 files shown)


plugins/BEdita/Core/tests/TestCase/Model/Entity/FolderTest.php
```diff
@@ -169,4 +169,40 @@ public function testListAssociations()
...
+     *
+     * @return void
+     *
+     * @dataProvider getPathProvider
+     * @covers ::_getPath()
+     */
+    public function testGetPath($expected, $id)
+    {
+        $folder = new Folder();
+        static::assertNull($folder->path);
```
fquffio: [minor] I'd split this test into its own test case, and use `$this->Folders->newEntity()` instead.

fquffio (approved):

plugins/BEdita/Core/tests/TestCase/Model/Entity/FolderTest.php
batopa: yeah, I think it's better :+1:

batopa: done

fquffio (approved):

didoda (approved):

stefanorosanelli (approved):

PR #1467, opened 2018-05-02
Allow resource subtypes in relationships

stefanorosanelli:
This PR allows use of a subtype in a relationship when its parent abstract type is allowed.

`JsonApiComponent::checkAllowedType()` method was added in order to check on allowed parent abstract types -> it can be **heavily** improved....

### DIFF
--- a/plugins/BEdita/API/src/Controller/Component/JsonApiComponent.php
+++ b/plugins/BEdita/API/src/Controller/Component/JsonApiComponent.php
@@ -18,6 +18,7 @@ use Cake\Controller\Component;
 use Cake\Network\Exception\BadRequestException;
 use Cake\Network\Exception\ConflictException;
 use Cake\Network\Exception\ForbiddenException;
+use Cake\ORM\TableRegistry;
 use Cake\Routing\Router;
 use Cake\Utility\Hash;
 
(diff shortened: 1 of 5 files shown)


plugins/BEdita/API/src/Controller/Component/JsonApiComponent.php
```diff
@@ -220,11 +221,38 @@ protected function allowedResourceTypes($types, array $data = null)
...
+    protected function checkAllowedType($type, array $allowed)
+    {
+        if (in_array($type, $allowed)) {
+            return true;
+        }
+
+        // See if `$type` is a sub-type of an `allowed` abstract type
+        $objectTypesTable = TableRegistry::get('ObjectTypes');
+        $abstract = $objectTypesTable->find('list', ['valueField' => 'id'])->where(['is_abstract' => true, 'name IN' => $allowed]);
+        foreach ($abstract as $id) {
```
fquffio, line 246: This is way too much time- and resource-consuming. An improvement could be:

```php
if (in_array($type, $allowed)) {
    // Explicitly allowed.
    return true;
}

$type = TableRegistry::get('ObjectTypes')->get($type);

return TableRegistry::get('ObjectTypes')->exists([
    // Check if there is an allowed type that is in the path of the type being checked.
    'tree_left <' => $type->tree_left,
    'tree_right >' => $type->tree_right,
    'name IN' => $allowed,
]);
```

However, I'm not sure that the check should happen here in the first place.

fquffio (commented):
These changes only solve the "API-part" of the problem. To achieve full-featured relationships inheritance we need to play with **ObjectTypes**, **Relations** and the behaviour that dynamically creates associations.

stefanorosanelli:
No more needed after #1469 
Closing

PR #1468, opened 2018-05-02
Included deleted objects #1466

### LINKED ISSUE #1466: Included deleted objects
Deleted objects MUST not appear as `included` with an `?include={relation}` query string.
Code snippet below is a test case on `ObjectsControllerTest` where after a delete:

* `GET /documents/2/test` is ok
* `GET /documents/2?include=test` assert fails, a deleted object is displayed

```php
    public function testIncludedDeleted()
    {
        $this->configRequestHeaders('DELETE', $this->getUserAuthHeader());
        $this->delete('/documents/3');
        $this->assertResponseCode(204);

        $this->configRequestHeaders();
        $this->get('/documents/2/test');
        static::assertEquals(1, count($result['data']));

        $this->configRequestHeaders();
        $this->get('/documents/2?include=test');
        static::assertEquals(1, count($result['included']));
    }
```

fquffio:
This PR fixes #1466.

Extra conditions are applied when associations are built in `RelationsBehavior` to exclude deleted objects from results. Some actions needed small fixes to filter conditions when building queries or creating junction entities.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/UpdateAssociatedAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/UpdateAssociatedAction.php
@@ -95,9 +95,10 @@ abstract class UpdateAssociatedAction extends BaseAction
         $prefix = sprintf('%s.', $junction->getAlias());
         $extraFields = [];
         foreach ($conditions as $field => $value) {
-            if (substr($field, 0, strlen($prefix)) === $prefix) {
-                $field = substr($field, strlen($prefix));
+            if (substr($field, 0, strlen($prefix)) !== $prefix) {
+                continue;
             }
+            $field = substr($field, strlen($prefix));
 
             $extraFields[$field] = $value;
         }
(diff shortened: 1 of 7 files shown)


stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Action/UpdateAssociatedAction.php
```diff
@@ -95,9 +95,10 @@ protected function prepareRelatedEntities($relatedEntities, EntityInterface $sou
         $prefix = sprintf('%s.', $junction->getAlias());
         $extraFields = [];
         foreach ($conditions as $field => $value) {
-            if (substr($field, 0, strlen($prefix)) === $prefix) {
-                $field = substr($field, strlen($prefix));
+            if (substr($field, 0, strlen($prefix)) !== $prefix) {
+                continue;
```
batopa, line 99: Why  `$extraFields` is populated differently from before.

fquffio: It is populated in a different way because before we used to have only conditions on the junction table: `$conditions === ['ObjectRelations.relation_id' => 42]`).
Now we also have conditions on the target table: `$conditions === ['ObjectRelations.relation_id' => 42, 'Objects.deleted' => false]`.

While `relation_id = 42` should be added as an "extra field" to the junction entity, `deleted = false` should not.

batopa: Clear!  Thank you!

batopa (approved):

PR #1469, opened 2018-05-02
Relations inheritance #1349

### LINKED ISSUE #1349: Relations inheritance
When a new object type is created its parent types relations should be inherited, just like properties.

This follows #908 and #1345 

Are there problematic use cases in this scenario? 
Can we avoid them?

fquffio:
This PR resolves #1349, and supersedes #1467.

When a relation is valid for an abstract object type (e.g.: **Objects** or **Media**), it becomes available also for all its sub-types (e.g.: for **Objects**, all object types; for **Media** only **Images**, **Files**, …).


didoda (approved):

stefanorosanelli (approved):
Not completely clear what happens in `ObjectTypesTable::findByRelation()` but we trust you :smiley:

PR #1473, opened 2018-05-08
Prevent login from blocked, deleted and off users

stefanorosanelli:
This PR fixes a problem in `POST /auth` login action.

Users having `"blocked": true`, `"status": "off"` or deleted users must not be able to login.

Is `status` draft acceptable? Leaved because is used in `UuidAuthenticateTest` but I'm not sure if it's correct.

In order to achieve this a `login` finder was introduced in `UsersTable` and in `LoginController`.

Same login conditions are applied to `externalAuth` finder.

### DIFF
--- a/plugins/BEdita/API/src/Controller/LoginController.php
+++ b/plugins/BEdita/API/src/Controller/LoginController.php
@@ -65,8 +65,8 @@ class LoginController extends AppController
                             'Weak' => ['hashType' => 'md5'],
                         ],
                     ],
-                    'finder' => 'userLogin',
-                ],
+                    'finder' => 'login',
+                 ],
                 'BEdita/API.Jwt' => [
                     'queryDatasource' => true,
                 ],
--- a/plugins/BEdita/API/src/Controller/LoginController.php
+++ b/plugins/BEdita/API/src/Controller/LoginController.php
@@ -65,6 +65,7 @@ class LoginController extends AppController
                             'Weak' => ['hashType' => 'md5'],
                         ],
                     ],
+                    'finder' => 'login',
                 ],
                 'BEdita/API.Jwt' => [
                     'queryDatasource' => true,
--- a/plugins/BEdita/API/src/Controller/LoginController.php
+++ b/plugins/BEdita/API/src/Controller/LoginController.php
@@ -65,7 +65,7 @@ class LoginController extends AppController
                             'Weak' => ['hashType' => 'md5'],
                         ],
                     ],
-                    'finder' => 'login',
+                    'finder' => 'userLogin',
                 ],
                 'BEdita/API.Jwt' => [
                     'queryDatasource' => true,
--- a/plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
@@ -26,6 +26,22 @@ use Cake\Utility\Hash;
 class LoginControllerTest extends IntegrationTestCase
 {
     /**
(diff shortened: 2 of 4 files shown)


plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -247,6 +249,39 @@ protected function findMine(Query $query)
...
+
+    /**
+     * Finder for users login with `username`
+     * Valid attributes for `blocked`, `deleted` and `status` are checked
+     *
+     * @param \Cake\ORM\Query $query Query object instance.
+     * @return \Cake\ORM\Query
+     * @throws \Cake\Network\Exception\BadRequestException if `username` is missing
+     */
+    protected function findLogin(Query $query, array $options)
```
fquffio: I'd change this finder. It shouldn't be necessary to add the condition on username on our side: the auth component should already handle that, and in a more flexible way than this (the "username" field can be configured).

Therefore, I'd make this finder only filter _all_ users that are able to login:

```php
function findLogin(Query $query)
{
    return $query
            return $exp
                ->eq($this->aliasField('deleted'), false)
                ->eq($this->aliasField('blocked'), false)
                ->in($this->aliasField('status'), ['on', 'draft']);
        });
}
```

Doing so, the `loginConditions()` method would become useless, and the `findExternalAuth()` finder could simply apply `$query->find('login')` to the query it manipulates.

plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
```diff
@@ -415,4 +416,119 @@ public function testUpdateIgnore(array $meta)
...
     }
+
+    /**
+     * Not successful login expected result
+     *
+     * @return array
+     */
+    protected function notSuccessfulExpectedResult()
+    {
+        return [
```
fquffio: (minor) This may become a class constant.

plugins/BEdita/Core/src/Model/Table/UsersTable.php
stefanorosanelli: good point, I'm changing it

stefanorosanelli: Ok, changed but `username` still necessary.
Ended up with two finders.

plugins/BEdita/API/src/Controller/LoginController.php
```diff
@@ -65,7 +65,7 @@ public function initialize()
                             'Weak' => ['hashType' => 'md5'],
                         ],
                     ],
-                    'finder' => 'login',
+                    'finder' => 'userLogin',
```
fquffio: Uhm… `login` finder should be enough here. `userLogin` finder should be redundant.

stefanorosanelli: And what about changing scope config like:
```php
    'scope' => [
        'blocked' => false,
        'deleted' => false,
        'status IN' => ['on', 'draft']
    ],
```
without custom finder?

fquffio: Should be fine, but looking forward to migrating to Auth plugin, `scope` will not be available any more, so if using a finder doesn't cause too much hassle, that should be the preferred way to go.

stefanorosanelli: ok, let's use a finder then

stefanorosanelli: done

fquffio (approved):

didoda (approved):

PR #1475, opened 2018-05-10
fix: 500 error getting path of orphan folder

batopa:
This PR replaces the previous **404** error with a **500** error getting `Folder::$path` virtual property since an orphan folder (folder without entry on `trees` table) should not exists.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Entity/Folder.php
+++ b/plugins/BEdita/Core/src/Model/Entity/Folder.php
@@ -110,6 +110,7 @@ class Folder extends ObjectEntity
      * Getter for `path` virtual property
      *
      * @return string|null
+     * @throws \RuntimeException If Folder is not found on tree.
      */
     protected function _getPath()
     {
@@ -120,7 +121,11 @@ class Folder extends ObjectEntity
         $trees = TableRegistry::get('Trees');
         $node = $trees->find()
             ->where(['object_id' => $this->id])
-            ->firstOrFail();
+            ->first();
+
+        if (!$node) {
+            throw new \RuntimeException(__d('bedita', 'Folder "{0}" is not on the tree.', $this->id));
+        }
 
         $path = $trees->find('list', [
                 'keyField' => 'id',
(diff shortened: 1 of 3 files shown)


stefanorosanelli (approved):

plugins/BEdita/Core/tests/TestCase/Model/Entity/FolderTest.php
```diff
@@ -217,4 +217,22 @@ public function testGetPathNull()
...
+     * @expectedException \RuntimeException
+     * @expectedExceptionMessage Folder "12" is not on the tree.
+     */
+    public function testGetPathOrphanFolder()
+    {
+        $treesTable = TableRegistry::get('Trees');
+        $entity = $treesTable->find()->where(['object_id' => 12])->firstOrFail();
+        $treesTable->delete($entity);
+
+        $folder = $this->Folders->get(12);
```
fquffio: [minor] Variable `$folder` is unused.

plugins/BEdita/Core/src/Model/Entity/Folder.php
```diff
@@ -120,7 +121,11 @@ protected function _getPath()
         $trees = TableRegistry::get('Trees');
         $node = $trees->find()
             ->where(['object_id' => $this->id])
-            ->firstOrFail();
+            ->first();
+
+        if (!$node) {
+            throw new \RuntimeException(__d('bedita', 'Folder "{0}" is not on the tree.', $this->id));
```
fquffio: Another option, that also keeps track of the "full stack" of the error, would be:
```php
try {
    $node = $trees->find()
        ->where(['object_id' => $this->id])
        ->firstOrFail();
} catch (RecordNotFoundException $previous) {
    throw new \RuntimeException($message, 0, $previous);
}
```

fquffio (approved):

plugins/BEdita/Core/src/Model/Entity/Folder.php
batopa: ok, but what advantage do we have chaining the exception? Given a look at stack trace I don't see any difference :thinking:

PR #1477, opened 2018-05-10
Fix multiple issues with Folders' relationships #1471

### LINKED ISSUE #1471: Set children fail + set parent allowed types
Two problems on folder relationships actions:

1. setting `children` via POST fails 
1. set parent via PATCH is allowed for non-folder objects

See this compare as reference on `stefanorosanelli:fix/v4/children-parent-fail`
https://github.com/bedita/bedita/compare/4-cactus...stefanorosanelli:fix/v4/children-parent-fail

1. `patch conflict` test passes with 200 instead of failing with 409
1. `testSetParent` is OK
1. `testSetChildren` fails

fquffio:
This PR fixes #1471 by adding the suggested tests and fixing some strange behaviours when linking parents and children together.

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Controller/FoldersControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/FoldersControllerTest.php
@@ -188,7 +188,7 @@ class FoldersControllerTest extends IntegrationTestCase
 
         $this->assertResponseCode(200);
         $this->assertContentType('application/vnd.api+json');
-        $this->assertEquals($expected, $result);
+        static::assertEquals($expected, $result);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


plugins/BEdita/API/tests/TestCase/Controller/FoldersControllerTest.php
```diff
@@ -518,20 +543,91 @@ public function setRelationshipsAllowedMethodsProvider()
      *
      * @param int $expected The expected HTTP status code.
      * @param string $method The http method.
-     * @param array|null $data Payload to use for the request.
+     * @param array $data Payload to use for the request.
      * @return void
+     * @throws \Exception
```
batopa: [minor] Not really :smile_cat:

stefanorosanelli (approved):

plugins/BEdita/API/tests/TestCase/Controller/FoldersControllerTest.php
fquffio: Yes, you're right… 😞 It's PHPStorm that keeps on complaining about the lack of this annotation because `IntegrationTestCase::_sendRequest()` method's docblock includes `@throws \Exception`. 😡

batopa (approved):

PR #1480, opened 2018-05-15
#1454 Move folders action

### LINKED ISSUE #1454: Move folders action
### Expected behavior

After #1431 : PATCH on folders should update the current row on trees table but it first removes the current row and then insert a new row. Deleting the row all children are lost!!

Moving folders to different parent MUST not cause children deletion.

batopa:
This PR fixes #1454 

### Folder move to other position

```http
PATCH /folders/:id/relationships/parent

{
    "data": {
        "type": "folders",
        "id": "12"
    }
}
```

### Folder become root

```http
PATCH /folders/:id/relationships/parent

{
    "data": null
}
```

All descendants are maintained on the move action.

#### Side note
Some minor bugs are been fixed:
* handle JSON with `"data": null`
* avoid error in `TableRegistry::get()` using sequentially `ListAssociatedAction`. That action add to registry a termporary alias for inverse association that was never removed.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
@@ -68,8 +68,12 @@ abstract class UpdateRelatedObjectsAction extends UpdateAssociatedAction
     /**
      * Get the right entity for the action.
      *
-     * @param EntityInterface $entity The starting entity.
-     * @return void
+     * For `Folder` entity with `Parents` association changes the point of view
+     * using `Tree` entity with `ParentObjects` association assuring to
+     * always use a "to one" relation.
+     *
+     * @param \Cake\Datasource\EntityInterface $entity The starting entity.
+     * @return \Cake\Datasource\EntityInterface
      */
     protected function getEntity(EntityInterface $entity)
     {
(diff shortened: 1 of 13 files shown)


fquffio (approved):

plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
```diff
@@ -53,4 +54,43 @@ function ($key) use ($conditionsPrefix) {
...
+
+        $table = $this->Association->junction();
+        $entity = $table->find()
+            ->where([$table->association('Objects')->getForeignKey() => $entity->id])
+            ->firstOrFail();
+
+        $this->Association = $table->association('ParentObjects');
+        $this->setConfig('association', $this->Association);
+
+        return $entity;
```
didoda, line 94: Scrutinizer suggests:
```
The expression return $entity also could return the type array which is incompatible with the documented return type Cake\Datasource\EntityInterface.
```
False positive?

plugins/BEdita/Core/tests/TestCase/Model/Action/RemoveRelatedObjectsActionTest.php
```diff
@@ -0,0 +1,155 @@
...
+                12,
+                null,
+            ],
+        ];
+    }
+
+    /**
+     * Test invocation of command.
+     *
+     * @param bool|\Exception Expected result.
```
didoda: typo: missing var.
`@param bool|\Exception $expected Expected result.`

stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
batopa: yes it seems a false positive

didoda (approved):

Issue #1453, opened 2018-04-16
Delete operations on folders


stefanorosanelli:
### Expected behavior

On a `delete` action, aka `move to trash`:

* all children `folders` are trashed as well
* other children leaf objects are not trashed, they may become `orphans` if not insider other folders
* [in another future issue] with a dedicated option/parameter also children leaf objects can be trashed

On a `permanent remove` action, aka `empty trash`:

 * all children `folders` are permanently removed as well

On a `restore` action:

 * all children `folders` are restored as well
 * action fails if restored `folder` parent is in the trashcan -> generally a parent folder on create/move actions MUST NOT be in the trashcan

[see #1431 for a discussion on this topic]

stefanorosanelli:
Closing
Permanent remove part moved to #1495

PR #1482, opened 2018-05-22
#1453 Soft delete and restore of folders

batopa:
This PR implements soft delete and restore of folders as seen in #1453 

1. deleting (soft) a folder all sub-folders are deleted (soft). Other children leaf objects are not trashed.
2. restoring a folder all sub-folders are restored. The restore is only possible when no ancestor is in a deleted status.

**Bonus track**
`TrashController` now try to use the Table of the object in view/restore/delete actions retrieving it from `id` in url path. This for correctly use the relative Table callbacks. Before, it did always use `ObjectsTable`.


stefanorosanelli (approved):

plugins/BEdita/API/src/Controller/TrashController.php
```diff
@@ -32,6 +33,30 @@ class TrashController extends AppController
      */
     public $modelClass = 'Objects';
```
fquffio, line 34: Is `$this->Objects` still used? If not, we might remove `$modelClass` and `@property` annotation in class' docblock.

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
```diff
@@ -85,6 +85,15 @@ public function buildRules(RulesChecker $rules)
...
         );
 
+        $rules->add(
+            [$this, 'isFolderRestorable'],
+            'isFolderRestorable',
+            [
+                'errorField' => 'deleted',
+                'message' => __d('bedita', 'Folder can be restored only if its ancestors are not deleted.'),
+            ]
+        );
```
fquffio, line 95: Can this be set as a validation rule?

```php
$validator
    ->add('deleted', 'restorable', [
        'rule' => function ($value, $context) {
            // check that folder's parents are not deleted
        },
        'when' => function ($context) {
            return $context['data']['deleted'] === true;
        },
    ]);
```

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
```diff
@@ -106,6 +115,40 @@ public function hasAtMostOneParent(Folder $entity)
...
+        $deletedParents = $this->find()
+            ->innerJoinWith('TreeNodes', function (Query $query) use ($node) {
+                return $query->where(function (QueryExpression $exp) use ($node) {
+                    return $exp
+                        ->lt($this->TreeNodes->aliasField('tree_left'), $node->get('tree_left'))
+                        ->gt($this->TreeNodes->aliasField('tree_right'), $node->get('tree_right'));
+                });
+            })
+            ->where([$this->aliasField('deleted') => true])
+            ->count();
```
fquffio, line 147: Can this be rewritten using `Table::exists()`? It shouldn't be a problem on trees with a limited depth, though.

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -459,4 +460,26 @@ protected function findByRelation(Query $query, array $options = [])
...
+     *
+     * @param \Cake\ORM\Query $query Query object.
+     * @param array $options Additional options. The `id` key is required.
+     * @return \Cake\ORM\Query
+     * @throws \LogicException When missing required parameters.
+     */
+    protected function findObjectId(Query $query, array $options = [])
+    {
+        if (empty($options['id'])) {
+            throw new \LogicException(__d('bedita', 'Missing required parameter "{0}"', 'id'));
```
fquffio: Maybe `BadFilterException` is more appropriate.

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -459,4 +460,26 @@ protected function findByRelation(Query $query, array $options = [])
...
+        }
+
+        return $query->innerJoinWith('Objects', function (Query $query) use ($options) {
+            return $query->where(function (QueryExpression $exp) use ($options) {
+                return $exp->or_([$this->Objects->aliasField('id') => $options['id']])
+                    ->add([$this->Objects->aliasField('uname') => $options['id']]);
```
fquffio: I'd find slightly more readable this notation:
```php
return $exp->or_([
    $this->Objects->aliasField('id') => $options['id'],
    $this->Objects->aliasField('uname') => $options['id'],
]);
```

fquffio (changes requested):
I left some comments, but I'm not completely sure that the strategies I suggested would work… 🤔

plugins/BEdita/API/src/Controller/TrashController.php
batopa: It is still used in TrashController::index() listing all trashed objects.

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
batopa: :thinking: I'm not sure simple validation is the right place to check if a folder is restorable. To me it seems more appropriate as application rule because we are checking the state of the application.

Moreover having already the entity we can check against `deleted` is dirty.

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
batopa: I'm not sure I can replicate that query using `exists()` method because it accepts only `$conditions` argument used from query builder in `where()` method.

I can give it a try or replicate what `exists()` do i.e. count the result of
```
->select(['existing' => 1])
->limit(1)
 ->enableHydration(false)
 ->toArray()
```

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
batopa: I had copied that exception from a finder above :smile_cat: 
I'll change it.

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
batopa: More readable, sure. I didn't understand that I can do it.

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
fquffio: Ok then, 👍 for leaving it here.

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
fquffio: Uhm… I'm not sure anymore. This query is only executed in very rare cases, and using `count()` vs `exists()` doesn't add a significant overhead. I'd go with the most readable option.

fquffio (approved):

Issue #1476, opened 2018-05-10
 Shell for recover tree and handling orphan folders


batopa:
We need a shell to easily recover a corrupted tree in `trees` table.

Basically it would launch `TreeBehavior::recover()` method.

Another useful feature would be to locate orphan folders and ask interactively (or not) what operations doing on those:

* `recover`: the folder will become a new root.
* `remove`: the folder will be removed definitively.

PR #1483, opened 2018-05-23
Tree maintenance shell #1476

fquffio:
This PR implements part of the shell requested in #1476.

A new `tree` shell has been added with two tasks:

* `bin/cake tree recover` performs a basic recovery of a corrupt tree, by simply invoking CakePHP's `TreeBehavior::recover()`
* `bin/cake tree check` performs some object-aware checks such as:
  - detect folders that are not in the tree
  - detect ubiquitous folders
  - detect non-folder objects that are roots
  - detect non-folder objects that have descendants
  - detect objects that are present more than once within same parent

It has not been implemented a shell that helps the user recover the tree from these kinds of semantic errors.

### DIFF
--- a/plugins/BEdita/Core/src/Shell/Task/CheckTreeTask.php
+++ b/plugins/BEdita/Core/src/Shell/Task/CheckTreeTask.php
@@ -174,14 +174,13 @@ class CheckTreeTask extends Shell
             ->select([
                 $this->Objects->aliasField('id'),
                 $this->Objects->aliasField('uname'),
-                'count' => $query->func()->count('*'),
             ])
             ->matching('TreeNodes')
             ->group([
                 $this->Objects->aliasField($this->Objects->getPrimaryKey()),
             ])
-            ->having(function (QueryExpression $exp) {
-                return $exp->gt('count', 1, 'integer');
+            ->having(function (QueryExpression $exp) use ($query) {
+                return $exp->gt($query->func()->count('*'), 1, 'integer');
             });
     }
 
@@ -217,15 +216,14 @@ class CheckTreeTask extends Shell
                 $this->Objects->aliasField('object_type_id'),
                 $this->Objects->Parents->aliasField('id'),
                 $this->Objects->Parents->aliasField('uname'),
-                'count' => $query->func()->count('*'),
             ])
             ->matching('Parents')
             ->group([
                 $this->Objects->aliasField('id'),
                 $this->Objects->Parents->aliasField('id'),
             ])
-            ->having(function (QueryExpression $exp) {
-                return $exp->gt('count', 1, 'integer');
+            ->having(function (QueryExpression $exp) use ($query) {
+                return $exp->gt($query->func()->count('*'), 1, 'integer');
             });
     }
 }
--- a/plugins/BEdita/Core/src/Shell/Task/CheckTreeTask.php
+++ b/plugins/BEdita/Core/src/Shell/Task/CheckTreeTask.php
@@ -154,7 +154,7 @@ class CheckTreeTask extends Shell
             $this->verbose('=====> <success>There are no other objects with children.</success>');
         }
 
-        // Checks on other objects with children.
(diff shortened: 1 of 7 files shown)


plugins/BEdita/Core/src/Shell/Task/CheckTreeTask.php
```diff
@@ -0,0 +1,284 @@
...
+                        $entity['uname'],
+                        $entity['id']
+                    )
+                );
+            });
+        } else {
+            $this->verbose('=====> <success>There are no other objects with children.</success>');
+        }
+
+        // Checks on other objects with children.
```
didoda: typo: I guess comment should be something like `Check objects twice in folder`, or somethink alike

plugins/BEdita/Core/src/Shell/Task/CheckTreeTask.php
```diff
@@ -0,0 +1,284 @@
...
+
+            $this->out(sprintf('=====> <warning>Found %d objects that are present multiple times within same parent!</warning>', $count));
+            $results->each(function (EntityInterface $entity) {
+                $this->verbose(
+                    sprintf(
+                        '=====>   - %s <info>%s</info> (#<info>%d</info>) is present multiple times within parent <info>%s</info> (#<info>%d</info>)',
+                        $this->Objects->ObjectTypes->get($entity['object_type_id'])->get('singular'),
+                        $entity['uname'],
+                        $entity['id'],
+                        Hash::get($entity, '_matchingData.Parents.uname', '(unknown)'),
```
didoda: maybe a cast to string would make scrutinizer happy?

plugins/BEdita/Core/src/Shell/Task/CheckTreeTask.php
```diff
@@ -0,0 +1,284 @@
...
+        $query = $this->Objects->find('type', ['folders']);
+
+        return $query
+            ->select([
+                $this->Objects->aliasField('id'),
+                $this->Objects->aliasField('uname'),
+            ])
+            ->innerJoinWith('TreeNodes')
+            ->group([
+                $this->Objects->aliasField($this->Objects->getPrimaryKey()),
```
didoda: Cast to string, to make scrutinizer happy?

plugins/BEdita/Core/src/Shell/Task/RecoverTreeTask.php
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    public function main()
+    {
+        $this->out('=====> <info>Beginning tree recovery...</info>');
+
+        $start = microtime(true);
+        $this->Trees->recover();
```
didoda, line 55: Is it a bug? It isn't, probably... however `@method annotation` could solve it

didoda (approved):
[minor] comments. It's ok

plugins/BEdita/Core/src/Shell/Task/CheckTreeTask.php
fquffio: Right!

plugins/BEdita/Core/src/Shell/Task/RecoverTreeTask.php
fquffio: No, it’s intentional. Scrutinizer apparently doesn’t read the `@mixin TreeBehavior` annotation in `TreesTable` class comment. 😒 My IDE does.

What do you prefer (and think is more accurate) when attaching a behavior to a table: one `@mixin` annotation or several `@method` annotations (one for each implemented method)?

stefanorosanelli (approved):

plugins/BEdita/Core/src/Shell/Task/RecoverTreeTask.php
didoda: nevermind... your choice is fine

PR #1487, opened 2018-05-26
Object status filter on applications #1485

### LINKED ISSUE #1485: Application related configurations and `status` filter
* add `config.application_id` foreign key to load `application` specific configuration.
* handle `status` configuration with possible values:
  - `on` -> only **on** status is allowed
 - `draft` -> only **on** and **draft** status are allowed
 - `all` (default) -> all statuses are allowed

stefanorosanelli:
This PR fixes #1485

* a new `config.application_id` foreign key has been created
* upon application identification `core` context configuration for the selected application is loaded, via `application_id`
* `Staus.level`  configuration introduced with possibile values:
  - `'on'` => only objects with status `on` are loaded
  - `'draft'` => only objects with status `on` or `draft` are loaded
* `BaseAction`, `GetObjectAction` and `ListObjectsAction` have been modified. Are there other actions that need to be changed?
* a new integration test `StatusLevel` was created


plugins/BEdita/API/tests/IntegrationTest/StatusLevelTest.php
```diff
@@ -0,0 +1,115 @@
...
+        $this->assertResponseCode(200);
+        $result = json_decode((string)$this->_response->getBody(), true);
+
+        $ids = Hash::extract($result, 'data.{n}.id');
+        sort($ids);
+        static::assertEquals($expected, $ids);
+    }
+
+    /**
+     * Provider for testListObjects()
```
batopa: typo: it should be `testSingleObject()`

plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
```diff
@@ -40,6 +46,16 @@ class DatabaseConfig implements ConfigEngineInterface
      */
     protected $reservedKeys = ['Datasources', 'Cache', 'EmailTransport', 'Session', 'Error', 'App'];
 
+    /**
+     * Setup application id if
```
batopa: `if` typo ?

plugins/BEdita/Core/src/Model/Action/BaseAction.php
```diff
@@ -72,4 +73,23 @@ protected function initialize(array $config)
     {
         return $this->execute($data);
     }
+
+    /**
+     * Allowed object `status` condition
+     *
+     * @return array Empty array if all `status` are allowed otherwise a list of allowed values
+     */
+    protected function statusCondition()
```
batopa: I'm not sure this is the right place for that piece of logic. `BaseAction` is intended as base class for all actions not necessarily for BEdita Objects action, so  I think is preferable to move elsewhere, maybe in a Trait.

batopa (changes requested):

plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
stefanorosanelli: if... I don't remember :smile_cat:

plugins/BEdita/Core/src/Model/Action/BaseAction.php
stefanorosanelli: I wasn't sure about that either.
Could be a base class, but probably an `ObjectConditionsTrait` is better.

stefanorosanelli: done

plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
stefanorosanelli: fixed

plugins/BEdita/API/tests/IntegrationTest/StatusLevelTest.php
stefanorosanelli: fixed

plugins/BEdita/Core/src/Model/Action/BaseAction.php
```diff
@@ -13,6 +13,7 @@
 
 namespace BEdita\Core\Model\Action;
 
+use Cake\Core\Configure;
```
batopa: unused `use` statement

batopa (approved):

plugins/BEdita/Core/src/Model/Action/BaseAction.php
stefanorosanelli: done

plugins/BEdita/Core/src/Model/Action/GetObjectAction.php
```diff
@@ -54,6 +56,7 @@ public function execute(array $data = [])
         $conditions = [
             'deleted' => (int)!empty($data['deleted']),
         ];
+        $conditions += $this->statusCondition();
```
fquffio: Couldn't we implement and use a custom finder in `ObjectsTable` instead? Finders are inherited from child tables, so it should be alright to use regardless of the actual table being used.

Something like:
```php
public function findStatus(Query $query, array $options)
{
    $mode = reset($options);
    switch ($mode) {
        case 'on':
            return $query->where([$this->aliasField('status') => 'on']);

        case 'draft':
            return $query
                    return $exp->in($this->aliasField('status'), ['on', 'draft']);
                });

        case 'off':
        case 'all':
            return $query;

        default:
            throw new BadFilterException('Invalid mode');
    }
}
```

fquffio:
After private discussions, I made the changes I had suggested in my review. The build is failing due to json-schema.org being currently unreachable: see #1490 for reference.

plugins/BEdita/Core/src/Model/Action/GetObjectAction.php
```diff
@@ -51,21 +53,44 @@ protected function initialize(array $data)
...
+        if (count($key) !== count($primaryKey)) {
+            $primaryKey = $primaryKey ?: [null];
+            $primaryKey = array_map(function ($key) {
+                return var_export($key, true);
+            }, $primaryKey);
+
+            throw new InvalidPrimaryKeyException(sprintf(
+                'Record not found in table "%s" with primary key [%s]',
+                $this->Table->getTable(),
+                implode($primaryKey, ', ')
```
stefanorosanelli: Should this be `implode(', ', $primaryKey)` ?
http://php.net/manual/en/function.implode.php

Also: why this change?
Should we check for multi column primary keys on objects?

fquffio: Fixed. It would work anyway because [`implode()` accepts parameters in either order](http://php.net/implode), but it's definitely way way better how you suggested.

This snippet has been basically copy-pasted from CakePHP's `Table::get()`. I had to duplicate this portion of code because it is impossible to chain multiple finders with `Table::get()`, and as a consequence I had to perform some basic checks on primary key.

We don't need to support multi-column primary keys on objects, but we need to block attempts to pass multiple values for `primaryKey`:
```php
(new GetObjectAction(['table' => TableRegistry::get('Objects')]))
    ->execute(['primaryKey' => [1, 2, 3]]);
// This throws:
// new InvalidPrimaryKeyException(
//     'Record not found in table "objects" with primary key [1, 2, 3]'
// );
```

stefanorosanelli: Got it. Since this is a critical action I would perform this check in a dedicated method and/or add a longer comment to explain why we have to do this....

fquffio: Agreed.

Issue #1486, opened 2018-05-25
Application only configurations


stefanorosanelli:
After #1485 

* handle `app` context where application configurations are stored, used by applications only
* `GET /config` endoint to read those configurations + `core` ones by client applications

PR #1488, opened 2018-05-28
Application configuration #1486

stefanorosanelli:
This PR solves #1486 

To merge after #1485 and  #1487 (this branch starts from there)

* a new `/config` endpoint, only GET method, to retrieve `core` and `app` related configuration - i.e. `core` and `app` context configurations where `application_id` is NULL or set to the current client app
* a new `mine` finder introduced in `ConfigTable` to retrieve this config


plugins/BEdita/API/src/Controller/ConfigController.php
```diff
@@ -0,0 +1,50 @@
...
+     * {@inheritDoc}
+     */
+    public $modelClass = 'Config';
+
+    /**
+     * {@inheritDoc}
+     */
+    public function index()
+    {
+        // Add `mine` filter to load common configuration and
```
fquffio: Is this controller only supposed to serve application configuration, and not be used for CRUD operations?

If that's the case, we might avoid extending `ResourcesController` at all:

```php
/**
 * Controller for `/config` endpoint.
 *
 * @since 4.0.0
 *
 * @property \BEdita\Core\Model\Table\ConfigTable $Config
 */
class ConfigController extends AppController
{

    /**
     * {@inheritDoc}
     */
    public $modelClass = 'Config';

    /**
     * Display available configurations.
     *
     * @return void
     */
    public function index()
    {
        $data = $this->Config->find()
            ->find('mine')
            ->find('context', ['core', 'app']);

        $this->set(compact('data'));
    }
}
```

fquffio (commented):
Seems ok, though I'd probably have chosen to not extend `ResourcesController`.

Also: couldn't this `GET /config` endpoint expose some pieces of configuration that aren't supposed to be exposed? I.e.: credentials, sensitive information, …?

plugins/BEdita/API/src/Controller/ConfigController.php
fquffio: I'd avoid creating an `Action`, too. Actions make sense when there's some logic that's worth making reusable. This piece of code is so trivial and so API-specific that IMHO just isn't worth it.

stefanorosanelli: Agreed, much simpler.
Initially I was thinking to add a single config view and other possible actions, but a trivial index will be enough for most use cases.

With `/admin/config` you can control everything.

stefanorosanelli:
Some configuration keys have to be hidden, currently we have as reserved:
```
'Datasources', 'Cache', 'EmailTransport', 'Session', 'Error', 'App'
```
There are probably more we need to check.
A `hidden` or `reserved` flag in `config` could help?

fquffio:
Most of the configuration keys you have mentioned are not present at all by the time the request is served: they are consumed (= read and deleted) in `bootstrap.php`, so we don't have to worry.

I was wondering what could happen with configurations like OAuth secrets, 3rd party API keys, tokens, …? Maybe some Braintree/Stripe credentials…?

fquffio (approved):

batopa (approved):

PR #1496, opened 2018-06-13
Include related entities in related endpoint

fquffio:
This PR enables the use of `?include` query parameter for `GET /:objectType/:id/:relationName` endpoints.

For instance, `GET /documents/10/attach?include=my_relation` should now yield the expected result.

Some minor refactors were necessary, but there shouldn't be any major change in this patch.

### DIFF
--- a/plugins/BEdita/API/src/Controller/ObjectsController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectsController.php
@@ -159,8 +159,7 @@ class ObjectsController extends ResourcesController
         } else {
             // List existing entities.
             $filter = (array)$this->request->getQuery('filter') + array_filter(['query' => $this->request->getQuery('q')]);
-            $include = $this->request->getQuery('include');
-            $contain = $include ? $this->prepareInclude($include) : [];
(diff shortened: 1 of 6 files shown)


plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -159,8 +159,7 @@ public function index()
         } else {
             // List existing entities.
             $filter = (array)$this->request->getQuery('filter') + array_filter(['query' => $this->request->getQuery('q')]);
-            $include = $this->request->getQuery('include');
-            $contain = $include ? $this->prepareInclude($include) : [];
+            $contain = $this->prepareInclude($this->request->getQuery('include'));
```
stefanorosanelli, line 162: Scrutinizer suggests some type check like:
```$this->prepareInclude((string)$this->request->getQuery('include'))```

but we should then change `prepareInclude` BadRequest test, at least.... :thinking:

stefanorosanelli (approved):

didoda (approved):

PR #1497, opened 2018-06-15
Only canonical plural form as object endpoint

stefanorosanelli:
This PR fixes a problem with object endpoints: only canonical plural form names are allowed.
`GET /document` will return a `404` - until now it was a synonym for `GET /documents`

### DIFF
--- a/plugins/BEdita/API/src/Controller/ObjectsController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectsController.php
@@ -80,7 +80,7 @@ class ObjectsController extends ResourcesController
 
         $type = $this->request->getParam('object_type', $this->request->getParam('controller'));
         try {
-            $this->objectType = TableRegistry::get('ObjectTypes')->get($type);
+            $this->objectType = TableRegistry::get('ObjectTypes')->get($type, ['denySingular' => true]);
             $this->modelClass = $this->objectType->alias;
             $this->Table = TableRegistry::get($this->modelClass);
         } catch (RecordNotFoundException $e) {
--- a/plugins/BEdita/API/src/Controller/ObjectsController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectsController.php
@@ -114,6 +114,7 @@ class ObjectsController extends ResourcesController
     {
         $type = $this->request->getParam('object_type', Inflector::underscore($this->request->getParam('controller')));
         try {
+            /** @var \BEdita\Core\Model\Entity\ObjectType $this->objectType */
             $this->objectType = TableRegistry::get('ObjectTypes')->get($type);
             if ($type !== $this->objectType->name) {
                 $this->log(
--- a/plugins/BEdita/API/src/Controller/ObjectsController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectsController.php
@@ -27,12 +27,14 @@ use Cake\Event\Event;
 use Cake\Network\Exception\ConflictException;
 use Cake\Network\Exception\ForbiddenException;
 use Cake\Network\Exception\InternalErrorException;
+use Cake\Network\Exception\NotFoundException;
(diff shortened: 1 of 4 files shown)


plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -80,7 +80,7 @@ public function initialize()
 
         $type = $this->request->getParam('object_type', $this->request->getParam('controller'));
         try {
-            $this->objectType = TableRegistry::get('ObjectTypes')->get($type);
+            $this->objectType = TableRegistry::get('ObjectTypes')->get($type, ['denySingular' => true]);
```
fquffio: This way is ok, but in this case we don't want to only exclude singular forms, but also IDs and camelized forms.

I'd simply build the query manually this way:
```php
$this->objectType = TableRegistry::get('ObjectTypes')->find()
    ->where(['name' => $type])
    ->firstOrFail();
```
…but this does not benefits from caching. Another option would be:
```php
$this->objectType = TableRegistry::get('ObjectTypes')->get($type);
if ($type !== $this->objectType->name) {
    throw new NotFoundException(__d('bedita', 'Did you mean "{0}"?', $this->objectType->name));
    // or, if we feel like doing so:
    return $this->redirect(/* url rebuilt using correct type */, 304);
}
```

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -195,11 +198,14 @@ public function get($primaryKey, $options = [])
                     ->cache('map', self::CACHE_CONFIG)
                     ->toArray()
             );
-            $allTypes += array_flip(
-                $this->find('list', ['valueField' => 'singular'])
-                    ->cache('map_singular', self::CACHE_CONFIG)
-                    ->toArray()
-            );
+            if (empty($options['denySingular'])) {
```
fquffio: Does this behave nicely with cache? 🤔 I'd say it does, but I'm asking just to be sure.

fquffio (approved):

plugins/BEdita/API/src/Controller/ObjectsController.php
stefanorosanelli: I'd go with the second solution, simplifies PR too.
It provides a suggestion which is a good idea. I'd throw an error without redirect for now.

fquffio: I agree that throwing the error is a much better solution, at least for now.

stefanorosanelli:
@fquffio PR updated:
 * an `ObjectsController::initObjectModel()` method to check correct plural naming
 * removed `denySingular` option

plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -108,6 +101,44 @@ public function initialize()
...
+     *  - $this->Table
+     *
+     * @return void
+     * @throws \Cake\Routing\Exception\MissingRouteException If `object_type` param is not valid
+     */
+    protected function initObjectModel()
+    {
+        $type = $this->request->getParam('object_type', Inflector::underscore($this->request->getParam('controller')));
+        try {
+            /** @var \BEdita\Core\Model\Entity\ObjectType $this->objectType */
```
fquffio, line 117: [minor] This annotation may be omitted: class member is declared and documented at lines 52-57.

fquffio (approved):

plugins/BEdita/API/src/Controller/ObjectsController.php
stefanorosanelli: Agreed, but `Scrutinizer` is complaining probably because we are using `TableRegistry::get('..')->get()`

Seems not solved though....

PR #1500, opened 2018-06-20
Add option to move siblings when updating a child's priority

fquffio:
This PR adds an option to move siblings when updating a child's priority. By default, this option is disabled to preserve backwards compatibility.

```http
PUT /objects/1/children/2
Authorization: Bearer aaaa.bbbb.cccc

{
    "data": {
        "priority": 5,
        "move_siblings": true
    }
}
```

When this option is turned on, all other siblings that are placed between the child's old and new position have their priority either increased or decreased by one.

### DIFF
--- a/bedita-app/controllers/api_base_controller.php
+++ b/bedita-app/controllers/api_base_controller.php
@@ -1419,6 +1419,8 @@ abstract class ApiBaseController extends FrontendController {
      * If 'priority' is not passed then a 400 is thrown
      * $childId must already be a child of $objectId
      *
+     * If `move_siblings` is set to `true`, all siblings with a priority greater than or
+     * equal to the `priority` passed will have their priority increased by one.
      *
      * @param int $objectId the parent object id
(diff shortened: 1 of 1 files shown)


didoda (approved):

didoda:
Brilliant solution!
After this PR is done, we should update the documentation:
https://docs.bedita.net/en/v3.8.0/endpoints/objects.html#update-child-position

fquffio:
Right! I'll open a PR for the docs referencing this one.

Actually there could be a faster solution… running a single query `UPDATE trees SET priority = priority + 1 WHERE parent_id = $parentId AND id <> $childId AND priority BETWEEN $min AND $max` should do the job. However I don't know if this other approach could cause problems with cache or corrupt other things around, so I chose to save rows one at a time.

batopa (approved):
Good job for me! I agree also to save rows one by one to avoid cache issues.

PR #1503, opened 2018-06-26
Password change #1501

### LINKED ISSUE #1501: Allow password change in `/auth/user`
Password change in `/auth/user` should be allowed if previous password is known, like:

```http
PATCH /auth/user
{
  "password": "#####",
  "old_password": "#####",
}
```
Where:
 * `old_password` is the current password, if not valid the operation will fail
 * `password` is the new password

stefanorosanelli:
This PR resolves #1501

Via `PATCH /auth/user` it is possible to update login password providing the current password.


plugins/BEdita/API/src/Controller/LoginController.php
```diff
@@ -207,10 +218,35 @@ public function update()
...
+     * @return void
+     */
+    protected function checkPassword(User $entity, array $data)
+    {
+        if (!isset($data['password'])) {
+            return;
+        }
+
+        if (empty($data['old_password'])) {
+            throw new BadRequestException(__('Missing current password'));
```
fquffio: I'd use `__d('bedita', 'Blah blah')` instead.

plugins/BEdita/API/src/Controller/LoginController.php
```diff
@@ -207,10 +218,35 @@ public function update()
...
+        }
+
+        $hasher = PasswordHasherFactory::build(self::PASSWORD_HASHER);
```
fquffio: Instead of explicitly re-building the hasher, we might access the one used by `FormAuthenticate`:

```php
$hasher = $this->Auth->getAuthenticate('Form')->passwordHasher();
if (!$hasher->check($data['old_password'], $entity->password_hash)) {
    throw new BadRequestException('You\'re basically a bad person');
}
```

This way we could avoid the class constant.

fquffio (approved):

plugins/BEdita/API/src/Controller/LoginController.php
stefanorosanelli: In this context `FormAuthenticate` is not initialized, but only in `login` action.
If we need to re-initialize it we could probably keep my version... :thinking:

batopa (approved):

plugins/BEdita/API/src/Controller/LoginController.php
fquffio: Sorry, I thought it would have worked. Your way is definitely better, then. 👍

PR #1504, opened 2018-06-26
I18n configuration and lang check #1502

### LINKED ISSUE #1502: i18n configuration - allowed and default lang
In order to define project languages an optional configuration is needed:

```php
'I18n' => [
  // list of allowed project language tags
  'languages' => [
    'en' => 'English',
    'de' => 'German',
  ],
 
  // default lang tag - may be null
  'default' => 'en',
];
```

 * `'languages'` may be empty (default), any `RFC 5646  / IETF lang` tag may be used/inserted
 * `'default'` can be `null` (default), otherwise this value is assumed on empty `lang` attribute upon creation

This configuration should also be available in `GET /config` for client apps.

stefanorosanelli:
This PR resolves #1502 

 * a new (optional) `I18n` configuration key is introduced
 * `lang` tag is checked in `ObjectsTable::beforeSave()` - maybe there's a better place to do it?
    - if `lang` is empty and a `default` is configured, `default` lang is used
    - if a set of `languages` are defined and `lang` must be in that set if not empty
 * `eng` replaced with `en` in unit tests according to `RFC 5646`


plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -151,10 +153,30 @@ public function beforeSave(Event $event, EntityInterface $entity)
...
+    /**
+     * Check `lang` tag using `I18n` configuration.
+     *
+     * @param \Cake\Datasource\EntityInterface $entity Entity being saved.
+     * @return void
+     * @throws \Cake\Network\Exception\BadRequestException If a wrong lang tag is specified
+     */
+    protected function checkLangTag(EntityInterface $entity)
+    {
+        $defaultConf = ['languages' => [], 'default' => null];
```
fquffio: Setting the default language should definitely be here. I'd have done it slightly differently, though:
```php
if ($entity->isDirty('lang') && empty($entity['lang']) && Configure::check('I18n.default')) {
    $entity->set('lang', Configure::read('I18n.default'));
}
```

This way, we would set the default only if the `lang` field has been changed somehow. Otherwise it might happen that some old object has the language set when it's receiving modifications of some other nature.

-------------

Also, I'd prefer to move the language validation part in a custom validator. We might add a method `\BEdita\Core\Model\Validation\Validation::languageTag()` that looks more or less like this:
```php
public static function languageTag($tag)
{
    $languages = Hash::normalize((array)Configure::read('I18n.languages'));
    if (!empty($languages)) {
        return array_key_exists($tag, $languages);
    }

    // validate using some evil regex
}
```

Then, in `\BEdita\Core\Model\Validation\ObjectsValidator::__construct()`:
```php
$this
    // ...
    ->scalar('lang')
    ->add('lang', 'languageTag', ['rule' => [Validation::class, 'languageTag']])
    ->allowEmpty('lang')
```

fquffio (approved):

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
stefanorosanelli: Done.
I've added Also a `I18nTest` integration test.

batopa (approved):

fquffio (approved):

PR #1506, opened 2018-06-29
#1495 permanent remove folder

### LINKED ISSUE #1495: Permanent remove folder -> children folders
[originally in #1453]

on a **permanent remove** action on folders, aka _empty trash_, all children folders must be permanently removed as well

batopa:
This PR fixes #1495 

Briefly:

1. Before deleting a folder all **descendants** are found and stored in `descendants` key of `\ArrayObject $options`. The `$options` is chained to `afterDelete()` by CakePHP.
2. After deleting the main folder and consequently its tree node, the **descendants** are deleted passing `$options['_isDescendant'] = true` to avoid useless query on `trees` that is already correct.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/FoldersTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/FoldersTable.php
@@ -205,6 +205,51 @@ class FoldersTable extends ObjectsTable
     }
 
     /**
+     * Prepare all descendants of type "folders" in `$options` to delete later in `static::afterDelete()`.
+     *
+     * ### Options
+     *
+     * `_isDescendant` default empty. When not empty means that the deletion was cascading from a parent so no other action needed.
+     *
+     * @param \Cake\Event\Event $event The event
+     * @param \Cake\Datasource\EntityInterface $entity The folder entity to delete
+     * @param \ArrayObject $options Delete options
+     * @return void
+     */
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Model/Table/FoldersTable.php
```diff
@@ -204,6 +204,51 @@ public function afterSave(Event $event, EntityInterface $entity)
...
+     * @return void
+     */
+    public function afterDelete(Event $event, EntityInterface $entity, \ArrayObject $options)
+    {
+        if (empty($options['descendants'])) {
+            return;
+        }
+
+        foreach ((array)$options['descendants'] as $subfolder) {
+            $this->delete($subfolder, ['_isDescendant' => true]);
```
stefanorosanelli: Use `deleteOrFail` instead?

stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Table/FoldersTable.php
batopa: Sure, definitely better!

fquffio (approved):
For very large trees it might be better to store in memory only the IDs of the descendant folders in `beforeDelete`, and then delete them in chunks in `afterDelete`, but this may become a future optimisation, in case we'll ever need it.

batopa:
@fquffio yes, the same thing was pointed to me from @stefanorosanelli

PR #1508, opened 2018-07-04
Update profiles title also if blank

stefanorosanelli:
This PR changes slightly the `Profiles` before save rule: 
in some applications (like BE4Web) `profiles`/`users` form data are passed like blank if not set or null.

With the previous rule in those cases `title` was never set as expected - for profiles and users title means `display name`

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
@@ -99,7 +99,7 @@ class ProfilesTable extends Table
     /**
      * Before save actions:
      *  - if `email` is empty set it to NULL to avoid unique constraint errors
-     *  - on empty `title` use `name` `surname` as default
+     *  - if `title` is blank use `name` `surname` as default
      *
      * @param \Cake\Event\Event $event The beforeSave event that was fired
      * @param \Cake\Datasource\EntityInterface $entity the entity that is going to be saved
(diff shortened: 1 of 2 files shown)


batopa (approved):

plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
```diff
@@ -110,9 +110,9 @@ public function beforeSave(Event $event, EntityInterface $entity)
         if (empty($entity->get('email'))) {
             $entity->set('email', null);
         }
-        if (!$entity->has('title') && ($entity->has('name') || $entity->has('surname'))) {
+        if ($entity->get('title') === '' && (!empty($entity->get('name')) || !empty($entity->get('surname')))) {
```
fquffio, line 113: If `$entity->get('title') === 'null'`, the title is left unchanged. Is this intentional?

fquffio (approved):

plugins/BEdita/Core/src/Model/Table/ProfilesTable.php
stefanorosanelli: you mean `=== null` ?

at first I used `empty($entity->get('title'))` condition, but, as suggested by @batopa, we probably need a way to be able to reset `title`. So if `title` is `null` it stays `null`

I know it's probably not so intuitive...

PR #1513, opened 2018-07-09
#1484 Sort contents in folder

### LINKED ISSUE #1484: Folder children ordering
API calls must be implemented to change children order inside a folder:

* move a `child` as first or last, or first left and last right
* move a `child` object after or before another `child` object

What's the expected API request and response format?

fquffio:
This PR resolves #1484 by implementing a way to sort children within a Folder.

## API Interface

The following examples assume that a Folder with ID 42 and a Document with ID 13 exist.

#### `POST|PATCH /folders/42/relationships/children`

```http
POST /folders/42/relationships/children

{
  "data": [
    {
      "id": "13",
      "type": "documents",
      "meta": {
        "relation": {
          "position": 3
        }
      }
    }
  ]
}
```

#### `POST|PATCH /documents/13/relationships/parents`

```http
POST /documents/13/relationships/parents

{
  "data": [
    {
      "id": "42",
      "type": "folders",
      "meta": {
        "relation": {
          "position": -2
        }
      }
    }
  ]
}
```

## Code changes

A new `TreeBehavior` has been added, that extends CakePHP `TreeBehavior` and adds two public methods: `getCurrentPosition(EntityInterface $node)` and `moveAt(EntityInterface $node, $position)`.

Some other minor changes have been applied to other pieces of code, mainly to clean up a few things.

### Valid values for `position`

Valid values for position are any integer value except `0` (zero), and the two strings `'first'` and `'last'`:

- `'first'` is an alias for `1`
- `'last'` is an alias for `-1`
- positive values represent the position from the beginning of the list, with `1` being the first element.
- negative values represent the position from the end of the list, with `-1` being the last element.

Values "out of bounds" are silently limited. For instance:
- `10` if the parent has only 5 children is equivalent to `last`.
- `-10` if the parent has only 5 children is equivalent to `first`.

## Known issues

It is not possible to sort a folder position within its parent, as it is not possible to set `position` via the `POST /folders/42/relationships/parent`. This problem can be circumvented for sub-folders (e.g. if Folder 42 is within Folder 101, we can use `POST /folders/101/relationships/children` instead), but at the present stage makes root Folders not-sortable.

A possible solution would be to pass `position` using a payload similar to this:
```http
POST /folders/42/relationships/parent

{
  "data": {
    "id": "101",
    "type": "folders"
  },
  "meta": {
    "relation": {
      "position": "first"
    }
  }
}
```

It is required to place `meta` outside of `data`, since `data` in this case may be `null`. However, such a solution hasn't been implemented in this PR.

### DIFF
--- a/plugins/BEdita/API/src/Error/ExceptionRenderer.php
+++ b/plugins/BEdita/API/src/Error/ExceptionRenderer.php
@@ -116,9 +116,16 @@ class ExceptionRenderer extends CakeExceptionRenderer
         $res = '';
         if (is_array($d)) {
             $d = Hash::flatten($d);
-            foreach ($d as $item => $errDetail) {
-                $res .= "[$item]: $errDetail ";
-            }
+            $res = implode(
+                ' ',
+                array_map(
+                    function ($key, $val) {
+                        return sprintf('[%s]: %s', $key, $val);
+                    },
+                    array_keys($d),
+                    array_values($d)
+                )
+            );
         }
 
         return $res;
--- a/plugins/BEdita/API/tests/IntegrationTest/I18nTest.php
+++ b/plugins/BEdita/API/tests/IntegrationTest/I18nTest.php
@@ -58,7 +58,7 @@ class I18nTest extends IntegrationTestCase
             'error' => [
                 'status' => '400',
                 'title' => 'Invalid data',
-                'detail' => '[lang.languageTag]: Invalid language tag "fi" ',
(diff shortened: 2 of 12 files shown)


fquffio:
I was thinking that maybe using `index` instead of `position` would be a better naming… 🤔 This would apply to both code and API interface. @bedita/core ?

stefanorosanelli (approved):

PR #1525, opened 2018-08-20
Update child position with `POST` action as well

### LINKED ISSUE #1523: Updating child position doesn't work
Trying to change the position of a child **already existing** do nothing.

Example:

Having a `documents` in first position and trying to move to second

```http
POST /folders/12/relationships/children

{
    "data": [
        {
            "id": "6",
            "type": "documents",
            "meta": {
                "relation": {
                    "position": 2
                }
            }
        }
    ]
}
```

The response is `204 No Content` and the position remains unchanged.

### Unit test

```php
public function testUpdateChildrenPosition()
{
    $data = [
        [
            'id' => '2',
            'type' => 'documents',
            'meta' => [
                'relation' => [
                    'position' => 1,
                ],
            ],
        ],
    ];
    $this->post('/folders/11/relationships/children', json_encode(compact('data')));

    $folder = $this->Folders->get(11, ['contain' => ['Children']]);
    $childrenIds = Hash::extract($folder->children, '{n}.id');

    static::assertEquals([2, 12], $childrenIds);
}
```

```
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 2
-    1 => 12
+    0 => 12
+    1 => 2
```

fquffio:
This PR fixes #1523.

The `AddAssociatedAction` has been slightly refactored so that it always saves an association if it detects that join data are present in the request. This is suboptimal because we should check if something actually changed before saving, but it works fine and it shouldn't cause big troubles.

A regression test and a unit test are included in this PR.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
@@ -44,7 +44,20 @@ class AddAssociatedAction extends UpdateAssociatedAction
         foreach ($relatedEntities as $relatedEntity) {
             $primaryKey = $relatedEntity->extract($bindingKey);
             if (in_array($primaryKey, $existing)) {
-                continue;
+                if (!($this->Association instanceof BelongsToMany) || !$relatedEntity->has('_joinData')) {
+                    continue;
+                }
+
+                $relatedEntity->_joinData = $this->Association->junction()->patchEntity(
+                    $this->Association->junction()->find()
+                        ->where([
+                            $this->Association->getForeignKey() => $entity->get($this->Association->getSource()->getPrimaryKey()),
+                            $this->Association->getTargetForeignKey() => $relatedEntity->get($this->Association->getBindingKey()),
+                        ])
+                        ->firstOrFail(),
+                    $relatedEntity->_joinData->toArray()
+                );
+                $relatedEntity->setDirty('_joinData', true);
             }
 
             $diff[] = $relatedEntity;
--- a/plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
@@ -13,6 +13,7 @@
 
 namespace BEdita\Core\Model\Action;
 
+use Cake\Database\Expression\QueryExpression;
 use Cake\Datasource\EntityInterface;
 use Cake\Network\Exception\BadRequestException;
 use Cake\ORM\Association\BelongsToMany;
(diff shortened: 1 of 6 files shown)


plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
```diff
@@ -44,7 +44,20 @@ protected function diff(EntityInterface $entity, array $relatedEntities)
         foreach ($relatedEntities as $relatedEntity) {
             $primaryKey = $relatedEntity->extract($bindingKey);
             if (in_array($primaryKey, $existing)) {
-                continue;
+                if (!($this->Association instanceof BelongsToMany) || !$relatedEntity->has('_joinData')) {
+                    continue;
+                }
+
+                $relatedEntity->_joinData = $this->Association->junction()->patchEntity(
```
didoda: Scrutinizer complains and suggests... do we ignore him?

plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
```diff
@@ -44,7 +44,20 @@ protected function diff(EntityInterface $entity, array $relatedEntities)
...
+                    $this->Association->junction()->find()
```
didoda: good or bad idea?

plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
```diff
@@ -44,7 +44,20 @@ protected function diff(EntityInterface $entity, array $relatedEntities)
...
+                        ->where([
+                            $this->Association->getForeignKey() => $entity->get($this->Association->getSource()->getPrimaryKey()),
```
didoda: again...

didoda (changes requested):
scrutinizer "Bug"(s)...

plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
```diff
@@ -44,7 +44,20 @@ protected function diff(EntityInterface $entity, array $relatedEntities)
...
+                            $this->Association->getTargetForeignKey() => $relatedEntity->get($this->Association->getBindingKey()),
+                        ])
+                        ->firstOrFail(),
+                    $relatedEntity->_joinData->toArray()
+                );
+                $relatedEntity->setDirty('_joinData', true);
```
batopa: Isn't already marked as `dirty` with the assignement?

plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
fquffio: I can use `$relatedEntity->set('_joinData', ...)` in place of `$relatedEntity->_joinData = ...`. Agreed!

plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
fquffio: Scrutinizer is right-ish. `firstOrFail()` can return an array if `->enableHydration(false)` is invoked on the Query object, but this is not our case, so the return value will always be an Entity object.

We might add a type check, but the code would become uselessly less readable and it would be a "ghost branch" that handles an edge case that never happens unless there's a bug in CakePHP. I'd leave it as is.

plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
fquffio: Again, Scrutinizer is right, but an array is returned only if the primary key is composite — i.e. more than one column. If the primary key was composite, however, it wouldn't have been possible to configure this type of association, because CakePHP doesn't handle composite foreign keys, AFAIK. So I'd leave it as is, as it would be a configuration problem.

plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
fquffio: Yes, right!

plugins/BEdita/Core/src/Model/Action/AddAssociatedAction.php
fquffio: No, sorry. My previous comment is wrong, and you were right.
It would be better to handle the case where there are composite primary/foreign keys. 😱

didoda: God forgives, Scrutinizer doesn't

batopa (approved):

didoda (approved):

PR #1530, opened 2018-08-31
Query string filter fields with `null`

stefanorosanelli:
This PR adds missing `null` support on query string fields filter.

~As a result using filters like `filter[lang]=null` or `filter[lang][ne]=null` will provide the expected result: objects with NULL or NOT NULL field are returned.~

**UPDATE**

New filter expressions are available:

* `filter[lang][null]=1` - return objects where `lang` field is `NULL`
* `filter[lang][null]=0` - return objects where `lang` field is `NOT NULL`

### DIFF
--- a/plugins/BEdita/Core/src/ORM/QueryFilterTrait.php
+++ b/plugins/BEdita/Core/src/ORM/QueryFilterTrait.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * BEdita, API-first content management framework
- * Copyright 2017 ChannelWeb Srl, Chialab Srl
+ * Copyright 2018 ChannelWeb Srl, Chialab Srl
  *
  * This file is part of BEdita: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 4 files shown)


fquffio:
This makes it impossible to search for the string "null". Quite uncommon, but a strange limitation. I'd prefer `filter[lang][null]=1` or `filter[lang][null]=0`.

batopa:
Having `null` as string seems very esoteric but I agree with @fquffio 
In that way we can be esoteric and give different meaning to `NULL` and `'NULL'`.

stefanorosanelli:
Ok, let's use `filter[lang][null]=1` or `filter[lang][null]=0`

plugins/BEdita/Core/src/ORM/QueryFilterTrait.php
```diff
@@ -126,4 +97,63 @@ public function fieldsFilter(Query $query, array $options)
...
+            case 'gte':
+            case 'ge':
+            case '>=':
+                $exp = $exp->gte($field, $value);
+                break;
+
+            case 'null':
+                if (empty($value) || $value === '0') {
+                    $exp = $exp->isNotNull($field);
+                } else {
```
batopa: In this case `filter[field][null]=gustavo` results in `field IS NULL`.
Is it correct or we should only accept `'1'`, `'true'` and `true`?

stefanorosanelli: Right.
 I've introduced `filter_var($value, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE);` 

Only `1`, `'1'`, `'true'`, `true`, `0`, `'0'`,`false` and `'false'` are accepted.
Otherwise filter is ignored.

fquffio (approved):

PR #1533, opened 2018-09-10
Existing user signup

stefanorosanelli:
This PR adds a custom exception `code` and event in case of existing user signup.

Useful for clients and be4 plugins to better handle this use case.

### DIFF
--- a/plugins/BEdita/Core/tests/TestCase/Model/Action/SignupUserActionTest.php
+++ b/plugins/BEdita/Core/tests/TestCase/Model/Action/SignupUserActionTest.php
@@ -201,7 +201,6 @@ class SignupUserActionTest extends TestCase
             static::assertTrue(is_string($arguments[3]));
         });
 
-        // Restore this part:
         if ($expected instanceof \Exception) {
             static::expectException(get_class($expected));
         }
--- a/plugins/BEdita/Core/tests/TestCase/Model/Action/SignupUserActionTest.php
+++ b/plugins/BEdita/Core/tests/TestCase/Model/Action/SignupUserActionTest.php
@@ -182,7 +182,7 @@ class SignupUserActionTest extends TestCase
     /**
      * Test command execution.
      *
-     * @param array|\Exception $expected Expected result.
+     * @param bool|\Exception $expected Expected result.
      * @param array $data Action data.
      * @return void
      *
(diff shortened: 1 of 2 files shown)


didoda (approved):

didoda:
Exception code for "already existing username" is good to handle the use case. Not necessary, but useful: a suggestion for available usernames, "similar" to the "already existing" one. I guess this could be addressed in a specific new issue.

plugins/BEdita/Core/tests/TestCase/Model/Action/SignupUserActionTest.php
```diff
@@ -180,13 +201,22 @@ public function testExecute($expected, array $data)
...
-        $result = $action($data);
+        try {
+            $action = new SignupUserAction();
+            $result = $action($data);
+            $success = true;
+        } catch (CakeException $e) {
+            $success = $e;
+            static::assertEquals($expected->getAttributes(), $e->getAttributes());
+            static::assertEquals($expected->getCode(), $e->getCode());
+        }
```
fquffio: I'd change this test a bit:

```php
// Restore this part:
if ($expected instanceof \Exception) {
    static::expectException(get_class($expected));
}

// ...

try {
    $action = new SignupUserAction();
    $result = $action($data);
} catch (CakeException $e) {
    static::assertInstanceOf(CakeException::class, $expected); // Ensure we're expecting an exception of the same kind.
    static::assertEquals($expected->getAttributes(), $e->getAttributes());
    static::assertEquals($expected->getCode(), $e->getCode());

    throw $e; // Re-throw exception.
}

// Leave below as it was before.
```

Basically, the test remains as it was before. We only catch the exception for the time strictly necessary to perform some additional assertions on it, and then re-throw it to continue the usual flow.

stefanorosanelli: Seems ok, applying what you suggest

stefanorosanelli: done

plugins/BEdita/Core/tests/TestCase/Model/Action/SignupUserActionTest.php
```diff
@@ -201,22 +201,26 @@ public function testExecute($expected, array $data)
             static::assertTrue(is_string($arguments[3]));
         });
 
+        // Restore this part:
```
fquffio: This line was not necessary. 😄

fquffio (approved):

plugins/BEdita/Core/tests/TestCase/Model/Action/SignupUserActionTest.php
stefanorosanelli: fixed

batopa (approved):

Issue #1494, opened 2018-06-08
I18n [3/3] - `lang` query string


didoda:
Provide `lang` filter for `GET`

(see https://github.com/orgs/bedita/teams/core/discussions/6?from_comment=7#discussion-6-comment-7)

`lang` query filter on an object or list of objects will add requested translations to `included` section.

PR #1534, opened 2018-09-11
Introduce `lang` query string #1494 

stefanorosanelli:
This PR solves #1494 

* `lang` query string used on single object and list GET
* only single language tag supported for now
* translation `status` not checked

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/GetObjectAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/GetObjectAction.php
@@ -74,6 +74,11 @@ class GetObjectAction extends BaseAction
         if (Configure::check('Status.level')) {
             $query = $query->find('status', [Configure::read('Status.level')]);
         }
+        if (!empty($data['lang'])) {
+            $query->contain('Translations', function ($q) use ($data) {
+                return $q->where(['Translations.lang' => $data['lang']]);
+            });
+        }
 
         return $query->firstOrFail();
     }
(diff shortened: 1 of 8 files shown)


batopa (approved):

plugins/BEdita/Core/src/Model/Action/GetObjectAction.php
```diff
@@ -74,6 +74,11 @@ public function execute(array $data = [])
         if (Configure::check('Status.level')) {
             $query = $query->find('status', [Configure::read('Status.level')]);
         }
+        if (!empty($data['lang'])) {
+            $query->contain('Translations', function ($q) use ($data) {
```
fquffio: Since this is being repeated in two places already, one might add a `translations` finder to Objects table:

```php
protected function findTranslations(Query $query, array $options)
{
    return $query->contain('Translations', function (Query $query) use ($options) {
        return $query->where(['Translations.lang' => $options['lang']]);
    });
}
```

Therefore, here and in `ListObjectsAction` would become:
```php
$query = $query->find('translations', ['lang' => $data['lang']]);
```

fquffio (approved):
Just a suggestion for a minor refactor to make part of this PR's logic slightly more reusable.

plugins/BEdita/Core/src/Model/Action/GetObjectAction.php
stefanorosanelli: Done!

fquffio (approved):

Issue #1535, opened 2018-09-11
OTP Auth flow


stefanorosanelli:
Implement OTP auth flow as described in https://github.com/orgs/bedita/teams/core/discussions/9

PR #1537, opened 2018-09-13
OTP auth flow #1535

stefanorosanelli:
This PR solves #1535 

* a new `user_tokens` table was created to save one time passwords as well as `refresh` tokens and other items regarding user auth
* a new `OTPAuthenticate` class to handle OTP Request and OTP Access flow as described in https://github.com/orgs/bedita/teams/core/discussions/9
  + client authorization codes are generated via UUID
  + secret tokens are generated as random numeric strings as default, but custom generatos can be defined
* `LoginController::login()` was modified in order to handle this new flow in `POST /auth` as well


plugins/BEdita/API/src/Auth/OTPAuthenticate.php
```diff
@@ -0,0 +1,209 @@
...
+     * @return string The generated secure token
+     */
+    public static function defaultSecretGenerator()
+    {
+        $res = '';
+        for ($i = 0; $i < 6; $i++) {
+            $res .= mt_rand(0, 9);
+        }
+
+        return $res;
```
fquffio: This is cryptographically insecure. Maybe it's not a big deal, but it would be enough to replace this method with something like this:

```php
return sprintf('%06d', hexdec(bin2hex(Security::randomBytes(2))));
```

plugins/BEdita/Core/config/Migrations/20180911142444_AddUserTokensTable.php
```diff
@@ -0,0 +1,150 @@
...
+                'comment' => 'secret token sent to user in a secure way',
+                'default' => null,
+                'limit' => 255,
+                'null' => true,
+            ])
+            ->addColumn('token_type', $enum, [
+                'comment' => 'type of token: otp, refresh, recovery, 2fa, access',
+                'default' => 'refresh',
+                'limit' => 255,
+                'values' => ['otp', 'refresh', 'recovery', '2fa', 'access'],
```
fquffio: Enforcing this constraint on the database seems a bit hardcore: if we wan to add new types we need to bake a migration. I'd enforce it only on the application level, and leave this a `VARCHAR(255)`.

plugins/BEdita/Core/config/Migrations/20180911142444_AddUserTokensTable.php
```diff
@@ -0,0 +1,150 @@
...
+                ]
+            )
+            ->addIndex(
+                [
+                    'secret_token',
+                ],
+                [
+                    'name' => 'usertokens_secrettoken_idx',
+                ]
+            )
```
fquffio, line 109: Two more indexes could be useful: `expires` and `used`. I would expect many queries built this way:

```sql
SELECT * FROM user_tokens
  WHERE
    /* conditions on user_id, application_id, token_type, … */
    AND used IS NULL -- !!
    AND expires > NOW() --!!
    LIMIT 1;
```

plugins/BEdita/Core/src/Model/Table/UserTokensTable.php
```diff
@@ -0,0 +1,147 @@
...
+            ->notEmpty('client_token');
+
+        $validator
+            ->scalar('secret_token')
+            ->maxLength('secret_token', 255)
+            ->allowEmpty('secret_token');
+
+        $validator
+            ->scalar('token_type')
+            ->requirePresence('token_type', 'create')
```
fquffio, line 100: I'd validate the type too:

```php
$validator
  // ...
  ->inList('token_type', ['foo', 'bar'])
  // ...
```

fquffio (changes requested):

plugins/BEdita/API/src/Auth/OTPAuthenticate.php
stefanorosanelli: I know didn't want to introduce cryptographic algorithm, but I'd use same used for passwords.

plugins/BEdita/Core/config/Migrations/20180911142444_AddUserTokensTable.php
stefanorosanelli: Still not sure how much we'll use `used`, but I'll add both

plugins/BEdita/Core/src/Model/Table/UserTokensTable.php
stefanorosanelli: Ok

plugins/BEdita/Core/config/Migrations/20180911142444_AddUserTokensTable.php
fquffio: Recovery tokens for MFA, for instance, should definitely be kept in this table, for security audits.

fquffio (approved):

plugins/BEdita/API/src/Auth/OTPAuthenticate.php
```diff
@@ -0,0 +1,205 @@
...
+     * @return array|bool User data array on success, false on failure
+     */
+    protected function otpAccess($username, ServerRequest $request)
+    {
+        $result = $this->_findUser($username);
+        if (empty($result)) {
+            return false;
+        }
+
+        if (empty($request->getData('authorization_code')) || empty($request->getData('token'))) {
```
batopa: [minor] I would move this check before of  `_findUser()` to avoid useless query on db if missing required params.

batopa (commented):
ok for me! Just a minor comment

batopa (approved):

PR #1538, opened 2018-09-18
Refactor cache using Redis hashes

fquffio:
This PR introduces a refactor on how cache keys are handled in Redis and File engines.

## Background

Since many cache keys are relative to a single object ID, we used to maintain an "index" that stores a list of cache keys related to that object ID. Such index was then read at the time we needed to invalidate cache for that object to know the list of cache keys that had to be deleted. This is problematic in case of concurrent writes to that index, since some keys could go lost thus never invalidated. Even worse, if the index goes lost for some reason (expired or whatever), all the cache keys generated until that point become orphaned and no one will ever clear them. At a high level, this resulted in inconsistent results in requests and API calls that were caused by outdated cache keys.

Generally speaking, knowing that a cache key's invalidation depends on the correctness of another cache key is, in my opinion, bad practice, as _every_ cache key should be considered ephemeral and volatile, while in our case this was true only for _groups_ of keys.

## Overview of the changes

After this refactor, all the cache keys related to an object ID, with the exception of the nickname-to-ID cache, are grouped under a single cache key, that is either an hash if cache engine is Redis, or a folder if cache engine is File. This means that all cache keys belonging to an object can be invalidated at once by just deleting that cache key, making invalidation process faster and more reliable.

Nickname cache, on the other hand, is now invalidated every time an object is saved or deleted, independently from the other cache keys.

## Considerations about backwards compatibility

All cache keys generated before this refactor become useless, and will then be regenerated. To clear useless cache keys and reclaim cache space, it is advisable to clear cache just after updating BEdita.

## Impact on modules and frontends.

Most modules and frontends do not interact with objects' cache directly, so I expect that the impact of this PR on dependencies' code to be very low.

## Other engines

Other engines supported by CakePHP include Memcache and APCu. Such engines have not been taken into account by these changes, so those are to be considered unsupported.

### DIFF
--- a/bedita-app/models/behaviors/cacheable.php
+++ b/bedita-app/models/behaviors/cacheable.php
@@ -191,9 +191,14 @@ class CacheableBehavior extends ModelBehavior {
      * @param bool $reset
      */
     public function clearCache(&$model, $objectId, $reset = true) {
-        if ($objectId && $reset) {
-            $this->setObjectsToClean($model, $objectId);
+        if ($objectId) {
+            if ($reset) {
+                $this->resetObjectsToClean($model);
+            }
+
+            $this->addObjectsToClean($model, $this->getObjectsToCleanById($model, $objectId));
         }
+
         foreach ($this->objectsToClean as $id) {
             $this->BeObjectCache->delete($id);
         }
--- a/bedita-app/models/behaviors/cacheable.php
+++ b/bedita-app/models/behaviors/cacheable.php
@@ -188,9 +188,10 @@ class CacheableBehavior extends ModelBehavior {
      *
      * @param Model $model
      * @param integer $objectId
+     * @param bool $reset
      */
-    public function clearCache(&$model, $objectId = null) {
-        if ($objectId) {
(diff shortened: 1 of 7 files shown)


didoda (approved):
Code fine!!! Difficult to test deeply in a non-prod environment (it needs concurrent writes and accesses, to be deeply tested)

bedita-app/models/behaviors/cacheable.php
```diff
@@ -188,9 +188,10 @@ protected function getDescendantSections($id) {
      *
      * @param Model $model
      * @param integer $objectId
+     * @param bool $reset
      */
-    public function clearCache(&$model, $objectId = null) {
-        if ($objectId) {
+    public function clearCache(&$model, $objectId, $reset = true) {
+        if ($objectId && $reset) {
```
batopa: if `$reset === false` maybe we need to add `$objectId` to current objects list to clean with `$this->addObjectsToClean()`?

fquffio: Right… 😞 should be fixed now.

batopa (approved):

PR #1539, opened 2018-09-18
`profiles.email` non unique index

stefanorosanelli:
This PR fixes some problems related to `profiles.email` uniqueness: it is now a `normal` index.

`email` is no more unique for `Profiles`, but a validation rule for `email` uniqueness on `Users` has been maintained


batopa (approved):

plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -315,4 +315,15 @@ public function beforeMarshal(Event $event, \ArrayObject $data)
...
         }
     }
+
+    /**
+     * Check for users email uniqueness
+     *
+     * @param mixed $value The value of column to be checked for uniqueness
+     * @return bool true if the value is unique
+     */
+    public function validateUniqueEmail($value)
```
batopa: I think we could use the CakePHP core validation rule `Table::validateUnique()`

batopa (changes requested):
If it's possible I would add an application rules in `UsersTable` that check the email uniqueness.

Indeed the validation layer rules are applied only when `Table::newEntity()` or `Table::patchEntity()` are used but doing

```php
$table = \Cake\ORM\TableRegistry::init('Users)`;
$user = $table->newEntity();
$user->email = 'g.supporto@channelweb.it';
$table->save($user);
```
will skip the validation rules but not the application rules that are intended to comparing data against the existing state of the application.

https://book.cakephp.org/3.0/en/orm/validation.html#validation-vs-application-rules

plugins/BEdita/Core/src/Model/Table/UsersTable.php
stefanorosanelli: done

batopa (approved):

PR #1540, opened 2018-09-20
#1422 collect application rules errors inherited in main entity

### LINKED ISSUE #1422: Error saving user with duplicated `email`
Creating or modifying `users` with a duplicated `email` result in a `500` `Save failed` error instead of a `400`

Example request `POST {apiUrl}/users` with body:
```json
{
    "data": {
        "type": "users",
        "attributes": {
			"username":"gustavo",
			"email":"{existing-email}"
		}
        }
}
```
Error comes from `Cake\ORM\Table::_processSave()` where
```php
        if ($options['checkRules'] && !$this->checkRules($entity, $mode, $options)) {
            return false;
        }
```

returns false in CTI as `Profiles` is the current table,
Same request on `/profiles` will return
```json
{
    "error": {
        "status": "400",
        "title": "Invalid data",
        "detail": "[email._isUnique]: This value is already in use "
    }
}
```

batopa:
This PR fixes #1422 

The CTI was missing of handle application rules errors in the right way losing errors info of inherited tables.

A new event listener on `Model.afterRules` event has been added to intercept application rules failure and collect other errors from inherited tables.
All errors found are set in the main entity, so doing

```php
$users = \Cake\ORM\TableRegistry::get('Users');
$user = $users->newEntity([
    // data here
]);
$res = $users->save($user);
if ($res === false) {
    debug($user->getErrors());
}
```
The entity `$user`  contains all application rules errors (if any) of the whole inheritance i.e. `users -> profiles -> objects`

### DIFF
--- a/plugins/BEdita/Core/src/ORM/Inheritance/InheritanceEventHandler.php
+++ b/plugins/BEdita/Core/src/ORM/Inheritance/InheritanceEventHandler.php
@@ -40,7 +40,49 @@ class InheritanceEventHandler implements EventListenerInterface
                 'callable' => 'afterDelete',
                 'priority' => 1,
             ],
+            'Model.afterRules' => [
+                'callable' => 'afterRules',
+                'priority' => 1,
+            ],
         ];
+    }
+
+    /**
+     * If application rules fail it collects all errors in inherited tables.
+     *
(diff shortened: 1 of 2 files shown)


stefanorosanelli (approved):

fquffio (approved):
🎩

PR #1541, opened 2018-09-21
`parent` filter using name

stefanorosanelli:
Current `parent` filter works only by `id` via `parent` association.

A new `parent` finder method has been introduced in order to use either `name` or `id`.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
@@ -349,6 +349,24 @@ class ObjectTypesTable extends Table
     }
 
     /**
+     * Find object types having a parent by `name` or `id`
+     *
+     * @param \Cake\ORM\Query $query Query object.
+     * @param array $options Additional options. The first element containing `id` or `name` is required.
+     * @return \Cake\ORM\Query
+     * @throws \BEdita\Core\Exception\BadFilterException When missing required parameters.
+     */
+    public function findParent(Query $query, array $options)
+    {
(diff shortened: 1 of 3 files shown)


didoda (approved):

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
```diff
@@ -348,6 +348,24 @@ public function findAll(Query $query, array $options)
...
+     * @param array $options Additional options. The first element containing `id` or `name` is required.
+     * @return \Cake\ORM\Query
+     * @throws \BEdita\Core\Exception\BadFilterException When missing required parameters.
+     */
+    public function findParent(Query $query, array $options)
+    {
+        if (empty($options[0])) {
+            throw new BadFilterException(__d('bedita', 'Missing required parameter "{0}"', 'parent'));
+        }
+        $parentId = $this->get($options[0])->get('id');
```
fquffio: Loading the whole object type is not necessary if the option is already numeric:
```php
$parentId = $options[0];
if (!is_numeric($parentId)) {
    $parentId = $this->get($parentId)->id;
}
```

Or, probably even better, we could extract the part of `ObjectTypes::get()` that builds the name-to-id map to a standalone method and use it here.

fquffio (approved):

plugins/BEdita/Core/src/Model/Table/ObjectTypesTable.php
stefanorosanelli: Ok, did only the numeric check

PR #1545, opened 2018-10-13
Add `include` in `/auth/user`

stefanorosanelli:
This PR adds `?include` query string support in `GET /auth/user`

 * `ResourcesController::prepareInclude()` method moved to `AppController` - better use a Trait instead?
* `findAssociation()` method will throw an exception if not defined in controller class
* test cases update: `another_test` relation now links `users` with `locations` -- many file changes as a consequence... I'm sorry :disappointed:


batopa (approved):
I would prefer (in another issue) to extract some general method useful to build JSON API response in a trait and use it when it needs.

For now, it's ok for me.

fquffio (approved):

PR #1546, opened 2018-10-15
fix: after save section clear only cache of parent

MichelaTarozzi:
After saving a section, the cache of section's parent and all descending elements was cleaned.
After this pull request only the parent's cache is cleared.

### DIFF
--- a/bedita-app/models/behaviors/cacheable.php
+++ b/bedita-app/models/behaviors/cacheable.php
@@ -190,7 +190,7 @@ class CacheableBehavior extends ModelBehavior {
      * @param integer $objectId
      * @param bool $reset
      */
-    public function clearCache(&$model, $objectId, $reset = true) {
+    public function clearCache(&$model, $objectId, $reset = true, $descendants = true) {
         if ($objectId) {
             if ($reset) {
                 $this->resetObjectsToClean($model);
@@ -205,7 +205,7 @@ class CacheableBehavior extends ModelBehavior {
 
         $notLeafs = array(Configure::read('objectTypes.area.id'), Configure::read('objectTypes.section.id'));
         if (empty($model->data['BEObject']['object_type_id']) || in_array($model->data['BEObject']['object_type_id'], $notLeafs)) {
-            $this->BeObjectCache->deletePathCache($objectId, $this->getDescendantSections($objectId));
+            $this->BeObjectCache->deletePathCache($objectId, $descendants ? $this->getDescendantSections($objectId) : array());
         }
 
         if (!empty($this->objectsToClean)) {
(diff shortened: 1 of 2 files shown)


batopa:
I have some doubt.

Usually the cache of an object contains some information on its parent path. This means that a descendant of a section will have cached something like

```php
'pathSection' => [] // array of its parents (only for section objects)
'canonicalPath' => '/publication-nick/ancestor-section/parent-section/current-section' // for all objects
```

Saving, for example, `ancestor-section` with a new nickname `ancestor` or changing the status from `on` to `draft` should delete the cache of all descendants because those fields will be invalid in the descendants cache. Moreover for descendants of type `section` other fields are involved (for example the  `title`) because `pathSection` is an array of sections.

Am I missing something?

:thinking:  Maybe we should mitigate the delete propagation on descendants checking if some critical field was changed (`status`, `nickname`) and if not we could delete the cache only for descendants of type section.

fquffio:
I'm afraid @batopa is right… Let's consider a simple tree like this:

```
P
|-A
| |-B
| \-C
\-X
  |-Y
  \-Z
```

When we save section **A** we want the following caches to be invalidated: **P**, **A**, **B**, **C**.

```php
function afterSave($created) {
    if (!empty($this->data[$this->name]['parent_id'])) {
        $this->BEObject->clearCache($this->data[$this->name]['parent_id'], true, false);
    }
    // ...
}
```

This piece of code aims to invalidate the cache of the _parent_ section only (if any) — in the previous example, **P**. Passing `true` as the third argument would invalidate the cache of parent's descendant, which is exactly what happened before. Cascading here would mean that every _siblings_' cache is invalidated too, and that's something we don't want 'cause siblings know nothing about each other so there's no reason to wipe their cache out — in my example, there's no reason to invalidate cache for **X**, **Y** and **Z** when **A** is being saved.

However, with these changes there's nothing invalidating cache of **B** and **C** any more! So perhaps we should split the invalidation in two parts:

```php
function afterSave($created) {
    // This cascades invalidating A, B and C:
    $this->BEObject->clearCache($this->id); // I'm not sure `$this->id` exists…

    if (!empty($this->data[$this->name]['parent_id'])) {
        // This DOES NOT cascade, and invalidates P only.
        $this->BEObject->clearCache($this->data[$this->name]['parent_id'], true, false);
    }
    // ...
}
```

**PS**: also… let's assume there's some section **Q** above **P**. Should its cache be invalidated too? This does not happen now, nor it did happen before.

batopa (approved):

fquffio (approved):

fquffio:
After internal discussion we came to the conclusion that my solution caused a redundant invalidation of the cache, thus we asked @MichelaTarozzi to revert her commit 8ffd65a.

Thanks and sorry for the confusion!

Issue #1552, opened 2018-10-23
Default `status` for new objects


stefanorosanelli:
Currently all new objects are created with status `draft` if no status is set.
On some object types different defaults should be available.

We could extend this feature to other object properties.

Should this be defined via configuration or as default model properties in object types?
Should this feature be application dependent?

didoda:
Will we really need custom defaults by object type (anytime soon)? maybe `YAGNI`.
I suggest that we start with a basic solution in configuration (`config/app.php`), i.e.:
```
'Object' => [
    'defaults' => [
        'status' => 'on',
    ],
],
```
Then we could think about a "default model properties in object types" solution, when it will be required (in a separate, future, issue).

@batopa , @fquffio , what do you think about it?

stefanorosanelli:
In BE3 we had only a `defaultStatus` configuration for every object type.
If we want to use a configuration for that I'd use something like:

```php
'DefaultValues' => [
   'images' => [
      'status' => 'on',
   ],
]
```

Probably listing only concrete types for simplicity.
Don't know if we should create a link to applications.

PR #1554, opened 2018-10-24
#1552 feat: default fields from config per object type

didoda:
This PR provides #1552 feature(s).

Fields values per object can be set in configuration, i.e.:
```
'DefaultValues' => [
    'cats' => [
        'status' => 'off',
    ],
    'dogs' => [
        'status' => 'on',
    ],
],
```


config/app.default.php
```diff
@@ -520,4 +520,16 @@
         // default lang tag - may be null
         'default' => null,
     ],
+
+    /**
+     * Default values per object type
+     */
+    'DefaultValues' => [
+        'documents' => [
```
stefanorosanelli: Add some more comment on array structure: object type names as keys, default property names and values as value.

This block should be commented in `app.default.php` since it can be copied as starting `app.php`

I'll remove `'status'` => `'draft'` since it's already a default...

plugins/BEdita/Core/src/Model/Behavior/DataDefaultsBehavior.php
```diff
@@ -51,6 +53,11 @@ class DataCleanupBehavior extends Behavior
     public function beforeMarshal(Event $event, \ArrayObject $data)
     {
         $config = $this->getConfig();
+        $objectType = strtolower($this->_table->getAlias());
```
stefanorosanelli:
```suggestion
        $objectType = Inflector::underscore($this->_table->getAlias());
```
`MyTypes` should become `my_types` not `mytypes`

plugins/BEdita/Core/src/Model/Behavior/DataDefaultsBehavior.php
```diff
@@ -51,6 +53,11 @@ class DataCleanupBehavior extends Behavior
...
+        $defaults = Configure::read(sprintf('DefaultValues.%s', $objectType), []);
+        if (!empty($defaults)) {
+            $config['fields'] = array_merge($config['fields'], $defaults);
```
stefanorosanelli: [super minor] empty array may be safely merged without `empty()` check...

stefanorosanelli (changes requested):

plugins/BEdita/Core/src/Model/Behavior/DataDefaultsBehavior.php
didoda: :+1: applyed in latest commit

stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Behavior/DataCleanupBehavior.php
```diff
@@ -51,6 +54,9 @@ class DataCleanupBehavior extends Behavior
     public function beforeMarshal(Event $event, \ArrayObject $data)
     {
         $config = $this->getConfig();
+        $objectType = Inflector::underscore($this->_table->getAlias());
```
batopa: Maybe it's better take the "real" object type `name` since the alias could is different for example for relation.

```php
if ($this->_table->behaviors()->has('ObjectType')) {
     $defaults = Configure::read(sprintf('DefaultValues.%s', $this->_table->objectType()->get('name')), []);
     $config['fields'] = array_merge($config['fields'], $defaults);
}
```

But since the behavior could be used with any table (not only tables that represents object type) we could also generalize as

```php
$key = Inflector::underscore($this->_table->getAlias());

if ($this->_table->behaviors()->has('ObjectType')) {
     $key = $this->_table->objectType()->get('name');
}

$defaults = Configure::read(sprintf('DefaultValues.%s', $key), []);
$config['fields'] = array_merge($config['fields'], $defaults);
```

batopa (changes requested):

batopa (approved):

fquffio (approved):

PR #1557, opened 2018-10-30
Fix fatal error on model save if object cache is disabled

fquffio:
This PR fixes an issue that caused a fatal error ☠️ on model save if objects cache was disabled.

### DIFF
--- a/bedita-app/models/behaviors/cacheable.php
+++ b/bedita-app/models/behaviors/cacheable.php
@@ -191,16 +191,16 @@ class CacheableBehavior extends ModelBehavior {
      * @param bool $reset
      */
     public function clearCache(&$model, $objectId, $reset = true, $descendants = true) {
+        if (!$this->on) {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


bedita-app/models/behaviors/cacheable.php
```diff
@@ -199,6 +199,10 @@ public function clearCache(&$model, $objectId, $reset = true, $descendants = tru
             $this->addObjectsToClean($model, $this->getObjectsToCleanById($model, $objectId));
         }
 
+        if (!$this->on) {
```
batopa: Why not at the beginning of the method?

fquffio: I thought that maybe something may be relying on the `$this->addObjectsToClean()` call… but maybe not. 🤔 

I'll move it at the beginning.

batopa: Maybe yes, but I think it's more clear there.

PR #1559, opened 2018-11-06
VTT Captions

fquffio:
This PR adds model **Captions** for video captions.

**Captions** extend annotations and are linked to a **Video**. There may be more than one **Caption** for each **Video**, and captions may be in different languages and have different publication statuses.

- [x] Add model **Caption** and ensure the correct **ObjectType** is added via database migrations.
- [x] Ensure captions are converted from SRT format to VTT for data consistency.
- [x] Leverage management of **Caption** model through the **Video** model — data is transparently read, created, updated and deleted via the `captions` property of videos.
- [ ] Develop an awesome UI to let users manage captions with ease — @edoardocavazza


bedita-app/models/objects/content/video.php
```diff
@@ -24,8 +24,101 @@
...
+            }
+        );
+
+        $CaptionModel = ClassRegistry::init('Caption');
+
+        foreach ($data as $datum) {
+            $datum['object_id'] = $videoId;
+            $CaptionModel->save($datum);
+        }
+
```
didoda, line 114: superminor: you could add an `unset($datum)` (as done previously... with `unset($result)`).

bedita-app/views/multimedia/inc/form_captions.tpl
```diff
@@ -0,0 +1,43 @@
...
+{if $object.object_type_id == $conf->objectTypes['video']['id']}
+<div class="tab"><h2>{t}Captions{/t}</h2></div>
+<div id="captions" data-start-idx="">
+
+    <table class="indexlist">
+		<thead>
+			<tr>
+				<th>{t}status{/t}</th>
+				<th>{t}language{/t}</th>
+				<th>IGHI</th>
```
didoda: :-)

didoda (approved):

bedita-app/views/multimedia/inc/form_captions.tpl
fquffio: Ooops.

bedita-app/models/objects/caption.php
```diff
@@ -0,0 +1,89 @@
...
+                $end = str_replace(',', '.', $match['end']);
+
+                return sprintf('%s --> %s', $start, $end);
+            },
+            $contents
+        );
+
+        return $contents;
+    }
+}
```
batopa: Add new line at the end of file ;)

bedita-app/config/sql/upgrade/1-3.8-to-3.9-schema.sql
```diff
@@ -1 +1,3 @@
-ALTER TABLE `objects` ADD INDEX `objects_publisher_idx` (`publisher`);
\ No newline at end of file
+ALTER TABLE `objects` ADD INDEX `objects_publisher_idx` (`publisher`);
+
+INSERT INTO `object_types` (`id`, `name`, `module_name`) VALUES (42, 'caption', 'multimedia') ON DUPLICATE KEY UPDATE `id` = 42;
```
batopa: Add new line please

bedita-app/models/objects/content/video.php
fquffio: The `unset($result)` is put there for "safety" reasons. Since the foreach loop iterates using a symbol passed by reference `&$result`, if we used `$result` for something else _after_ the loop, this would affect the last element in the iteration:

```php
$arr = [1, 2, 3];
foreach ($arr as &$item) {
  $item = $item * 2;
}
// Now: $arr === [2, 4, 6]

$item = 'hello world';
// Now: $arr === [2, 4, 'hello world'] 😱
```

The `unset($result)` "looses" the reference and makes it safe to re-use the symbol, so I usually add it every time I do a `foreach ($arr as &$item)` loop so that I won't get burnt in the future should I make changes to the same part of code. In this case, since `$datum` ain't a reference, it shouldn't be necessary. But it's perfectly OK to add it anyway if you wish.

didoda: ok, not necessary. thanks for the detailed explanation ;)

bedita-app/config/sql/upgrade/1-3.8-to-3.9-schema.sql
didoda: in my local db I have `questionnaire_result` with `object_type_id` 42. Should we be worried about it? (42 already in use). I think that 43 would be fine :)

Take a look at https://github.com/bedita/bedita/blob/3-corylus/bedita-app/config/sql/upgrade/0-3.0-to-3.1-data.sql#L8

batopa (approved):

didoda (approved):

PR #1560, opened 2018-11-16
feat: add scrutinizer

didoda:
This PR add scrutinizer configuration to project to enable scrutinizer runs on push


didoda:
not necessary

PR #1562, opened 2018-11-17
Issue/1561 status filter

### LINKED ISSUE #1561: [API] filter[status] weird behavior
filter[status] = on works correctly
filter[status] = draft,off always return on, draft object (probably it uses some default config)

stefanorosanelli:
This PR fixes #1561 

* `status` finder was renamed to avoid conflict with query string `filter[status]`
* tests with `filter[status]` query string have been added


batopa:
What happens when `statusLevel` is set at `on` and `filter[status]=draft`?

stefanorosanelli:
Filter will give 0 results anyway: no `draft` objects are visible.
It's in the tests with `filter[status]=off`

batopa (approved):

PR #1564, opened 2018-11-20
Set perms for webroot files in entrypoint.sh

didoda:
This PR move some `chmod` logic from `Dockerfile` to `entrypoint.sh`.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -18,7 +18,7 @@ if [ ! -z "${DATABASE_URL}" ]; then
         bin/cake migrations seed -p BEdita/Core --seed AdminFromEnvSeed
     fi
 
-    chmod a+rwx /var/www/html/webroot/_files
+    chmod a+rwX /var/www/html/webroot/_files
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -18,6 +18,8 @@ if [ ! -z "${DATABASE_URL}" ]; then
         bin/cake migrations seed -p BEdita/Core --seed AdminFromEnvSeed
     fi
 
+    chmod a+rwx /var/www/html/webroot/_files
```
fquffio: I think it'd be better to use `a+rwX`, since that sets the execution bit only on directories — and we don't want regular files uploaded by who-knows-who to be executable!

From man pages:
```
           x       The execute/search bits.
           X       The execute/search bits if the file is a directory or any of the execute/search bits are set in the original (unmodified) mode.  Operations with the perm symbol ``X''
                   are only meaningful in conjunction with the op symbol ``+'', and are ignored in all other cases.
```

stefanorosanelli:
@fquffio can we merge this PR? Seems ok to me

PR #1566, opened 2018-11-22
#1531 fix update children position

### LINKED ISSUE #1531: Updating child position still has some issue
### Expected behavior

changing all children's position at once should work

### Actual behavior

the change it's not applied to all child (maybe none) but no error is thrown

1. go into a folder view
2. add some children to it if none is attached
3. try to change all children's position at once

### Environment information

be4-web@issue/155-move-children-object
bedita4@master

batopa:
This PR fixes #1531 

There were two issues that prevent to save correctly the children position.

The first one was that saving tree nodes in a loop the actual left and right position was not used. The previous value in the entity was used instead.
The second one was that the data in the payload had to be ordered to reach the expected behavior.

The children data of the request is now sort following this schema

1. `null`
2. positive integer desc
3. negative integer desc

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Controller/FoldersControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/FoldersControllerTest.php
@@ -1145,7 +1145,7 @@ class FoldersControllerTest extends IntegrationTestCase
      * @param array $objData The data for object to create as child and move.
      * @param array $orphansToAdd The data for object to add as new children.
      * @return void
-     * @return void
+     *
      * @dataProvider moveChildrenProvider
      * @covers ::relationships()
      * @covers ::getDataSortedByPosition()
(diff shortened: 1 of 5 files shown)


plugins/BEdita/API/tests/TestCase/Controller/FoldersControllerTest.php
```diff
@@ -1026,4 +1029,203 @@ public function testUpdateChildPosition()
...
+    }
+
+    /**
+     * Test moving children position.
+     *
+     * @param array $expected The expected children order.
+     * @param array $objData The data for object to create as child and move.
+     * @param array $orphansToAdd The data for object to add as new children.
+     * @return void
+     * @return void
```
didoda: you can remove this line

didoda (approved):

fquffio (approved):
😱 👏

stefanorosanelli (approved):

PR #1567, opened 2018-11-26
custom types from config for multimedia module

didoda:
This PR provides the possibility to customize the types to handle in multimedia module, using a configuration.
For instance:
```
// default: 'application', 'audio', 'b_e_file', 'image', 'video'
$config['multimedia'] = array(
    'types' => array('audio', 'image', 'video'),
);
```

### DIFF
--- a/bedita-app/controllers/modules/multimedia_controller.php
+++ b/bedita-app/controllers/modules/multimedia_controller.php
@@ -3,7 +3,7 @@
  * 
  * BEdita - a semantic content management framework
  * 
- * Copyright 2008-2016 ChannelWeb Srl, Chialab Srl
+ * Copyright 2018 ChannelWeb Srl, Chialab Srl
  * 
  * This file is part of BEdita: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published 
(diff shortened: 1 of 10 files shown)


stefanorosanelli:
This information should be read from API using something like:
```

File: bedita-app/controllers/modules/multimedia_controller.php [hunk 1]
```diff
@@ -20,85 +20,110 @@
...
+        parent::beditaBeforeFilter();
+        $objectTypes = Configure::read('multimedia.types');
+        if ($objectTypes) {
+            $this->object_types = $objectTypes;
+        }
+        foreach ($this->object_types as $type) {
+            $models[] = Inflector::camelize($type);
+        }
+        $mime = Configure::read('validate_resource.mime');
+        foreach ($mime as $modelMyme => $data) {
```
fquffio: [minor] Typo: `modelMime`. 😄

fquffio (approved):
What if we access a disabled media type directly?

Imagine we have disabled video and object 123 is a video: `/multimedia/view/123` would result in an error?

didoda:
Good point, @fquffio We should handle `view/:not_allowed_item_id` case

fquffio (approved):

bedita-app/controllers/modules/multimedia_controller.php
```diff
@@ -20,85 +20,117 @@
...
+     * @var array
+     */
+    var $uses = array('BEObject', 'Application', 'Stream', 'Image', 'Audio', 'Video', 'Tree', 'User', 'Group', 'Category', 'BEFile');
+
+    /**
+     * The object types for module multimedia. Can be customized in configure 'multimedia.types'
+     * 
+     * @var array
+     */
+    protected $object_types = array('application', 'audio', 'b_e_file', 'image', 'video');
```
batopa: should be `$objectTypes`

batopa (changes requested):

batopa (approved):

PR #1569, opened 2018-12-05
Captions behavior for audio and video

didoda:
This PR provides captions for audio and video, using a captions behavior

### DIFF
--- a/bedita-app/models/behaviors/captions.php
+++ b/bedita-app/models/behaviors/captions.php
@@ -1,0 +1,136 @@
+<?php
+/*-----8<--------------------------------------------------------------------
+ * 
+ * BEdita - a semantic content management framework
+ * 
+ * Copyright 2018 ChannelWeb Srl, Chialab Srl
+ * 
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published 
+ * by the Free Software Foundation, either version 3 of the License, or 
+ * (at your option) any later version.
+ * BEdita is distributed WITHOUT ANY WARRANTY; without even the implied 
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License 
+ * version 3 along with BEdita (see LICENSE.LGPL).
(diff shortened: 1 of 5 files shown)


fquffio (approved):

fquffio:
Some tests are failing due to outdated assertions on models' behaviours. Overall, 👍

bedita-app/models/behaviors/captions.php
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Load captions in model data.
+     *
+     * @param object $model The model
+     * @param array $results Fetched models data
+     * @param bool $primary Is primary
+     * @return array
+     */
+    public function afterFind(&$model, $results, $primary)
```
fquffio: [minor] The `&` notation when parameter's value is an object is no longer needed as of PHP 5 — it's a leftover from PHP 4. Objects are always passed by reference.

bedita-app/models/behaviors/captions.php
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Compact find result
+ * 
+ * in model var actAs = array("CompactResult" => array('model to esclude'));
+ *
+ */
+class CaptionsBehavior extends ModelBehavior {
+
+    var $config = array();
```
fquffio: [minor] `var` is the PHP 4 way to declare class members. Better:

```php
public $config = array();
```

(Or `protected` since `$config` is unlikely to be accessed outside of this class… and might probably be removed at all.)

bedita-app/models/behaviors/captions.php
```diff
@@ -0,0 +1,136 @@
...
+ * version 3 along with BEdita (see LICENSE.LGPL).
+ * If not, see <http://gnu.org/licenses/lgpl-3.0.html>.
+ * 
+ *------------------------------------------------------------------->8-----
+ */
+
+/**
+ * Compact find result
+ * 
+ * in model var actAs = array("CompactResult" => array('model to esclude'));
```
fquffio: [minor] Comment refers to `CompactResultBehavior`.

fquffio (approved):

bedita-app/models/behaviors/captions.php
didoda: right. I got the signature from cakephp cookbook for version 1.3

PR #1570, opened 2018-12-07
feat: Allow to append HTML before/after filters

batopa:
This PR introduce an option in `filters` element to append HTML before and/or after the default filters.

### DIFF
--- a/bedita-app/views/elements/filters.tpl
+++ b/bedita-app/views/elements/filters.tpl
@@ -29,7 +29,10 @@
     'customProp' => ['showObjectTypes' => false],
     'categories' => true,
     'mediaTypes' => false,
-    'url' => $html->url("/")|cat:$view->params.controller|cat:'/'|cat:$view->params.action,
+    'url' => $html->url([
+        'controller' => $view->params.controller,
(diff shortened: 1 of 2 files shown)


bedita-app/views/elements/filters.tpl
```diff
@@ -12,29 +12,35 @@
...
+    'word' => true,
+    'tree' => true,
+    'treeDescendants' => true,
+    'relations' => true,
+    'type' => false,
+    'language' => true,
+    'customProp' => ['showObjectTypes' => false],
+    'categories' => true,
+    'mediaTypes' => false,
+    'url' => $html->url("/")|cat:$view->params.controller|cat:'/'|cat:$view->params.action,
```
fquffio: [very minor] Maybe `$html->url(['controller' => $view->params.controller, 'action' => $view->params.action])` would work too.

fquffio (approved):

bedita-app/views/elements/filters.tpl
batopa: Yes it shoud. It was an old thing... Fix is on the way

PR #1572, opened 2018-12-18
Make `params` optional when updating relations

### LINKED ISSUE #1532: need to save relation priority without params
When i save/update a relation priority i get a validation error on the property params of relation->meta, it cannot be either empty or null and it needs to be present.

this behavior prevent to save only the updated priority of the relation. 

endpoint: {{be4Url}}/documents/2/relationships/modern_voter
body: 
```
{
  "data": {
    "id":"48",
    "type":"profiles",
    "meta":{
       "relation":{
         "priority":1
       }
     }
  }
}

```

fquffio:
This PR fixes #1532 by allowing updates to associations without forcing clients to pass _all_ the required fields in the payload, as long as the link was already present and those fields were previously set.

Due to how CakePHP and us handle links between entities, a significant refactor was required. Most of the logic present in the `*AssociatedAction`s and `*RelatedObjectsAction`s has been extracted into an `AssociatedTrait` that is now being used by all the aforehand mentioned Actions. Said trait is abstract enough to handle all current use cases, but it can be easily extended in the future.

The `UpdateRelatedObjectsAction` and its descendants (`AddRelatedObjectsAction`, `SetRelatedObjectsAction`, `RemoveRelatedObjectsAction`) are now much thinner and have a lot less custom logic. We may want to evaluate whether they can be dismissed in the future. However, this refactor has not been done in this PR.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/AssociatedTrait.php
+++ b/plugins/BEdita/Core/src/Model/Action/AssociatedTrait.php
@@ -266,7 +266,7 @@ trait AssociatedTrait
      */
     protected function diff(EntityInterface $source, array $targetEntities, $replace, &$affected = [])
     {
-        $existing = $this->existing($source);
+        $existing = (array)$this->existing($source);
         $kept = $this->intersection($existing, $targetEntities);
 
         $added = array_map(
--- a/plugins/BEdita/Core/src/Model/Action/AssociatedTrait.php
+++ b/plugins/BEdita/Core/src/Model/Action/AssociatedTrait.php
@@ -1,0 +1,302 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2018 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
+namespace BEdita\Core\Model\Action;
+
+use Cake\Datasource\EntityInterface;
+use Cake\Network\Exception\BadRequestException;
+use Cake\ORM\Association\BelongsToMany;
+
+/**
+ * Trait to help with operations on and with associated entities.
+ *
+ * @since 4.0.0
+ *
+ * @property-read \Cake\ORM\Association\BelongsToMany|\Cake\ORM\Association\HasMany $Association
+ */
+trait AssociatedTrait
+{
+
+    /**
+     * Find entity among list of entities.
+     *
+     * @param \Cake\Datasource\EntityInterface $needle Entity being searched.
+     * @param \Cake\Datasource\EntityInterface[] $haystack List of entities.
+     * @return \Cake\Datasource\EntityInterface|null
+     */
+    protected function findMatchingEntity(EntityInterface $needle, array $haystack)
+    {
+        $bindingKey = (array)$this->Association->getBindingKey();
+        foreach ($haystack as $candidate) {
+            $found = array_reduce(
+                $bindingKey,
+                function ($found, $field) use ($candidate, $needle) {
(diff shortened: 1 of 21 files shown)


plugins/BEdita/Core/src/Model/Action/AssociatedTrait.php
```diff
@@ -0,0 +1,302 @@
...
+     *
+     * @param \Cake\Datasource\EntityInterface $source Source entity.
+     * @param \Cake\Datasource\EntityInterface[] $targetEntities Target entities.
+     * @param bool $replace Is this a full-replacement operation?
+     * @param \Cake\Datasource\EntityInterface[] $affected Entities affected by this operation.
+     * @return \Cake\Datasource\EntityInterface[]
+     */
+    protected function diff(EntityInterface $source, array $targetEntities, $replace, &$affected = [])
+    {
+        $existing = $this->existing($source);
```
stefanorosanelli: add `(array)` cast as suggested by scrutinizer?

plugins/BEdita/Core/src/Model/Action/RemoveAssociatedAction.php
```diff
@@ -70,7 +47,7 @@ protected function update(EntityInterface $entity, $relatedEntities)
             }
 
             return $this->Association->getConnection()->transactional(function () use ($entity, $relatedEntities) {
-                $relatedEntities = $this->diff($entity, $relatedEntities);
+                $relatedEntities = $this->intersection($this->existing($entity), $relatedEntities);
```
stefanorosanelli: same as above, add `(array)` cast?

plugins/BEdita/Core/src/Model/Action/SetRelatedObjectsAction.php
```diff
@@ -95,10 +39,14 @@ protected function update(EntityInterface $entity, $relatedEntities)
...
-        $relatedEntities = $this->prepareRelatedEntities($relatedEntities, $entity);
+        $relatedEntities = $this->diff($entity, $relatedEntities, true, $affectedEntities);
 
-        $diff = $this->diff($entity, $relatedEntities);
+        if (!$this->Association->replaceLinks($entity, $relatedEntities)) {
+            return false;
+        }
 
-        return $this->Association->replaceLinks($entity, $relatedEntities) ? $diff : false;
+        return collection($affectedEntities)
```
stefanorosanelli, line 48: parent abstract method in `UpdateAssociatedAction` has return type `@return int|false` here we have `array`, which one is correct?

stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Action/SetRelatedObjectsAction.php
fquffio: Both are correct 😄 that also means none of them actually is correct. 😱

The return type for `UpdateAssociatedAction` should be `int|array|false`.

plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
```diff
@@ -74,23 +42,27 @@ public function execute(array $data = [])
...
-        if (!($entity instanceof Folder) || $this->Association->getName() !== 'Parents') {
-            return $entity;
+        if (empty($data['entity']) || !($data['entity'] instanceof Folder) || $this->Association->getName() !== 'Parents') {
+            return $data;
         }
 
         $table = $this->Association->junction();
         $entity = $table->find()
-            ->where([$table->association('Objects')->getForeignKey() => $entity->id])
+            ->where([$table->association('Objects')->getForeignKey() => $data['entity']->id])
```
batopa, line 56: Do we need an additional check on `$data['entity']` to verify that it's an `Entity` or it is unnecessary?

fquffio: If `!($data['entity'] instanceof Folder)` this method returns `$data` as is (see rows 50-52).

batopa: ok, I missed it ;)

batopa (approved):

PR #1576, opened 2019-01-04
#1575 update tree paths of section's children

### LINKED ISSUE #1575: Lost children path information updating parent for section using Tree::updateTree()
Trying to change parent for a section via API call

```http
POST /objects

{
	"data": {
		"id": 123,
		"object_type": "section",
		"title": "section title",
		"parents": [456]
	}
}
```

update rightly the relative section row on tree but it leaves unchanged the section's children `parent_path` and `object_path`.

batopa:
This PR fixes #1575 

Now `Tree::updateTree()` method can be used consistently with every object type. Before this PR using that method with `section` object led to corrupted `object_path` and `parent_path` in section's children.

### DIFF
--- a/bedita-app/models/tree.php
+++ b/bedita-app/models/tree.php
@@ -263,11 +263,11 @@ class Tree extends BEAppModel
         }
 
         $objectTypeId = ClassRegistry::init('BEObject')->field('object_type_id', array('BEObject.id' => $id));
-        if ($objectTypeId === Configure::read('objectTypes.area.id')) {
+        if ($objectTypeId == Configure::read('objectTypes.area.id')) {
             throw new BeditaBadRequestException(__('Publication cannot be moved to destination', true));
         }
 
-        $isSection = $objectTypeId === Configure::read('objectTypes.section.id');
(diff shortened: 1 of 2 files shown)


bedita-app/models/tree.php
```diff
@@ -242,39 +242,70 @@ public function getParents($id = null, $area_id=null, $status = array(), $useCac
...
-				}
-			}
-		}
+
+        $objectTypeId = ClassRegistry::init('BEObject')->field('object_type_id', array('BEObject.id' => $id));
+        if ($objectTypeId === Configure::read('objectTypes.area.id')) {
+            throw new BeditaBadRequestException(__('Publication cannot be moved to destination', true));
+        }
+
+        $isSection = $objectTypeId === Configure::read('objectTypes.section.id');
```
fquffio: [minor] Is strict comparison (`===`) safe?

fquffio (approved):

bedita-app/models/tree.php
batopa: :scream: good point! It isn't... I fix it.

PR #1579, opened 2019-01-10
Permanent user removal and anonymization #1556

### LINKED ISSUE #1556: Permanently remove users
Permanent removal of users may fail for different reasons:

-  other objects with `created_by` / `modified_by` linked to user id
- auto-created users (from signup for instance) having `created_by` equal to user id

Proposale: in those cases all users fields must be emptied and user objects records not removed from tables but no more visible in trashcan. 
Could this be achieved with a combination of `deleted=1` and `locked=1`??

stefanorosanelli:
This PR fixes #1556 

 * objects having `locked == true` and `deleted == true` are now ghost objects, invisible even in trashcan, a new filter in `GetObjectAction` was introduced for that reason
 * soft-deleted objects will have `locked == false` as default, see `DeleteObjectAction`
 * `UsersTable` delete method overridden: if there are no existing `created_by` or `modified_by` constraints user records are physically removed, otherwise user is anonymized via `anonymizeUser` method
* anonymization works like this:
  -  `username` and `uname` are set to a fixed value `__deleted-{id}` 
  - all other not nullable fields are preserved
  - all remaining fields are set to NULL
* check that `username` and `uname` must not start with reserved word `__deleted-{id}`, `username` value may be checked upon validation, instead `uname` not, both are now checked in `beforeSave` for simplicity 

We have to decide what `locked` means in other scenarios: 
 * `locked` objects are not removable (soft or hard) - this for sure
 * some properties may not be changed (like `status` or `uname` ??) - this is more tricky


plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -295,6 +302,73 @@ public function beforeDelete(Event $event, EntityInterface $entity)
...
+        $this->saveOrFail($entity);
+    }
+
+    /**
+     * Retrieve not nullable columns for a table
+     *
+     * @param \Cake\ORM\Table $table Table class
+     * @return array
+     */
+    protected function notNullableColumns($table)
```
fquffio, line 337: [minor] This method may be static. Since it's a helper function, I personally would feel it would be slightly more explicit if it was `protected static`, but of course this is a matter of personal taste.

plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -295,6 +302,73 @@ public function beforeDelete(Event $event, EntityInterface $entity)
...
+        foreach ($properties as $name) {
+            $entity->set($name, null);
+        }
+        $deletedValue = self::DELETED_USER_PREFIX . $entity->get('id');
+        $entity->set('username', $deletedValue);
+        $entity->set('uname', $deletedValue);
+        $entity->set('password', null);
+        $entity->set('deleted', true);
+        $entity->set('locked', true);
+        $this->saveOrFail($entity);
```
fquffio: I would rather `return $this->save($entity);` here… (see below)

plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -295,6 +302,73 @@ public function beforeDelete(Event $event, EntityInterface $entity)
...
+    {
+        $exists = $this->exists([
+            'OR' => ['created_by' => $entity->get('id'), 'modified_by' => $entity->get('id')],
+        ]);
+        if (!$exists) {
+            return parent::delete($entity, $options);
+        }
+
+        $this->beforeDelete(new Event('Model.beforeDelete'), $entity);
+        $this->anonymizeUser($entity);
```
fquffio: (continues from above) …because here we could `return $this->anonymizeUser($entity);`.

This would make `delete()` behaviour more consistent to the original one. One may still use `$this->deleteOrFail()` in case they want an exception to be raised upon failure.

fquffio (changes requested):

plugins/BEdita/Core/src/Model/Table/UsersTable.php
batopa: :thinking: maybe as helper function it should be placed in another class/trait and not in `UsersTable` class. So I agree for `static` if we move it in a more meaningful place else I think we can leave it as no static method.

plugins/BEdita/Core/src/Model/Table/UsersTable.php
batopa: I agree it is so much better.

fquffio (approved):

PR #1580, opened 2019-01-14
Signup new configuration options

stefanorosanelli:
This PR adds new signup configurations and use cases

 * new configuration keys are introduced:
   - `'requireEmail'` - require email upon signup (default: true)
   - `'requirePassword'` - require password upon signup (default: true)
   - `'activationUrl'` - default activation url, used if not passed by client app
   - `'defaultRoles'` - roles to add upon signup as default if no roles are passed (MUST be listed in `roles`)
 * in `PATCH /auth/user` it is now possibile to update a user's email if NULL

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
@@ -443,7 +443,7 @@ class LoginControllerTest extends IntegrationTestCase
         $data = [
             'name' => 'Gustavo',
             'surname' => 'Trump',
-            'email' => 'gustavo@trump.com'
+            'email' => 'gustavotrump@example.com',
         ];
         $this->configRequest(compact('headers'));
         $this->patch('/auth/user', json_encode($data));
--- a/plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
@@ -427,6 +427,12 @@ class LoginControllerTest extends IntegrationTestCase
      */
     public function testUpdate(array $meta)
     {
+        // set email to NULL to be able to change it
+        $table = TableRegistry::get('Users');
+        $user = $table->get(1);
+        $user->set('email', null);
+        $table->saveOrFail($user);
+
         $headers = [
             'Host' => 'api.example.com',
             'Accept' => 'application/vnd.api+json',
@@ -437,6 +443,7 @@ class LoginControllerTest extends IntegrationTestCase
         $data = [
             'name' => 'Gustavo',
             'surname' => 'Trump',
+            'email' => 'gustavo@trump.com'
         ];
         $this->configRequest(compact('headers'));
         $this->patch('/auth/user', json_encode($data));
@@ -448,6 +455,7 @@ class LoginControllerTest extends IntegrationTestCase
         static::assertEquals(1, $result['data']['id']);
         static::assertEquals($data['name'], $result['data']['attributes']['name']);
         static::assertEquals($data['surname'], $result['data']['attributes']['surname']);
+        static::assertEquals($data['email'], $result['data']['attributes']['email']);
     }
 
     /**
@@ -472,6 +480,7 @@ class LoginControllerTest extends IntegrationTestCase
 
         $data = [
             'username' => 'gustavo',
+            'email' => 'another@email.com',
         ];
 
         $this->configRequest(compact('headers'));
@@ -483,6 +492,7 @@ class LoginControllerTest extends IntegrationTestCase
         static::assertNotEmpty($result['data']);
         static::assertEquals(1, $result['data']['id']);
         static::assertNotEquals($data['username'], $result['data']['attributes']['username']);
+        static::assertNotEquals($data['email'], $result['data']['attributes']['email']);
     }
 
     /**
--- a/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
@@ -28,6 +28,7 @@ use Cake\Mailer\MailerAwareTrait;
 use Cake\Network\Exception\BadRequestException;
 use Cake\Network\Exception\UnauthorizedException;
 use Cake\ORM\TableRegistry;
+use Cake\Utility\Hash;
 use Cake\Validation\Validator;
 
 /**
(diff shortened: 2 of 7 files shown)


plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -303,10 +308,14 @@ protected function getOAuth2Response($url, $accessToken)
      */
     protected function addRoles(User $entity, array $data)
     {
-        if (empty($data['roles'])) {
+        $signupRoles = (array)Configure::read('Signup.defaultRoles');
```
batopa: [minor] why not the more concise
```php
$signupRoles = Hash::get($data, 'roles', Configure::read('Signup.defaultRoles'));
```

batopa (approved):
just a minor comment, not really important

plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
```diff
@@ -437,6 +443,7 @@ public function testUpdate(array $meta)
         $data = [
             'name' => 'Gustavo',
             'surname' => 'Trump',
+            'email' => 'gustavo@trump.com'
```
fquffio: 😂

I'd prefer a [IANA reserved domain](https://www.iana.org/domains/reserved), though.

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -91,6 +92,10 @@ protected function initialize(array $config)
     public function execute(array $data = [])
     {
         $data = $this->normalizeInput($data);
+        // add activation url from config if not set
+        if (empty($data['data']['activation_url'])) {
+            $data['data']['activation_url'] = Configure::read('Signup.activationUrl');
```
fquffio: It'd be nice to pass this URL through `Router::url()`, so that it can be an array, e.g.:

```php
[
  'controller' => 'Foo',
  'action' => 'myActivationAction',
]
```

plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -122,11 +122,17 @@ public function validationSignup(Validator $validator)
...
-            ->requirePresence('password_hash')
-            ->notEmpty('password_hash');
+        if (Configure::read('Signup.requirePassword', true)) {
+            $validator
+                ->requirePresence('password_hash')
+                ->notEmpty('password_hash');
+        }
+
+        if (Configure::read('Signup.requireEmail', true)) {
+            $validator->requirePresence('email');
```
fquffio: (I'm leaving this comment here, but it applies to several lines)

Using this approach, the configuration key is read at the time the validator is built. This is probably fine in most cases, but makes it impossible to change the configuration at runtime in a reliable and predictable way.

I'd prefer another approach that evaluates the requirements at the time the data is being validated:

```php
$validator->requirePresence('email', function () {
    return Configure::read('Signup.requireEmail', true);
});
```

plugins/BEdita/Core/tests/TestCase/Model/Table/UsersTableTest.php
```diff
@@ -422,8 +437,12 @@ public function validationSignupProvider()
      * @covers ::validationSignup()
      * @dataProvider validationSignupProvider
      */
-    public function testValidationSignup($expected, array $data)
+    public function testValidationSignup($expected, array $data, array $config = [])
     {
+        Configure::write('Signup', []);
+        if ($config) {
+            Configure::write('Signup', $config);
+        }
```
fquffio: No need for this `if` construct: `Configure::write('Signup', $config);` is enough thanks to the parameter's default value. 🙃

plugins/BEdita/Core/tests/TestCase/Model/Table/UsersTableTest.php
```diff
@@ -422,8 +437,12 @@ public function validationSignupProvider()
      * @covers ::validationSignup()
      * @dataProvider validationSignupProvider
      */
-    public function testValidationSignup($expected, array $data)
+    public function testValidationSignup($expected, array $data, array $config = [])
```
fquffio, line 441: PHP doc missing?

plugins/BEdita/Core/tests/TestCase/Model/Action/SignupUserActionTest.php
```diff
@@ -397,6 +397,63 @@ public function testExceptionSendMail()
...
+        $data = [
+            'data' => [
+                'username' => 'testsignup',
+                'password' => 'testsignup',
+                'email' => 'test.signup@example.com',
+            ],
+        ];
+
+        EventManager::instance()->on('Auth.signup', function (...$arguments) {
+            static::assertCount(4, $arguments);
```
fquffio, line 420: In case this event listener is never dispatched, the test is still green. I'd add a rough check:

```php
$invoked = 0;
EventManager::instance()->on('Auth.signup', function (...$arguments) use (&$invoked) {
    $invoked++;
    // ...
});

$action = new SignupUserAction();
$action($data);
static::assertSame(1, $invoked);
```

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -161,6 +166,7 @@ protected function validate(array $data)
                 ->notEmpty('username');
         } else {
             $validator
+                ->allowEmpty('activation_url')
```
fquffio: Maybe only when `Signup.activationUrl` is empty? 🤔 

```php
$validator
    ->allowEmpty('activation_url', function () {
        return !Configure::check('Signup.activationUrl');
    })
    // ...
```

fquffio (changes requested):

plugins/BEdita/API/tests/TestCase/Controller/LoginControllerTest.php
stefanorosanelli: :smiley_cat: better indeed...

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
stefanorosanelli: Could be useful, but usually this URL is served via another webapp... we can add it anyway

fquffio: `Router::url('https://example.org/whatev.er') === 'https://example.org/whatev.er'`, so it should be alright.

stefanorosanelli: yes, of course...

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
stefanorosanelli: Not really. In this case we are using an external auth provider, so we don't usually need an activation URL.
But we can write assignement better in `execute()`

fquffio (approved):

PR #1595, opened 2019-02-27
feat: streams removeOrphans shell

### LINKED ISSUE #1593: Shell script to remove orphan streams
Shell script is needed to remove streams (and related files) if:
* `object_id` is NULL
* `created` is older than 1 day - with option to change period

Something like ```bin/cake streams removeOrphans```

didoda:
This PR fixes #1593 
Streams shell, to remove orphans.
Usage:

```
bin/cake streams removeOrphans --days 10 // streams older than 10 days
bin/cake streams removeOrphans // streams older than 1 day
```
WIP: unit tests

### DIFF
--- a/plugins/BEdita/Core/src/Shell/StreamsShell.php
+++ b/plugins/BEdita/Core/src/Shell/StreamsShell.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2017 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+namespace BEdita\Core\Shell;
+
+use BEdita\Core\Model\Entity\Stream;
+use Cake\I18n\Date;
+use Cake\Console\Shell;
+use Cake\ORM\TableRegistry;
+
+/**
+ * Stream shell commands: removeOrphans
+ *
+ * @since 4.0.0
+ */
+class StreamsShell extends Shell
+{
+
+    /**
+     * {@inheritDoc}
+     *
+     * @codeCoverageIgnore
+     */
+    public function getOptionParser()
+    {
+        $parser = parent::getOptionParser();
+        $parser->addSubcommand('removeOrphans', [
+            'help' => 'remove obsolete/orphans streams and related files',
+            'parser' => [
+                'description' => [
+                    'Remove orphans streams.'
+                ],
+                'options' => [
+                    'days' => [
+                        'help' => 'Days to consider for stream research for orphans',
+                        'required' => false,
+                        'default' => 1,
+                    ],
+                ],
+            ]
+        ]);
+
+        return $parser;
+    }
+
+    /**
+     * Create a new Stream
+     *
+     * @return void
+     */
+    public function removeOrphans()
+    {
+        $days = (int)$this->param('days');
+        $streamsTable = TableRegistry::get('Streams');
+        $query = $streamsTable
+            ->find()
+            ->where([
+                'object_id IS NULL',
+                'created >=' => new \DateTime(sprintf('-%d days', $days)),
+            ]);
+        $n = $query->count();
+        if ($n > 0) {
+            $query = $streamsTable->query();
+            $query->delete()
+                ->where([
+                    'object_id IS NULL',
+                    'created >=' => new \DateTime(sprintf('-%d days', $days)),
+                ])
+                ->execute();
+        }
+        $this->out(sprintf('%d stream(s) deleted', $n));
+    }
+}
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        $days = (int)$this->param('days');
+        $streamsTable = TableRegistry::get('Streams');
+        $query = $streamsTable
+            ->find()
+            ->where([
+                'object_id IS NULL',
+                'created >=' => new \DateTime(sprintf('-%d days', $days)),
+            ]);
+        $n = $query->count();
```
stefanorosanelli: [minor] we could use directly:
```
$n = $streamsTable->find()->....->count();
```

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -0,0 +1,82 @@
...
+                    ],
+                ],
+            ]
+        ]);
+
+        return $parser;
+    }
+
+    /**
+     * Create a new Stream
```
stefanorosanelli: Method comment...

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -0,0 +1,82 @@
...
+        $query = $streamsTable
+            ->find()
+            ->where([
+                'object_id IS NULL',
+                'created >=' => new \DateTime(sprintf('-%d days', $days)),
+            ]);
+        $n = $query->count();
+        if ($n > 0) {
+            $query = $streamsTable->query();
+            $query->delete()
```
stefanorosanelli: could we use `$streamsTable->deleteAll([...])` instead and count the number of the actually removed records?

otherwise we can use `$streamsTable->query()->delete()....`

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -0,0 +1,82 @@
...
+
+        return $parser;
+    }
+
+    /**
+     * Create a new Stream
+     *
+     * @return void
+     */
+    public function removeOrphans()
```
stefanorosanelli, line 65: we need a test case, I know it's not easy but we need also make sure that stream files are actually removed...

stefanorosanelli (changes requested):

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        $days = (int)$this->param('days');
+        $streamsTable = TableRegistry::get('Streams');
+        $query = $streamsTable
+            ->find()
+            ->where([
+                'object_id IS NULL',
+                'created >=' => new \DateTime(sprintf('-%d days', $days)),
```
fquffio: I think it would make much more sense if the comparison was made the other way around: delete streams **older** than _N_ days:
* you wouldn't need to pass a very big integer to dispose of very old streams
* you wouldn't break things for a client that's uploading a stream right now, that would see its stream deleted for no reason

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -0,0 +1,82 @@
...
+            ]);
+        $n = $query->count();
+        if ($n > 0) {
```
fquffio: IMHO this makes no sense. There's absolutely no point in performing an extra query to decide if another query should be performed, given that the second query would affect no rows.

Also, I'm pretty sure that performing a delete query that way (`$this->Streams->query()->delete()` or `$this->Streams->deleteAll()`) would **NOT** invoke `beforeDelete` and `afterDelete` callbacks, and therefore the filesystem entries would not be deleted. This would clean up the database, but would leave the filesystem dirty, and the cleanup would become even harder.

My two cents:

```php
$query = $this->Streams->find()
    ->where([
        // ...
    ]);

$count = 0;
foreach ($query as $stream) { // No need to `if`: if the result set is empty, this part is not executed.
    $this->verbose(sprintf('Deleting stream %s...', $stream->id));
    $this->Streams->deleteOrFail($stream);
    $count++;
}
$this->out(sprintf('%d stream(s) deleted', $count));
```

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Create a new Stream
+     *
+     * @return void
+     */
+    public function removeOrphans()
+    {
+        $days = (int)$this->param('days');
+        $streamsTable = TableRegistry::get('Streams');
```
fquffio: Use `$modelClass` instead, and you'll have `$this->Streams` available. See https://github.com/bedita/bedita/blob/4-cactus/plugins/BEdita/Core/src/Shell/JobsShell.php#L31

This is better because on newer versions of CakePHP, `TableRegistry::get()` has been deprecated.

plugins/BEdita/Core/src/Shell/StreamsShell.php
fquffio: Also: I'd prefer to use `Cake\I18n\Time::now()->subDays($days)`, as it's easier to test thanks to `Time::setTestNow()`. Otherwise, testing would be quite a pain.

fquffio (changes requested):
Unit tests are missing.

stefanorosanelli (approved):

fquffio (approved):

PR #1598, opened 2019-03-14
fix: move check and creation of cache dir when resample images

batopa:
This PR fixes an issue checking and creating cache directory for thumbs.

That directory is now checked and created just when a resample is executed and when image info data is put in `.imginfo` dir.

### DIFF
--- a/bedita-app/libs/be_thumb.php
+++ b/bedita-app/libs/be_thumb.php
@@ -315,7 +315,6 @@ class BeThumb {
      * Setup internal imageInfo data array
      * On error $data['error'] is populated with:
      *    - 'notFund' if image is missing or unreachable
-     *    - 'fileSys' on a local filesystem related error
      *    - 'unsupported' if image format is not supported 
      * 
      * @param array $data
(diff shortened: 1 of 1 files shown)


bedita-app/libs/be_thumb.php
```diff
@@ -91,7 +91,14 @@ class BeThumb {
 	 * All known mime types
 	 * internal use (if needed), read from config/mime.types.php
 	 */
-	private $knownMimeTypes = array();
+    private $knownMimeTypes = array();
+    
+    /**
+     * It says if cache dir for thumbnail has been already checked and eventually created.
```
fquffio: [minor] "eventually" means "in the end", it's nothing like Italian's "eventualmente". 🙃 I'd go with "checked and created", simply.

bedita-app/libs/be_thumb.php
```diff
@@ -887,18 +890,28 @@ private function targetFileName() {
...
-	}
+        if ($this->cacheDirChecked === true) {
+            return true;
+        }
+
+        if (file_exists($this->imageInfo['cacheDirectory'])) {
+            if (!is_dir($this->imageInfo['cacheDirectory'])) {
+                $this->triggerError($this->imageInfo['cacheDirectory'] . ' is not a directory');
+            
+                return $this->cacheDirChecked = false;
```
fquffio, line 902: In this case, subsequent calls to `checkCacheDirectory()` would perform the same checks again and again.

How 'bout using `null` as default value, and then `if ($this->cacheDirChecked !== null) { return $this->cacheDirChecked; }` ?
This might cause issues I haven't thought of, so please evaluate this suggestion _very_ carefully. 😱

fquffio (approved):

bedita-app/libs/be_thumb.php
batopa: You are right :+1: :smile_cat:

fquffio (approved):

PR #1604, opened 2019-05-07
[#1603] feat: logged user cannot trash his own user 

### LINKED ISSUE #1603: Trash logged user
A trash request like `DELETE /users/{id}` where `id` is the current logged user should fail.

Instead it is working just fine... :scream_cat: 

A new `BadRequest` or `Forbidden` exception should be rased with message like: `Logged users cannot delete their own account`

didoda:
This PR fixes #1603

### DIFF
--- a/plugins/BEdita/Core/tests/TestCase/Model/Table/UsersTableTest.php
+++ b/plugins/BEdita/Core/tests/TestCase/Model/Table/UsersTableTest.php
@@ -286,6 +286,24 @@ class UsersTableTest extends TestCase
     }
 
     /**
+     * Test soft delete logged user
+     *
+     * @return void
+     *
+     * @expectedException \BEdita\Core\Exception\ImmutableResourceException
+     * @expectedExceptionCode 403
+     * @expectedExceptionMessage Could not self-delete logged user, "User" 5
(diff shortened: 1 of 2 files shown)


stefanorosanelli (approved):

plugins/BEdita/Core/tests/TestCase/Model/Table/UsersTableTest.php
```diff
@@ -285,6 +285,24 @@ public function testSoftDeleteAdminUser()
         $this->Users->save($user);
     }
 
+    /**
+     * Test soft delete logged user
+     *
+     * @return void
+     *
+     * @expectedException \Cake\Network\Exception\BadRequestException
+     * @expectedExceptionCode 403
```
stefanorosanelli: should be `400` see Travis failure...

stefanorosanelli (changes requested):

stefanorosanelli:
Not covered in #1603, but also necessary I guess, same check should be done in
* permanent removal `DELETE /trash/{id}`
* restore `PATCH /trash/{id}`

@didoda @batopa @fquffio ?

stefanorosanelli (approved):

PR #1608, opened 2019-06-03
Avoid potential SQL injection when updating relations with params

fquffio:
This PR fixes a (potentially serious) vulnerability. When saving a relation with parameters, since the encoded JSON was not being properly escaped, it was possible to inject custom SQL by simply adding an apostrophe (`'`) to one relation parameter.

For instance, when linking a Document and an Image together, it would have been possible to give a maliciously-crafted "label" to expose arbitrary data from any table.

This PR fixes this vulnerability by ensuring data is explicitly sanitized before it is used to build queries.

**Note**: #1462 would have been a safer alternative… if it worked! 😄 It turned out to have bad effects since some relations were wiped out and not being updated altogether.

### DIFF
--- a/bedita-app/models/object_relation.php
+++ b/bedita-app/models/object_relation.php
@@ -45,6 +45,31 @@ class ObjectRelation extends BEAppModel
     }
 
     /**
+     * Escape all arguments that have been passed to this function.
+     * 
+     * @param any[] ...$values Values to be prepared.
+     * @return string[]
+     */
+    private static function prepareAll()
+    {
+        $values = func_get_args();
+        foreach ($values as &$val) {
+            if ($val === null) {
+                $val = 'NULL';
+                continue;
+            }
+            
+            if (!is_scalar($val)) {
+                $val = json_encode($val);
+            }
+            $val = sprintf('%s', Sanitize::escape($val));
+        }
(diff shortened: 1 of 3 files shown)


batopa (commented):
:scream: I think we should take a look at `ObjectRelation` model too.

bedita-app/models/object_relation.php
```diff
@@ -44,6 +44,31 @@ public function afterFind($results) {
...
     }
 
+    /**
+     * Escape all arguments that have been passed to this function.
+     * 
+     * @param any[] ...$values Values to be prepared.
+     * @return string[]
+     */
+    private static function prepareAll()
+    {
```
batopa: [minor] in [BE3 coding styles](https://github.com/bedita/bedita/wiki/Resources-(BEdita-3)#php) opening braces for classes and methos MUST be on the same line.

please could you check and fix all methods that don't follow the guidelines.

batopa (changes requested):

bedita-app/models/object_relation.php
fquffio: Done.

didoda (approved):

batopa (approved):

PR #1609, opened 2019-06-17
Require protocol when validating URLs

fquffio:
This PR fixes an issue with fields of type URL: the validator did _not_ require protocol to be present. Thus, `www.example.com` would have been a valid value for a **Profile**'s `website` or an **AuthProvider**'s `url`.

Now, URL fields are required to include a protocol.

Instead of using `Validator::urlWithProtocol()` I've explicitly added the rule for the field. This is required to avoid a breaking change that would have caused some clients to not recognize that type of error anymore: if `Validator::urlWithProtocol` was used, a failure due to an invalid URL would have resulted in a validation error for rule `website.urlWithProtocol`, which differs from the rule `website.url` we have now. Using `Validator::add()` I was able to explicitly set the name of the rule.


plugins/BEdita/Core/tests/TestCase/Model/Entity/ProfileTest.php
```diff
@@ -88,4 +90,51 @@ public function testAccessible()
...
+        ];
+    }
+
+    /**
+     * Test that Website URL is correctly standardized.
+     *
+     * @param mixed $expected Expected result.
+     * @param mixed $website Website value.
+     * @return void
+     *
```
stefanorosanelli: [super minor] remove empty lines between annotations (Cake 4 cs rules)

stefanorosanelli (approved):

plugins/BEdita/Core/tests/TestCase/Model/Entity/ProfileTest.php
fquffio:
```suggestion
```

plugins/BEdita/Core/tests/TestCase/Model/Table/AuthProvidersTableTest.php
```diff
@@ -88,15 +89,17 @@ public function validationProvider()
...
                 [
                     'name' => 'some_unique_value',
                     'url' => 'https://example.com/oauth2',
                     'params' => [],
                 ],
             ],
             'notUnique' => [
-                false,
+                [
+                    'name.unique',
```
stefanorosanelli, line 102: Now we are using `$authProvider = $this->AuthProviders->newEntity([]);` in this test case (Cake 4 compatibility) validation output differs

```suggestion
                    'name._required'
                    'name.unique'
```

plugins/BEdita/Core/tests/TestCase/Model/Table/AuthProvidersTableTest.php
```diff
@@ -88,15 +89,18 @@ public function validationProvider()
...
                     'name' => 'some_unique_value',
                     'url' => 'https://example.com/oauth2',
                     'params' => [],
                 ],
             ],
             'notUnique' => [
-                false,
+                [
+                    'name._required'
+                    'name.unique'
```
fquffio:
```suggestion
                    'name._required',
                    'name.unique',
```

PR #1610, opened 2019-06-19
Add and use mask utility

fquffio:
This PR improves the anonymization of email addresses in logs.

Right now, emails are anonymized by just masking the last three characters. However, due to how an email address is built, the last three characters are the least-significant ones:

- `alice@example.com` would be "masked" as `alice@example.******`
- `bob@example.net` would be "masked" as `bob@example.******`

This PR introduces a `Mask` utility class that helps masking email addresses and make them less- or un-guessable:

- `alice@example.com` is masked as `ali**@example.com`
- `bob@example.com` is masked as `bo*@example.com`

However, the helper library can be configured to mask more characters or to mask the domain part as well (`ali**@***********`).

### DIFF
--- a/bedita-app/app_controller.php
+++ b/bedita-app/app_controller.php
@@ -495,7 +495,7 @@ class AppController extends Controller {
                     $msg .= ': ' . $extUserId;
                 }
                 // avoid userid in logs / use ID instead or partially hide userid
-                $this->eventWarn(sprintf('external login failed: %s******', substr($userid, 0, -3)));
+                $this->eventWarn(sprintf('external login failed: %s', Mask::email($extUserId)));
                 $this->userWarnMessage($msg);
                 $this->Session->delete('externalLoginRequestFailed');
             };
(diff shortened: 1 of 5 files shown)


bedita-app/app_controller.php
```diff
@@ -495,7 +495,7 @@ protected function checkLogin() {
                     $msg .= ': ' . $extUserId;
                 }
                 // avoid userid in logs / use ID instead or partially hide userid
-                $this->eventWarn(sprintf('external login failed: %s******', substr($userid, 0, -3)));
+                $this->eventWarn(sprintf('external login failed: %s', Mask::email($extUserId)));
```
didoda, line 498: I'd expect `Mask::email($userid)` in here... but maybe `$extUserId` already contains the email to mask

didoda (approved):

bedita-app/app_controller.php
fquffio: `$userid` seems undefined in this context… 🤔 `$extUserId` looked like it was the right variable to use, but I may be mistaken.

didoda: I'll take a look on the file...

PR #1607, opened 2019-05-21
Avoid bad side effects on DataCleanup (see #1601)

stefanorosanelli:
This PR fixes #\<issue-number\>

Provide a general description of the code changes in your pull request.

Issue #1601, opened 2019-03-22
DataCleanupBehavior problem with `patchEntity`


stefanorosanelli:
`DataCleanupBehavior` logic needs to be changed.
Performing a simple `patchEntity` on an existing object entity may cause an unwanted `status` change.

Unit test method to add to `DataCleanupBehaviorTest` shows the problem.
`status` is `on` in fixture definition, but becomes `draft` after `->patchEntity`

```php
    public function testPatchDataCleanup()
    {
        Configure::write('DefaultValues', []);
        $Documents = TableRegistry::get('Documents');
        $document = $Documents->get(2);
        $document = $Documents->patchEntity($document, ['title' => 'new title']);

        static::assertEquals('new title', $document->get('title'));
        static::assertEquals('on', $document->get('status'));
    }
```

PR #1612, opened 2019-06-24
Fix `DefaultValues` config on core types

stefanorosanelli:
This PR fixes a problem with `DefaultValues` config on some core object types.

Even if `DataCleanupBehavior` needs a refactor (see #1601 and #1607), `DefaultValues` configuration should perform as expected on core object types.

Currently it's not working on `images`, `files`, `videos`, `audio` and `locations`

A new related integration test has been added.


batopa (approved):

plugins/BEdita/API/tests/IntegrationTest/DefaultValuesTest.php
```diff
@@ -0,0 +1,119 @@
...
+     * @param array $expected Expected result.
+     * @param string $type Object type.
+     * @param array $attributes Object attributes.
+     * @param array $config DefaultValues config.
+     * @return void
+     *
+     * @dataProvider createProvider
+     * @coversNothing
+     */
+    public function testCreate(array $expected, string $type, array $attributes, array $config)
```
fquffio: [minor] Since we’re starting using PHP 7 parameter type hinting (e.g. `string $type`), we could also add return types hinting:

```php
public function createProvider(): array
{
    // ...
}

public function testCreate(/* ... */): void
{
    // ...
}
```

fquffio (approved):

plugins/BEdita/API/tests/IntegrationTest/DefaultValuesTest.php
stefanorosanelli: Done

PR #1613, opened 2019-07-03
Allow `+02` and `+0200` as TZ format in input date

stefanorosanelli:
This fixes a problem with timezone input format.

Valid ISO 8601 timezone formats are:

1. `<time>Z`
1. `<time>±hh:mm`
1. `<time>±hhmm`
1. `<time>±hh`

see https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators

currently only the first 2 formats are accepted as valid input, last 2 are denied as invalid dates but should be accepted instead


plugins/BEdita/Core/src/Model/Validation/Validation.php
```diff
@@ -166,7 +167,7 @@ public static function dateTime($value)
             return true;
         }
 
-        if (is_string($value) && preg_match('/^\d{4}(-\d\d(-\d\d([T ]\d\d:\d\d(:\d\d)?(\.\d+)?(([+-]\d\d:\d\d)|Z)?)?)?)?$/i', $value)) {
+        if (is_string($value) && preg_match('/^\d{4}(-\d\d(-\d\d([T ]\d\d:\d\d(:\d\d)?(\.\d+)?(([+-]\d\d(:)?(\d\d)?)|Z)?)?)?)?$/i', $value)) {
```
fquffio: `:?` would be shorter than `(:)?`

fquffio (approved):
Just a minor syntax suggestion.

batopa (approved):

plugins/BEdita/Core/src/Model/Validation/Validation.php
stefanorosanelli: right

stefanorosanelli: done

PR #1622, opened 2019-08-21
#1621 publisher field null value (avoid empty string)

### LINKED ISSUE #1621: Objects.publisher field is set to empty string when an object is saved from bedita-app
After #1025 in publisher field we should have NULL value, defaultPublisher or string value inserted by user in frontend.

When an object is created by API and updated from bedita-app or when from bedita-app is cleared the value of field, in database is saved empty string and not null value.

When we order object by publisher to give priority on editorial content, objects with publisher = "" and publisher = null is ordered separately.

mikelina:
This PR fixes #1621 

Changed function setPublisher to use null value if publisher field is empty.

Added shell script to update publisher field from empty string  to null value.
Usage: ./cake.sh publisher_user_contents


batopa:
About the shell script I would prefer to use the already present `editorial_contents` launched with a specific command or parameter.

For example:

```bash
# with command
./cake.sh editorial_contents resetEmptyPublisher
# or with parameter
./cake.sh editorial_contents --reset-empty-publisher
```

Since the logic seems very similar if it's possible I would parametrize the code.

bedita-app/vendors/shells/editorial_contents.php
```diff
@@ -38,23 +38,53 @@ class EditorialContentsShell extends BeditaBaseShell
...
-        if (empty($publisher)) {
-            $this->out('=====> Configuration item `editorialContents.defaultPublisher` must not be empty');
-            $this->err('Aborting');
 
-            exit(1);
+        $resetEmptyPublisher = !empty($this->params['-reset-empty-publisher']);
+
+        if ($resetEmptyPublisher) {
+
+            $backendAuth = 0;
```
batopa: Is there a specific reason because the reset should work just for objects created from users belong to frontend groups?

mikelina: You are right, the reset must work for all objects independently from user who created it

batopa (approved):

PR #1625, opened 2019-09-25
Relations and Properties utilities

stefanorosanelli:
This PR adds simple utility classes to be added in migrations, shell scripts and other scenarios to easily handle `relations` and custom `properties` creation.

Migration use case:

 * define a `$relations` array like:
```
    protected $relations = [
        [
            // docs_media (media_of_doc): documents / media
            'name' => 'docs_media',
            'label' => 'Docs media',
            'inverse_name' => 'media_of_docs',
            'inverse_label' => 'Media of Docs',
            'description' => 'docs having media',
            'left' => ['documents'],
            'right' => ['media'],
        ],
    ];
```
Then a simple call like
```
        Relations::create(
            $this->relations,
            ['connection' => $this->getAdapter()->getCakeConnection()]
        );
```
will create the relations using migration transaction

### DIFF
--- a/plugins/BEdita/Core/src/Utility/Properties.php
+++ b/plugins/BEdita/Core/src/Utility/Properties.php
@@ -13,7 +13,9 @@
 
 namespace BEdita\Core\Utility;
 
+use Cake\Network\Exception\BadRequestException;
 use Cake\ORM\TableRegistry;
+use Cake\Utility\Hash;
 use Cake\Utility\Inflector;
 
 /**
@@ -50,10 +52,12 @@ class Properties
         $Properties = TableRegistry::getTableLocator()->get('Properties', $options);
 
         foreach ($properties as $p) {
+            static::validate($p);
             $property = $Properties->newEntity([
                 'name' => $p['name'],
                 'property_type_name' => $p['property'],
                 'object_type_name' => $p['object'],
+                'description' => Hash::get($p, 'description'),
             ]);
 
             $Properties->saveOrFail($property);
@@ -73,6 +77,7 @@ class Properties
         $ObjectTypes = TableRegistry::getTableLocator()->get('ObjectTypes', $options);
 
         foreach ($properties as $p) {
+            static::validate($p);
             $objectType = $ObjectTypes->get(Inflector::camelize($p['object']));
 
             $property = $Properties->find()
@@ -83,6 +88,24 @@ class Properties
                 ->firstOrFail();
 
             $Properties->deleteOrFail($property);
+        }
+    }
+
+    /**
+     * Validate properties before creation or removal
+     *
+     * @param array $data Properties data
+     * @return void
+     * @throws BadRequestException
+     */
+    protected static function validate(array $data) : void
+    {
+        $required = ['name', 'object', 'property'];
+        $diff = array_diff_key(array_flip($required), array_filter($data));
+        if (!empty($diff)) {
+            throw new BadRequestException(
+                __d('bedita', 'Missing mandatory property data "{0}"', implode(array_keys($diff)))
+            );
         }
     }
 }
--- a/plugins/BEdita/Core/src/Utility/Properties.php
+++ b/plugins/BEdita/Core/src/Utility/Properties.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2019 ChannelWeb Srl, Chialab Srl
+ *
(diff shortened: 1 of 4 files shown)


plugins/BEdita/Core/src/Utility/Properties.php
```diff
@@ -0,0 +1,88 @@
...
+     * @param array $properties Properties data
+     * @param array $options Table locator options
+     * @return void
+     */
+    public static function create(array $properties, array $options = []) : void
+    {
+        $Properties = TableRegistry::getTableLocator()->get('Properties', $options);
+
+        foreach ($properties as $p) {
+            $property = $Properties->newEntity([
```
batopa, line 56: I would add a simple validation (on presence) on required fields in `$p`

plugins/BEdita/Core/src/Utility/Properties.php
```diff
@@ -0,0 +1,88 @@
...
+     * @param array $options Table locator options
+     * @return void
+     */
+    public static function remove(array $properties, array $options = []) : void
+    {
+        $Properties = TableRegistry::getTableLocator()->get('Properties', $options);
+        $ObjectTypes = TableRegistry::getTableLocator()->get('ObjectTypes', $options);
+
+        foreach ($properties as $p) {
+            $objectType = $ObjectTypes->get(Inflector::camelize($p['object']));
```
batopa, line 81: Same as above

plugins/BEdita/Core/src/Utility/Relations.php
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * @param array $relations Relation data
+     * @param array $options Table locator options
+     * @return void
+     */
+    public static function create(array $relations, array $options = []) : void
+    {
+        $Relations = TableRegistry::getTableLocator()->get('Relations', $options);
+        foreach ($relations as $data) {
+            $relation = $Relations->newEntity($data);
```
batopa, line 54: I would add a simple validation on required fields. 

In this case maybe it's enough to check `left` and `right` keys since the other data are used directly for hydrating a new entity, so I expected the `save` will fails if some field is not valid fot `RelationsTable`.

plugins/BEdita/Core/src/Utility/Relations.php
```diff
@@ -0,0 +1,135 @@
...
+     * @param array $relations Relation data
+     * @param array $options Table locator options
+     * @return void
+     */
+    public static function remove(array $relations, array $options = []) : void
+    {
+        $Relations = TableRegistry::getTableLocator()->get('Relations', $options);
+        foreach ($relations as $r) {
+            $relation = $Relations->find()
+                ->where(['name' => $r['name']])
```
batopa: We could add validation on expected keys or at least using `Hash::get()` to avoid notice in case of missing keys.
Also below.

plugins/BEdita/Core/src/Utility/Properties.php
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * Provides static methods to create and remove properties using an array format
+ *
+ * Example:
+ *   [
+ *     [
+ *       'name' => 'custom_one',
+ *       'object' => 'documents',
+ *       'property' => 'boolean',
+ *       'description' => 'my custom description', // optional
```
batopa, line 32: `description` is mentioned here but not used in the method

batopa (changes requested):

plugins/BEdita/Core/src/Utility/Properties.php
stefanorosanelli: done

plugins/BEdita/Core/src/Utility/Properties.php
stefanorosanelli: done

plugins/BEdita/Core/src/Utility/Relations.php
stefanorosanelli: agree, done

plugins/BEdita/Core/src/Utility/Relations.php
stefanorosanelli: yes, done

plugins/BEdita/Core/src/Utility/Properties.php
stefanorosanelli: added

batopa (approved):

PR #1631, opened 2019-10-30
Avoid migrations transaction error on SQlite

stefanorosanelli:
This PR fixes a problem on migrations with SQlite.

(Subsitutes previous PR #1630)

Launching migrations on an empy SQlite db we get an error like:
```
 == 20171005105447 ObjectTypesInheritance: migrating
PDOException: There is no active transaction in /var/www/bedita4/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/SQLiteAdapter.php:216
Stack trace:
#0 /var/www/bedita4/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/SQLiteAdapter.php(216): PDO->commit()
#1 /var/www/bedita4/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/AdapterWrapper.php(329): Phinx\Db\Adapter\SQLiteAdapter->commitTransaction()
#2 /var/www/bedita4/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/AdapterWrapper.php(329): Phinx\Db\Adapter\AdapterWrapper->commitTransaction()
#3 /var/www/bedita4/vendor/robmorgan/phinx/src/Phinx/Migration/Manager/Environment.php(133): Phinx\Db\Adapter\AdapterWrapper->commitTransaction()
```

This is caused by a `PDO->commit()` call where a previous `commit()` has already been called from `TreeBehavior::recover()`

Current solution is to override `recover` with an option to avoid `transactional(...)` callback

### DIFF
--- a/plugins/BEdita/Core/config/Migrations/20171005105447_ObjectTypesInheritance.php
+++ b/plugins/BEdita/Core/config/Migrations/20171005105447_ObjectTypesInheritance.php
@@ -143,7 +143,7 @@ class ObjectTypesInheritance extends AbstractMigration
         ]);
         /* @var \BEdita\Core\Model\Behavior\TreeBehavior $tree */
         $tree = $table->behaviors()->get('Tree');
-        $tree->recover(false);
+        $tree->nonAtomicRecover();
     }
 
     /**
--- a/plugins/BEdita/Core/config/Migrations/20171005105447_ObjectTypesInheritance.php
+++ b/plugins/BEdita/Core/config/Migrations/20171005105447_ObjectTypesInheritance.php
@@ -137,13 +137,13 @@ class ObjectTypesInheritance extends AbstractMigration
         );
 
         // Now let's fix NSM (nested-set model) left and right indexes from tree data.
-        $table->addBehavior('Tree', [
+        $table->addBehavior('BEdita/Core.Tree', [
             'left' => 'tree_left',
             'right' => 'tree_right',
         ]);
-        /* @var \Cake\ORM\Behavior\TreeBehavior $tree */
+        /* @var \BEdita\Core\Model\Behavior\TreeBehavior $tree */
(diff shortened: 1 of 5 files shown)


batopa (approved):
Just a note. I see that cakephp migrations has been updated to 2.3.0.
Myabe we could give a try using that version before merge this fix.
What do you think?

fquffio:
> Just a note. I see that cakephp migrations has been updated to 2.3.0.
> Myabe we could give a try using that version before merge this fix.
> What do you think?

I'm afraid it's not gonna help… the problem is more likely to be in TreeBehavior, or more specifically in CakePHP `Connection::transactional()` not realising there's already an open transaction, so `transactional()` should not start a new one (or even worse, issuing `COMMIT`!) unless the datasource has savepoints, which SQLite hasn't.
This, in turn, may be caused by the fact that Phinx doesn't use CakePHP connection abstraction, thus `Connection` doesn't get involved when Phinx starts a transaction.

Running `TreeBehavior::_recoverTree()` in a non-transactional environment is IMHO the simpler solution to go.
However, since that may cause corruption unless a transaction is started by other means, I'd be even more explicit.

Rather than overriding `recover()` with an optional flag, I'd create an entirely new method `public function nonAtomicRecover()` which invokes `parent::_recover()` without surrounding it with a transaction. One would immediately realise how dangerous `$table->nonAtomicRecover()` is when they see it, whereas `$table->recover(false)` would be less "catchy".

fquffio (approved):

stefanorosanelli:
> Rather than overriding `recover()` with an optional flag, I'd create an entirely new method `public function nonAtomicRecover()` which invokes `parent::_recover()` without surrounding it with a

Ok, that was my first idea more or less. I'll go with thie one.

PR #1634, opened 2019-11-14
Fix/v4/1633 timezone db

### LINKED ISSUE #1633: POST and GET of dates with timezone
### Expected behavior

Save dates with timezone as UTC dates.
I.e.: '2018-08-02T16:23:23+02:00' => save into db as '2018-08-02T14:23:23+00:00'

### Actual behavior

Save dates without timezone.
I.e.: '2018-08-02T16:23:23+02:00' => save into db as '2018-08-02T16:23:23+00:00'

Unit test to reproduce the problem:

```php
    public function testDateTimezone()
    {
        $data = [
            'type' => 'documents',
            'id' => '2',
            'attributes' => [
                'publish_start' => '2018-08-02T16:23:23+02:00',
            ],
        ];

        $this->configRequestHeaders('PATCH', $this->getUserAuthHeader());
        $this->patch('/documents/2', json_encode(compact('data')));
        $attributes = Hash::get($result, 'data.attributes', []);
        static::assertEquals('2018-08-02T14:23:23+00:00', $attributes['publish_start']);

        $this->configRequestHeaders('GET', $this->getUserAuthHeader());
        $this->get('/documents/2');

        $attributes = Hash::get($result, 'data.attributes', []);
        static::assertEquals('2018-08-02T14:23:23+00:00', $attributes['publish_start']);
    }
```

stefanorosanelli:
This PR fixes #1633 

Solution not optimal, but it seems that actual `timezone` conversion when saving to DB will be covered in Cake 4.0

See https://github.com/cakephp/cakephp/issues/13646 

Response to `PATCH` request will actually contain date times with same timezone as in request.
We simply return entity data returned from `Table::save()`
Not sure what the expected behavior should be here, but it could be addressed in a separate issue.

For now we make sure `UTC` is used as timezone for date time fields stored on db.


didoda (approved):

plugins/BEdita/Core/config/bootstrap.php
```diff
@@ -60,10 +60,16 @@
...
+ * Timezone conversion is not currently supported in CakePHP 3.8 saving to Database
+ * We should use `timezone` as defined in `Datasource` configuration, with 'UTC' as default
+ * For now is safe to force 'UTC' as timezone
+ *
+ * See https://github.com/cakephp/cakephp/issues/13646
  */
 Type::set('date', new DateType());
-Type::set('datetime', new DateTimeType());
-Type::set('timestamp', new DateTimeType());
+Type::set('datetime', (new DateTimeType())->setTimezone('UTC'));
```
batopa: Maybe we could slightly improve this setting server timezone
```php
Type::set('datetime', (new DateTimeType())->setTimezone(date_default_timezone_get()));
```
but anyway I think this solution is ok for now.

batopa (approved):

plugins/BEdita/Core/config/bootstrap.php
stefanorosanelli: You're right, but the timezone setting should probably come from datasource connection config.

What about a constant like:

```php
if (!defined('BEDITA_DEFAULT_TIMEZONE')) {
    define('BEDITA_DEFAULT_TIMEZONE', 'UTC');
}
```

or better an env var

```php
date_default_timezone_set(env('BEDITA_DEFAULT_TIMEZONE', 'UTC'));
```

Then always use `date_default_timezone_get()` in database types, datasource config etc etc...

PR #1636, opened 2019-11-20
Object history #1624

### LINKED ISSUE #1624: Object history data model
Create data model layer for objects history 
see https://github.com/orgs/bedita/teams/core/discussions/10

stefanorosanelli:
This PR fixes #1624 

A new data model for object history has been introduced following https://github.com/orgs/bedita/teams/core/discussions/10 

### Datasource

Default history datasource is the new `history` table, but other tables and datasources are possible via this configuration

```php
 'History' => [
    'table' => '\MyPlugin\Model\Table\HistoryTable',
```
where a new custom table class may be used to point to another db/table or NoSQL instance.
This custom class has to provide field filters or finders at least for  `resource_id`, `resource_type` and `user_id`.
Pagination must also be supported.

### Model

* a new `HistoryBehavior` has been created using history table class seen above to automatically store history events for all possible actions: `'create', 'update', 'trash', 'restore'` and `'remove'`
* default table class is `History` table with `history` and `activity` finders seen above.

## API 

New **/history** endpoint has been defined providing only `GET` method on history events
Using filters you can retrieve events of single object or user activity

**Object history** => `GET /history?filter[resource_id]={id}&filter[resource_type]=objects` - where `{id}` is the object id

**User activity** => `GET /history?filter[user_id]={id}` -  where `{id}` is the user id


plugins/BEdita/Core/src/Model/Behavior/HistoryBehavior.php
```diff
@@ -0,0 +1,162 @@
...
+     * @param EntityInterface $entity Object entity.
+     * @return EntityInterface
+     */
+    protected function historyEntity(EntityInterface $entity): EntityInterface
+    {
+        /** @var \BEdita\Core\Model\Entity\History $history */
+        $history = $this->Table->newEntity();
+        $history->resource_id = $entity->get('id');
+        $history->resource_type = $this->getConfig('resource_type');
+        $history->application_id = CurrentApplication::getApplicationId();
```
didoda, line 155: scrutinizer suggests a type check. maybe a cast is enough

```
$history->application_id = (int)CurrentApplication::getApplicationId();
```

didoda (approved):

plugins/BEdita/Core/src/Model/Behavior/HistoryBehavior.php
stefanorosanelli: Actually the annotation `CurrentApplication::getApplicationId()` is wrong, should be `int|null` instead it's `Application|null` that's why Scrutinizer complains

Fixed annotation and pushed in this branch

didoda (approved):

plugins/BEdita/Core/tests/TestCase/Model/Behavior/HistoryBehaviorTest.php
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    /**
+     * Test `initialize` method
+     *
+     * @covers ::initialize()
+     */
+    public function testInitialize()
+    {
+        $Documents = TableRegistry::get('Documents');
```
didoda, line 75: Ok with `Documents`. But it seems that with `Users`, `Profiles` and other types, history doesn't work.
We could add one of those types in tests.

stefanorosanelli: Solved now, tests have been updated.
Solution far from optimal, a model table refactoring is necessary. Will be done in a new issue.

batopa:
@stefanorosanelli there are some conflicts to solve.

stefanorosanelli:
> @stefanorosanelli there are some conflicts to solve.

should be ok now, using new `ObjectModelBehavior`

plugins/BEdita/Core/src/Model/Table/HistoryTable.php
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public function initialize(array $config)
+    {
+        parent::initialize($config);
+
+        $this->setTable('history');
+        $this->setDisplayField('id');
+        $this->setPrimaryKey('id');
+
+        $this->addBehavior('Timestamp');
```
batopa: Since the `created` value is always calculated in `HistoryBehavior::historyEntity()` the `Timestamp` behavior doesn't need here.

Maybe a better solution will be:
* remove from here 
* the `HistoryBehavior::initialize()` should add `Timestamp` behavior to `HistoryBehavior::$Table` object
* remove the set of `created` field from `HistoryBehavior::historyEntity()`

plugins/BEdita/Core/src/Model/Behavior/HistoryBehavior.php
```diff
@@ -0,0 +1,194 @@
...
+     */
+    public $Table = null;
+
+    /**
+     * The changed properties.
+     * Array of visible properties changed by user
+     *
+     * @var array
+     */
+    public $changed = [];
```
batopa: I think that define `$changed` as `protected` would be better to avoid direct write on that property. I would expose it with a getter.

batopa (changes requested):

plugins/BEdita/Core/src/Model/Table/HistoryTable.php
stefanorosanelli: Sounds right, done!

plugins/BEdita/Core/src/Model/Behavior/HistoryBehavior.php
stefanorosanelli: Ok, done!

batopa (approved):

PR #1641, opened 2019-12-05
Refactor object table hieararchy and behaviors

stefanorosanelli:
This PR adds 

* a new abstract `ObjectsBaseTable` as base class for tables implementing a BEdita object type
* a new `ObjectModel` behavior to load common behaviors

This way new common behaviors can be introduced without any code change in core and plugin object table classes.

### Caveat

* `ObjectsTable` will not extend `ObjectsBaseTable` but simply load `ObjectModel` behavior.
* any new change in `ObjectsBaseTable` should normally be added to `ObjectsTable`

After this change every Table class implementing a BEdita object on a table other than `objects` MUST 
 * extend `ObjectsBaseTable`
 * invoke `$this->extensionOf('Objects');`  (using `'Objects'` or other objects table class) in `initialize()` 

Have a look at `ProfilesTable` for an example

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/LocationsTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/LocationsTable.php
@@ -13,8 +13,8 @@
 
 namespace BEdita\Core\Model\Table;
 
+use BEdita\Core\Model\Table\ObjectsBaseTable as Table;
 use BEdita\Core\Model\Validation\LocationsValidator;
-use BEdita\Core\ORM\Inheritance\Table;
 
 /**
  * Locations Model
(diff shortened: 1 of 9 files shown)


plugins/BEdita/Core/src/Model/Table/LocationsTable.php
```diff
@@ -51,24 +51,20 @@ public function initialize(array $config)
...
-                'title' => 10,
-                'description' => 7,
-                'body' => 5,
-                'address' => 1,
-                'locality' => 2,
-                'country_name' => 2,
-                'region' => 2,
-            ],
-        ]);
+        if ($this->hasBehavior('Searchable')) {
```
batopa: Maybe not necessary since its parent adds `Searchable` behavior. Isn't it?
In case we could remove the check on behaviors added by `ObjectsBaseTable` in other tables too.

stefanorosanelli: Right, check is not necessary. If for some reasons these behaviors are remove we would have other bad side effects.

batopa (approved):

PR #1649, opened 2020-01-16
Auth user opt-out #1578

### LINKED ISSUE #1578: User Opt-out
API method for a user to opt-out: user will be logged out and permanently deleted as in #1556

stefanorosanelli:
This PR closes #1578 

A new `/auth/optout` endpoint was created in order to definitively be removed as a user from a BE4 Project.

Any user invoking `POST /auth/optout` must pass identification data in request body (via usual methods like classic username/password, OAuth2, OTP, 2FA...). 
If identification is successful:

- user data are completely removed or anonymized - using same logic seen in #1579 
- an `Auth.optout` event is dispatched in order to trigger other actions via plugins

It's up to the client to completely remove access and refresh token from the client app afterwards.

### DIFF
--- a/plugins/BEdita/API/src/Controller/LoginController.php
+++ b/plugins/BEdita/API/src/Controller/LoginController.php
@@ -15,6 +15,7 @@ namespace BEdita\API\Controller;
 
 use BEdita\Core\Model\Action\ChangeCredentialsAction;
 use BEdita\Core\Model\Action\ChangeCredentialsRequestAction;
+use BEdita\Core\Model\Action\GetObjectAction;
 use BEdita\Core\Model\Action\SaveEntityAction;
 use BEdita\Core\Model\Entity\User;
 use Cake\Auth\PasswordHasherFactory;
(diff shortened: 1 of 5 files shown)


plugins/BEdita/API/src/Controller/LoginController.php
```diff
@@ -86,22 +88,81 @@ public function initialize()
...
+        if (!empty($result['authorization_code']) && count($result) === 1) {
+            $meta = ['authorization_code' => $result['authorization_code']];
+        } else {
+            $result = $this->reducedUserData($result);
+            $meta = $this->jwtTokens($result);
+        }
+
+        $this->set('_serialize', []);
+        $this->set('_meta', $meta);
+    }
```
didoda, line 120: minor... function 'paths' could be reduced... i.e.:
```
    public function login(): void
    {
        $this->set('_serialize', []);
        $result = $this->identify();
        // Check if result contains only an authorization code (OTP & 2FA use cases)
        if (!empty($result['authorization_code']) && count($result) === 1) {
            $this->set('_meta', ['authorization_code' => $result['authorization_code']]);
            
            return;
        }
        $result = $this->reducedUserData($result);
        $meta = $this->jwtTokens($result);
        $this->set('_meta', $meta);
    }
```

didoda (approved):

plugins/BEdita/API/src/Controller/LoginController.php
stefanorosanelli: done

PR #1657, opened 2020-02-20
Inverse relation priority

stefanorosanelli:
This PR introduces a more transparent `priority` handling when dealing with inverse relations.

### Problem

When we see related objects or add new ones and we are using an inverse relation we should use `inv_priority` instead of `priority` as desired order.

With an API call like `GET /images/123/media_of` where `media_of` is an inverse relation we see a list objects having in `meta.relation` properties like

```json
 "priority": 2,
 "inv_priority": 1,
 "params": null
```

But we don't know that the real priority used is `inv_priority` and not `priority`.
We should check if this relation name is direct or not.

### Proposed solution

When we list related objects or we set new relations on the API we can always use `priority` to see and set the relation order. 
Internally (via simple switch) we decide to use `inv_priority` in case of inverse relation.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
@@ -95,7 +95,10 @@ class ListRelatedObjectsAction extends ListAssociatedAction
      */
     protected function prepareJoinEntity(EntityInterface $joinData): void
     {
-        if (!$joinData instanceof ObjectRelation || $this->Association->getForeignKey() !== 'right_id') {
+        if (
+            !$joinData instanceof ObjectRelation ||
+            ($this->Association instanceof RelatedTo && !$this->Association->isInverse())
+        ) {
             return;
         }
 
--- a/plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
@@ -13,10 +13,12 @@
 
 namespace BEdita\Core\Model\Action;
 
+use BEdita\Core\Model\Entity\ObjectRelation;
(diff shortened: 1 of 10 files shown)


plugins/BEdita/Core/src/Model/Action/ListRelatedObjectsAction.php
```diff
@@ -83,4 +85,23 @@ protected function buildQuery($primaryKey, array $data, Association $inverseAsso
...
+    /**
+     * Prepare `joinData` entity on ObjectRelation associations.
+     * If relation is inverse switch between `priority`/`inv_priority`.
+     *
+     * @param \Cake\Datasource\EntityInterface $joinData Join data entity.
+     * @return void
+     */
+    protected function prepareJoinEntity(EntityInterface $joinData): void
+    {
+        if (!$joinData instanceof ObjectRelation || $this->Association->getForeignKey() !== 'right_id') {
```
batopa: Maybe we could use `$this->Association instanceof RelatedTo` and add a method as `RelatedTo::isInverse()` that return `$this->Association->getForeignKey() === 'right_id'`

plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
```diff
@@ -65,4 +68,27 @@ protected function prepareData(array $data)
...
+    /**
+     * Setup `priority` on `_joinData`.
+     * If relation is inverse switch `priority` and `inv_priority`.
+     *
+     * @param array $data Action data.
+     * @return void
+     */
+    protected function setupPriority(array &$data): void
+    {
+        if ($this->Association->getForeignKey() !== 'right_id') {
```
batopa: As above we could have a method in `RelatedTo` association to know if it's an inverse relation.

PR #1659, opened 2020-02-21
Max + 1 as default priority

### LINKED ISSUE #1654: Default priority on relations: max + 1
On object relation creation if no `priority` is set use current max priority +1 as default value, on both directions also for inverse relation.

This will add am initial natural ordering for related objects that can be of course ignored.

We could then add an ordering property like `priority_order` to determine if `priority` should be used in asc or desc order when reading related objects.

stefanorosanelli:
This PR fixes #1654 

`PriorityBehavior` was modified in order to use current max + 1 as `priority` / `inv_priority` value.
New relations will have 1 and not 0 as starting value.

Bonus track:  `RelationshipsPriority` integration test.


plugins/BEdita/Core/src/Model/Table/ObjectRelationsTable.php
```diff
@@ -3,8 +3,11 @@
 
 use BEdita\Core\Model\Validation\Validation;
 use Cake\Database\Schema\TableSchema;
+use Cake\Datasource\EntityInterface;
```
batopa: unused `use` statements.
See below too.

batopa (changes requested):

plugins/BEdita/Core/src/Model/Table/ObjectRelationsTable.php
stefanorosanelli: done

batopa (approved):

PR #1662, opened 2020-02-27
Allow non assoc config `Plugins`

stefanorosanelli:
This PR provides:

 * refactor con plugins loaded via `Plugin` configuration, default options have been changed and a simpler configuration is now possibile:

```php
'Plugins' => [
    'MyPlugin',
 ],
```
previous associative array version is still valid 

```php
'Plugins' => [
    'MyPlugin' => ['bootstrap' => true, 'ignoreMissing' => true],
 ],
```

 * unit test for `Application` was added and `AssetMiddleware` has been removed

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -18,7 +18,6 @@ use Cake\Core\Configure;
 use Cake\Core\Exception\MissingPluginException;
 use Cake\Error\Middleware\ErrorHandlerMiddleware;
 use Cake\Http\BaseApplication;
-use Cake\Routing\Middleware\AssetMiddleware;
 use Cake\Routing\Middleware\RoutingMiddleware;
 
 /**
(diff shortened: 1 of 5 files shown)


src/Application.php
```diff
@@ -63,27 +75,40 @@ protected function bootstrapCli()
...
-                }
+        $plugins = (array)Configure::read('Plugins');
+        if (empty($plugins)) {
+            return;
+        }
+
+        foreach ($plugins as $plugin => $options) {
+            if (!is_string($plugin) && is_string($options)) {
+                // plugin listed not in form 'PluginName' => [....]
+                // but as non associative array like 0 => 'PluginName'
```
batopa: [minor] maybe it's better **numeric array** instead of  **non associative array**

stefanorosanelli: done

batopa (approved):

PR #1665, opened 2020-03-05
WIP: reindex ElasticSearch concurrently by object types

le0m:
This PR adds a bash script to parallelize reindex of ElasticSearch.

The script requires `tmux` and `cake.sh`, and is dependent on the `dbadmin` Cake shell to handle the actual reindex. It accepts a few arguments:
- `-t|--types <string>` a comma-separated list of object types to reindex
- `-c|--concurrency <number>` how many types to reindex at the same time (default 4)
- `-s|--session <string>` to use an existing tmux session, otherwise a new one will be created (default to 'reindex')

When started It will run the reindex for the first type with the `-delete` flag, then will run a reindex for each type in a separate tmux pane. Don't increase concurrency too much, reindexing is a cpu-intense task!

So far It's not a completely automated task since the `dbadmin` Cake shell requires user interaction if reindexing fails (see https://github.com/bedita/bedita/blob/3-corylus/bedita-app/vendors/shells/dbadmin.php#L142), so you should attach to the tmux session after starting this script.

### DIFF
--- a/reindex.sh
+++ b/reindex.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+
+function print_usage () {
+  printf "Usage: ./reindex.sh -t|--types <types> [-c|--concurrency <number>] [-s|--session <name>]\n"
+  printf "\n"
+  printf "\t-t|--types\t\tcomma-separated list of types\n"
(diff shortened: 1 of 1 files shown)


le0m:
moved to: https://github.com/bedita/bedita-tools/pull/2

PR #1667, opened 2020-03-06
Virtual methods to get/set parent by uname

stefanorosanelli:
This PR adds `_getParentUname()` and `_setParentUname()` virtual methods to `Folder` entity in order to get/set a parent via `uname` instead of ID

### DIFF
--- a/plugins/BEdita/Core/src/Model/Entity/Folder.php
+++ b/plugins/BEdita/Core/src/Model/Entity/Folder.php
@@ -95,7 +95,7 @@ class Folder extends ObjectEntity
      * @param int|null $parentId The parent id to set
      * @return int|null
      */
-    protected function _setParentId($parentId)
+    protected function _setParentId($parentId): ?int
     {
         if ($parentId === null) {
             $this->parent = null;
@@ -132,14 +132,14 @@ class Folder extends ObjectEntity
      * Setter for `parent_uname` virtual property.
      *
      * @param string|null $parentUname The parent uname to set
-     * @return void
+     * @return string|null
      */
-    protected function _setParentUname(?string $parentUname): void
+    protected function _setParentUname(?string $parentUname): ?string
     {
         if ($parentUname === null) {
             $this->parent = null;
 
-            return;
+            return null;
         }
 
         $table = TableRegistry::getTableLocator()->get($this->getSource());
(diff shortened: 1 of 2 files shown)


didoda (approved):

plugins/BEdita/Core/src/Model/Entity/Folder.php
```diff
@@ -110,6 +114,43 @@ protected function _setParentId($parentId)
...
+        return (string)Hash::get((array)$this->parents, '0.uname');
+    }
+
+    /**
+     * Setter for `parent_uname` virtual property.
+     *
+     * @param string|null $parentUname The parent uname to set
+     * @return void
+     */
+    protected function _setParentUname(?string $parentUname): void
```
batopa: > Mutator methods should always return the value that should be stored in the field.

From cakephp book.

plugins/BEdita/Core/src/Model/Entity/Folder.php
```diff
@@ -110,6 +114,43 @@ protected function _setParentId($parentId)
         return $parentId;
     }
 
+    /**
+     * Getter for `parent_uname` virtual property
+     *
+     * @return string|null
+     */
+    protected function _getParentUname(): ?string
```
batopa, line 122: We should mark the property as virtual in constructor I think.

batopa (changes requested):

plugins/BEdita/Core/src/Model/Entity/Folder.php
stefanorosanelli: Ok, good to know. I'll change it,

plugins/BEdita/Core/src/Model/Entity/Folder.php
stefanorosanelli: We didn't do it for `parent_id` virtual property and we generally don't need it in folder `meta` response where we already have `path`

A `uname` version of the path could indeed be useful, probably...

stefanorosanelli: done

plugins/BEdita/Core/src/Model/Entity/Folder.php
stefanorosanelli: Done

batopa (approved):

PR #1670, opened 2020-03-27
Custom Signup action

stefanorosanelli:
This PR introduces custom signup actions.

You can now set a custom class in configuration like this:

```php
   Configure::write('Actions.signupUserAction', '\MyPlugin\Model\Action\MySignupAction')
 ```
Or with same setting in `config/app.php` (recommended)
Custom class **must** extend  `BaseAction`.

A new trait `ActionTrait` with a `createAction` method was introduced to dynamically load actions, see method annotations for its usage.

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -530,6 +530,15 @@ return [
     ],
 
     /**
+     * Uncomment to define Custom actions to load
+     * This way some action beahavior can be overridden
+     */
+    // 'Actions' => [
+    //     'signupUserAction' => '\MyPlugin\Model\Action\SignupUserAction',
+    //     'signupUserActivationAction' => '\MyPlugin\Model\Action\SignupUserActivationAction',
+    // ],
+
+    /**
      * Default values per object type
      * object type names as keys (lower case), default property names and values as value
      */
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -534,8 +534,8 @@ return [
      * This way some action beahavior can be overridden
      */
     // 'Actions' => [
-    //     'signupUserAction' => '\MyPlugin\Model\Action\SignupUserAction',
-    //     'signupUserActivationAction' => '\MyPlugin\Model\Action\SignupUserActivationAction',
+    //     'SignupUserAction' => '\MyPlugin\Model\Action\SignupUserAction',
+    //     'SignupUserActivationAction' => '\MyPlugin\Model\Action\SignupUserActivationAction',
     // ],
 
     /**
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -530,7 +530,7 @@ return [
     ],
 
     /**
-     * Uncomment to define Custom actions to load
+     * Uncomment to define custom actions to load
      * This way some action beahavior can be overridden
      */
     // 'Actions' => [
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: plugins/BEdita/Core/src/Model/Action/ActionTrait.php [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+     * Custom class must extend BaseAction.
+     *
+     * @param string $action Action class to create
+     * @param array $options Action options
+     * @param string $config Configuration key to use
+     * @return \BEdita\Core\Model\Action\BaseAction
+     */
+    protected function createAction(string $action, array $options = [], ?string $config = null): BaseAction
+    {
+        if (empty($config)) {
```
batopa: :thinking: Calling `createAction('\MyNameSpace\MyAction')` I would expect that class was loaded, not a class defined in `Actions.MyAction` configuration. I think it can be confusing.

I would take advantage of `App::className()` to resolve class name.

```php
protected function createAction(string $class, array $options = []): BaseAction
{
    $className = App::className($class, 'Model/Action', 'Action');
    if ($className === false) {
          throw new \Exception(); 
   }

   return new $className($options);
}
```

We can create action with:

```php
$action = $this->createAction('\MyNameSpace\CustomSignupAction');

// or

$action = $this->createAction('MyPlugin.CustomSignup'); // it will be instantiate \MyPlugin\Model\Action\CustomSignupAction
```

config/app.default.php
```diff
@@ -529,6 +529,15 @@
         'default' => null,
     ],
 
+    /**
+     * Uncomment to define custom actions to load
+     * This way some action beahavior can be overridden
+     */
+    // 'Actions' => [
```
batopa, line 536: I'd like more having the configured action in `Signup` namespace. See below for consideration on `createAction()` method.

batopa (changes requested):

plugins/BEdita/Core/src/Model/Action/ActionTrait.php
stefanorosanelli: Ok, after our discussion method has been refactored

plugins/BEdita/Core/src/Model/Action/ActionTrait.php
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+namespace BEdita\Core\Model\Action;
+
+use Cake\Core\App;
+use Cake\Core\Configure;
+use Cake\Utility\Inflector;
+
+/**
+ * Trait to create actions with possible
```
batopa: Is the description incomplete?

plugins/BEdita/Core/src/Model/Action/ActionTrait.php
```diff
@@ -0,0 +1,86 @@
...
+     *      in namespaced or plugin syntax format
+     *  - otherwise class is searched using plugin syntax with 'BEdita/Core` as default in `Model\Action` namespace
+     *
+     * Examples:
+     * ```
+     *  // `SignupUserAction` from `\BEdita\Core\Model\Action\` is created, with no other check
+     *  $this->createAction('\BEdita\Core\Model\Action\SignupUserAction');
+     *
+     *  // First we look in `Actions.ListObjectsAction` config for a custom class,
+     *  // if nothing is found `BEdita/Core.ListObjectsAction` is used, looking om `Model\Action` namespace
```
batopa: [typo] looking **on**

batopa (approved):

PR #1672, opened 2020-04-03
Fix entity virtual props recursion

### LINKED ISSUE #1602: CustomProperties behavior recursion problem
There is a recursion problem in ```CustomPropertiesBehavior::isFieldSet($entity, $field)```

Suppose you have an entity virtual getter method like this: 

```php
    protected function _getMyProperty()
    {
        // ....
        $this->getTable()->loadInto($this, ['RelationName']);
       // ... 
     }
```

If `CustomProperties` behavior is used by entity's table class in `isFieldSet($entity, $field)` method this line

```php
$allProperties = $entity->toArray();
``` 
will cause a recursion since the same virtual method is called again.

stefanorosanelli:
This PR fixes #1602 a potential recursion problem in entities.

In a virtual getter or setter if you call for instance:

```
 $this->getTable()->loadInto($this, ['MyRelation']);
```

to retrieve related entities, will cause internally a new call on `$entity->toArray()` via `CustomPropertiesBehavior::isFieldSet()` that will call the above virtual method again...

### DIFF
--- a/plugins/BEdita/Core/src/Model/Behavior/CustomPropertiesBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/CustomPropertiesBehavior.php
@@ -212,13 +212,11 @@ class CustomPropertiesBehavior extends Behavior
     {
         $allProperties = $entity;
         if ($entity instanceof EntityInterface) {
-            $hidden = $entity->getHidden();
-            try {
-                $entity->setHidden([]);
-                $allProperties = $entity->toArray();
-            } finally {
-                $entity->setHidden($hidden);
-            }
+            $allProperties = array_flip(array_merge(
+                $entity->getHidden(),
+                $entity->getVisible(),
+                $entity->getVirtual()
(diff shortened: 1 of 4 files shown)


plugins/BEdita/Core/src/Model/Behavior/CustomPropertiesBehavior.php
```diff
@@ -212,15 +212,12 @@ protected function isFieldSet($entity, $field)
...
         $allProperties = $entity;
         if ($entity instanceof EntityInterface) {
-            $hidden = $entity->getHidden();
-            try {
-                $entity->setHidden([]);
-                $allProperties = $entity->toArray();
-            } finally {
-                $entity->setHidden($hidden);
-            }
+            $allProperties = array_flip(array_merge(
```
batopa: But in this way we take all entity properties not just those set. Is it correct?

batopa (approved):

PR #1673, opened 2020-04-05
Object creation on file upload #1551

### LINKED ISSUE #1551: Create media object with stream in one request
Currently an image o file with associated stream requires three steps

1. upload file as stream
1. create media object
1. link stream to media object

Introduce a new API method to upload stream and create media object in one request like:

`POST /images/upload/{filename.ext}`

where the above three steps are executed on a single request, media object attributes will have defaults like file name as title

Will this be possible only for `media` objects, object types having `media` as parent, or for every object with associated `streams`?

stefanorosanelli:
This PR fixes #1551 

With a single request like `POST /images/upload/{filename.ext}`

A new object + related stream is created in one shot.

* every object type with a `Streams` association can be created this way, otherwise a `403` error is returned
* initial object `title` will be `{filename.ext}`

Code changes:

* new `UploadController` to handle `/:object_type/upload/:file` route
* new `UploadComponent` for common stream upload logic between `StreamsController` and `UploadController`

### DIFF
--- a/plugins/BEdita/API/src/Controller/UploadController.php
+++ b/plugins/BEdita/API/src/Controller/UploadController.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * BEdita, API-first content management framework
- * Copyright 2018 ChannelWeb Srl, Chialab Srl
+ * Copyright 2020 ChannelWeb Srl, Chialab Srl
  *
  * This file is part of BEdita: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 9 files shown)


plugins/BEdita/API/src/Controller/UploadController.php
```diff
@@ -0,0 +1,65 @@
...
+        };
+
+        $this->request = $this->request
+            ->withData('title', $fileName)
+            ->withData('type', Inflector::underscore($this->Table->getAlias()));
+        // create media object from POST request
+        $this->index();
+
+        $objectId = (int)Hash::get($this->viewVars, 'data.id');
+        $this->Upload->upload($fileName, $objectId);
```
batopa: I think these operations should stay in a transaction.

batopa (changes requested):

batopa (approved):

didoda (approved):

Issue #1674, opened 2020-04-20
Modules index, filter by "created by session user"


didoda:
### Expected behavior

User wants the possibility to filter objects list, to show only objects created by him, or all.  

### Actual behavior

No "user created" filter available

didoda:
Done with #1675

PR #1675, opened 2020-04-20
filter by user_created session user

didoda:
This PR provides #1674 

If `$filter['user_created']` exists, objects search apply a filter by user_created = `$filter['user_created']`.
UI provides the possibility to choose "created by me", meaning: `$filter['user_created'] = $BEAuthUser.id`.

### DIFF
--- a/bedita-app/models/behaviors/build_filter.php
+++ b/bedita-app/models/behaviors/build_filter.php
@@ -992,6 +992,9 @@ class BuildFilterBehavior extends ModelBehavior {
         $this->fields .= $locFields;
         $this->from .= ", {$s}users{$e} AS {$s}User{$e}";
         $this->conditions[] = "{$s}User{$e}.{$s}id{$e}={$s}BEObject{$e}.{$s}user_created{$e}";
+        if (!empty($this->filter['user_created'])) {
+            $this->conditions[] = "{$s}BEObject{$e}.{$s}user_created{$e}={$this->filter['user_created']}";
+        }
         $this->group .= $locFields;
     }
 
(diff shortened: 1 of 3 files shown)


bedita-app/models/behaviors/build_filter.php
```diff
@@ -992,6 +992,9 @@ protected function user_createdFilter($s, $e) {
         $this->fields .= $locFields;
         $this->from .= ", {$s}users{$e} AS {$s}User{$e}";
         $this->conditions[] = "{$s}User{$e}.{$s}id{$e}={$s}BEObject{$e}.{$s}user_created{$e}";
+        if (!empty($this->filter['user_created'])) {
```
batopa: the value of `$this->filter['user_created']` is passed directly to filter function. You can add a `$value` parameter to this method e use it
```php
protected function user_createdFilter($s, $e, $value) {
    ....
    if (!empty($value)) {
        $this->conditions[] = "{$s}BEObject{$e}.{$s}user_created{$e}={$value}";
    }
    ....
}
```

bedita-app/views/elements/filters_form.tpl
```diff
@@ -83,6 +84,23 @@ available options:
...
+				{strip}
+				<option value="" {if !$createdbyme}selected="selected"{/if}>
+					{t}anybody{/t}
+				</option>
+				<option value="{$BEAuthUser.id}" {if $createdbyme}selected="selected"{/if}>
+					{$BEAuthUser.userid} (you)
+				</option>
+				{/strip}
+			</select>
+		</div>
```
batopa: I would indent rows as for other filters

batopa (changes requested):

batopa (approved):

PR #1677, opened 2020-04-21
Invalidate destination on first save

### LINKED ISSUE #1676: Parent tree cache invalidation on model save [first time "destination" is set]
### Expected behavior

Saving an object from module view invalidates properly parents caches

### Actual behavior

Saving an object from module view sometimes doesn't invalidate properly parents caches on first save, an extra save is needed to achieve the task.

1. Edit a document (not in the tree) [from `http://<bedita-local-vhost>/documents/view/<document-id-or-nickname>`), set a destination and save it  [first time "destination" is set]
2. Verify its parent cache: not invalidated (not expected)
3. Save the same document again, without changing data [from `http://<bedita-local-vhost>/documents/view/<document-id-or-nickname>`]
4. Verify its parent cache: invalidate (expected)

=> 2 saves to invalidate parents cache properly

didoda:
This PR fixes #1676 

Cacheable behavior, via beforeSave and afterSave, clears object cache, including parents. The first time an object is saved in a destination, tree data is still not available, so parent cache is not invalidated properly. This provides a fix in `getObjectsToCleanById`: the parent id from model data is handled as well, so that cache clear on parent is performed also on first save

Bonus: handle object indexing, considering parents explicitely, before trees data is saved

### DIFF
--- a/bedita-app/models/behaviors/cacheable.php
+++ b/bedita-app/models/behaviors/cacheable.php
@@ -112,6 +112,10 @@ class CacheableBehavior extends ModelBehavior {
             'fields' => array('parent_id'),
             'conditions' => $treeConditions
         ));
+        // handle first save in tree (object still not in tree, but parent is in destination data)
+        if (!empty($model->data['BEObject']['destination']) && !in_array($model->data['BEObject']['destination'], $excludeIds)) {
+            $parents = array_merge($parents, $model->data['BEObject']['destination']);
+        }
 
         // get related object to clean
         $objectRelation = ClassRegistry::init('ObjectRelation');
--- a/bedita-app/models/behaviors/cacheable.php
+++ b/bedita-app/models/behaviors/cacheable.php
@@ -103,34 +103,35 @@ class CacheableBehavior extends ModelBehavior {
      */
     public function getObjectsToCleanById(&$model, $objectId, array $excludeIds = array()) {
         // get parents to clean
-        $tree = ClassRegistry::init('Tree');
-        $treeConditions = array('id' => $objectId);
-        if (!empty($excludeIds)) {
-            $treeConditions['NOT'] = array('parent_id' => $excludeIds);
+        $parents = array();
+        // if destination data, use it instead of calling a tree->find
+        if (!empty($model->data['BEObject']['destination'])) {
+            // use $model->data['BEObject']['destination'], excluding $excludeIds
+            $parents = array_unique(array_diff((array)$model->data['BEObject']['destination'], $excludeIds));
         }
-        $parents = $tree->find('list', array(
-            'fields' => array('parent_id'),
-            'conditions' => $treeConditions
-        ));
-        // handle first save in tree (object still not in tree, but parent is in destination data)
-        if (!empty($model->data['BEObject']['destination']) && !in_array($model->data['BEObject']['destination'], $excludeIds)) {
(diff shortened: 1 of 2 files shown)


bedita-app/models/search_text.php
```diff
@@ -67,8 +67,12 @@ public function createSearchText($model) {
 		}
 		
 		$this->checkIndexModel();
-		if($this->indexModel) {
-			$res = $this->indexModel->indexObject($searchFields, $data);
+		if ($this->indexModel) {
+			$parents = array();
+			if (!empty($model->data[$model->alias]['destination'])) {
+				$parents = $model->data[$model->alias]['destination'];
+			}
```
le0m: there's a tab character after the `}`

bedita-app/models/behaviors/cacheable.php
```diff
@@ -112,6 +112,10 @@ public function getObjectsToCleanById(&$model, $objectId, array $excludeIds = ar
             'fields' => array('parent_id'),
             'conditions' => $treeConditions
         ));
+        // handle first save in tree (object still not in tree, but parent is in destination data)
+        if (!empty($model->data['BEObject']['destination']) && !in_array($model->data['BEObject']['destination'], $excludeIds)) {
```
batopa: :thinking: `in_array($model->data['BEObject']['destination'], $excludeIds)` seems incorrect since `$model->data['BEObject']['destination']` should be an array and the operation results `true` only if all destinations are present in one value of `$excludeIds`. 

I think it's more safe to do
```php
if (!empty($model->data['BEObject']['destination'])) {
    $addParents = array_diff((array)$model->data['BEObject']['destination'], $excludeIds);
    $parents = array_unique(array_merge($parents, $addParents));
}
```

batopa (changes requested):

bedita-app/models/behaviors/cacheable.php
```diff
@@ -103,30 +103,35 @@ public function setObjectsToClean(&$model, $objectId, array $excludeIds = array(
      */
     public function getObjectsToCleanById(&$model, $objectId, array $excludeIds = array()) {
         // get parents to clean
-        $tree = ClassRegistry::init('Tree');
-        $treeConditions = array('id' => $objectId);
-        if (!empty($excludeIds)) {
-            $treeConditions['NOT'] = array('parent_id' => $excludeIds);
+        $parents = array();
+        // if destination data, use it instead of calling a tree->find
+        if (!empty($model->data['BEObject']['destination'])) {
```
batopa, line 108: parents in `destination` should be correctly handled in `beforeSave()` I think.

```php
$treeIds = (!empty($data['destination'])) ? $data['destination'] : array();
$excludeIdsFromQuery = array_merge($treeIds, $relatedIds);
// prepare objects to clean
if (!empty($data['id'])) {
    $this->setObjectsToClean($model, $data['id'], $excludeIdsFromQuery);
}
$this->addObjectsToClean($model, $excludeIdsFromQuery);
```

Before saving an object the parents of the object we're going to save and all those in `destination` array are collects to clear them cache later.

didoda: yes... but the problem is that a find is always performed (in 3-corylus), and sometimes data is not yet in tree. that's why we use here this patch: do not use tree->find, if parents are already available

didoda:
Obsolete

PR #1679, opened 2020-04-26
Avoid `included` repetitions

stefanorosanelli:
This PR removes duplicates in `included` response array: when same objects/resources are included in a GET list response those items are repeated in `included` array.


didoda (approved):

batopa (changes requested):
Assure to have an ordered indexed array as included.

batopa (approved):

PR #1693, opened 2020-05-25
Message error for duplicate permissions on bulk edit

le0m:
This PR adds error details for bulk edit of permissions.

The new error message shows to the user which objects already had the permissions.

### DIFF
--- a/bedita-app/app_controller.php
+++ b/bedita-app/app_controller.php
@@ -1334,7 +1334,7 @@ abstract class ModulesController extends AppController {
                     return sprintf("object: %s, skipped: %s", $objectId, implode(', ', $permissionNames));
                 }, $duplicatePermissions, array_keys($duplicatePermissions));
                 
-                throw new BeditaException(__("Some permissions where already set and have been skipped", true),
+                throw new BeditaException(__("Some permissions were set already and have been skipped", true),
                     implode(" - ", $errorDetails));
             }
         }
--- a/bedita-app/app_controller.php
+++ b/bedita-app/app_controller.php
@@ -1304,38 +1304,44 @@ abstract class ModulesController extends AppController {
     {
         if (!empty($this->params['form']['objects_selected'])) {
             $objectIds = $this->params['form']['objects_selected'];
-            /**
-             * [objectID][permissionName]
-             */
-            $duplicatePermissions = array();
-            $this->Transaction->begin();
+            $permission = ClassRegistry::init('Permission');
+            $group = ClassRegistry::init('Group');
+            $errors = array();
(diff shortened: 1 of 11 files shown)


bedita-app/app_controller.php
```diff
@@ -1304,14 +1304,39 @@ public function addPermissionsObjects()
...
+                        }
+                        
+                        $duplicatePermissions[$objectId][] = $permissionData['name'];
+                    } else {
+                        $permission->add($objectId, array($permissionData));
+                    }
+                }
             }
 
             $this->Transaction->commit();
```
batopa: If `$permission->add()` fails all transaction is rollback.
Maybe we should remove the transaction here.

`Permission::isPermissionSet()` check just the `permissions.flag`, so you should check againts `flag` and `ugid` (`group_id` related to group name you have in the data array) and `'switch' => 'group'`.

Something as:

```php
$Permission = ClassRegistry::init('Permission');
$Group = ClassRegistry::init('Group');

foreach ($objectIds as $objectId) {
    foreach ($this->data['Permission'] as $permissionData) {
        $ugid = $group->field('id', array('name' => $permissionData['name']));
        $countPerms = $Permission->find('count', array(
            'conditions' => array(
                'switch' => 'group',
                'ugid' => $ugid,
                'flag' => $permissionData['flag'],
            ),
        ));
        if ($countPerms > 0) {
            $duplicatePermissions[$objectId][] = $permissionData['name']; // don't need check on key exists
            continue;
        }
        
        try {
            $Permission->add($objectId, array($permissionData));
        } catch (Exception $e) {
            // trace errros for view???
            $this->log('Error saving permission'); // more info is better
        }
    }
}
```

batopa (changes requested):

bedita-app/app_controller.php
le0m: I think we can leave the transaction, It's ok to rollback if It fails for any other reason. If an exception is raised, the transaction is rolled back and `AppController::handleError()` writes info to logs and shows details to the user.

Thanks for your snippet, I'll change the code. I guess I should add `object_id => $objectId` to the `$Permission->find(...)` conditions, right?

batopa: correct, the condition on `object_id` is missing

batopa (approved):

manuelzane:
Metterei nel `bedita.css` per la classe `messageDetail`
questi 2 valori:
```
left: 300px;
min-height: 140px;
```
la classe penso sia usata anche per altri messaggi ma non mi sembra un problema modificarla per tutti.

PR #1697, opened 2020-06-14
Feat: date from/to filters on `DateRanges` 

stefanorosanelli:
This PR introduces new finder options and related query string options on `DateRanges`: `from_date` and `to_date`. 
These options allow to filter date ranges that are *valid* from/to a specific date.

API example:

 * `GET /events?filter[date_ranges][from_date]=2020-03-01` - date ranges having a `start_date` >= '2020-03-01' and a NULL `end_time` or and `end_time` >= '2020-03-01' 
 * `GET /events?filter[date_ranges][to_date]=2020-03-01` - date ranges having a `start_date` <= '2020-03-01' and a NULL `end_time` or having `end_time`  <=  '2020-03-01' 
 * `GET /events?filter[date_ranges][from_date]=2020-03-01T00:00:00&filter[date_ranges][to_date]=2020-03-01T23:59:59` - date_ranges valid on a specific day (we can use dates or date time with ISO 8601 format)

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
@@ -205,7 +205,7 @@ class DateRangesTable extends Table
                     $q->newExpr()
                         ->gte($this->aliasField('end_date'), $from),
                 ]);
-            });
+        });
     }
 
     /**
@@ -225,7 +225,7 @@ class DateRangesTable extends Table
                     $q->newExpr()
                         ->lte($this->aliasField('end_date'), $to),
                 ]);
-            });
+        });
     }
 
     /**
--- a/plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
@@ -17,6 +17,7 @@ use BEdita\Core\Exception\BadFilterException;
 use BEdita\Core\Model\Validation\Validation;
 use BEdita\Core\ORM\QueryFilterTrait;
 use Cake\Database\Expression\QueryExpression;
+use Cake\I18n\Time;
 use Cake\ORM\Query;
 use Cake\ORM\RulesChecker;
 use Cake\ORM\Table;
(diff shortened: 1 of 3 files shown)


plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
```diff
@@ -127,19 +138,139 @@ public function buildRules(RulesChecker $rules)
...
+        unset($options['from_date'], $options['to_date']);
+
+        // filter on `start_date` and `end_date` fields
         $options = array_combine(
             array_map([$this, 'aliasField'], array_keys($options)),
             array_values($options)
         );
 
-        if (empty($options)) {
+        return $this->fieldsFilter($query, $options);
```
didoda: Scrutinizer suggests:
```
It seems like $options can also be of type false; however, parameter $options of BEdita\Core\Model\Table\...esTable::fieldsFilter() does only seem to accept array, maybe add an additional type check?
```

plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
```diff
@@ -127,19 +138,139 @@ public function buildRules(RulesChecker $rules)
...
+     * @param string $time Input time.
+     * @return null|\Cake\I18n\Time
+     */
+    protected function getTime(string $time): ?Time
+    {
+        if (empty($time)) {
+            return null;
+        }
+        try {
+            $res = new Time($time);
```
didoda: minor: it could be `return new Time($time);` and then remove `return $res;` (lines 186-187)

plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
stefanorosanelli: ok, done

plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
stefanorosanelli: right, done

batopa (approved):

didoda (approved):

PR #1701, opened 2020-06-25
`/trees` endpoint

stefanorosanelli:
See https://github.com/orgs/bedita/teams/core/discussions/13

New `/trees` endpoint has been implemented

A single object is loaded via path specification using uname or ID like:

* `GET /trees/root-folder/sub-folder/my-document`
* `GET /trees/9/10/11`

Path validity is checked, a `404` error response is displayed on invalid paths
In `meta.extra.uname_path` a uname version of the current path is displayed.

`canonical_path` is not displayed, it will be handled in a future PR

**Bonus track**: 

* you may use `include` to get related objects and `fields` query strings on `/trees`
* a new `TreesTable::findParentNodes()` finder has been added for path calculation


plugins/BEdita/API/src/Controller/TreesController.php
```diff
@@ -0,0 +1,269 @@
...
+    /**
+     * Path information with ID, object type and uname of each object
+     * Associative array having keys:
+     *  - 'ids': ID path list
+     *  - 'unames': uname path list
+     *  - 'types': object types id list
+     *
+     * @var array
+     */
+    protected $pathInfo;
```
batopa: Why do not setup here the array instead of  in`initialize()`?

stefanorosanelli: Right, I'll do it here

batopa (approved):

PR #1703, opened 2020-06-29
Sanitize uname

le0m:
The current behavior is that when object's `uname` is empty, it gets generated as a slug of `title`.

This PR adds the "slugification" even when `uname` is manually inserted.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
@@ -85,10 +85,10 @@ class UniqueNameBehavior extends Behavior
         if (empty($uname)) {
             $uname = $this->generateUniqueName($entity);
         } else {
-            $uname = Text::slug($uname, [
+            $uname = strtolower(Text::slug($uname, [
                 'replacement' => $config['replacement'],
                 'preserve' => $config['preserve'],
-            ]);
+            ]));
         }
         $count = 0;
         while (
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
```diff
@@ -82,6 +82,9 @@ public function uniqueName(EntityInterface $entity)
         if (empty($uname)) {
             $uname = $this->generateUniqueName($entity);
         }
+        else {
```
stefanorosanelli: * coding style =>  use `} else {` - see code sniff failure in Travis CI  https://travis-ci.org/github/bedita/bedita/jobs/703211809
* we should add some test to `UniqueNameBehaviorTest` 
* we have a problem with  `UsersTable::DELETED_USER_PREFIX` - `__deleted-` on CI tests in case of user anonymoization -- we could use a different uname prefix in that case... (see https://travis-ci.org/github/bedita/bedita/jobs/703211808)

stefanorosanelli (changes requested):

plugins/BEdita/Core/src/Model/Behavior/UniqueNameBehavior.php
le0m: I'll add some tests and fix the code style.
About the third point, we can allow underscore character by using the options parameter of `Text::slug()`

stefanorosanelli (approved):

stefanorosanelli (approved):

PR #1708, opened 2020-07-22
Feat/v4/trees roots

stefanorosanelli:
This PR adds a response to `/trees` endpoint without path: a list of tree root folders is displayed.


edoardocavazza:
It seems that it does not handle pagination, or am i missing something?

stefanorosanelli:
> It seems that it does not handle pagination, or am i missing something?

You're right, I just forgot it... should be trivial, but we could add it later I guess...

didoda:
obsolete

PR #1709, opened 2020-07-29
Increase `extra` to 16MB on MySQL

stefanorosanelli:
This PR increases `objects.extra` column size via migration


le0m (approved):
looks good to me :+1:

PR #1710, opened 2020-07-31
JSON Schema associations and relations metadata

stefanorosanelli:
This PR adds some metadata to current JSON Schema

* `categories`, `tags` and `date_ranges` properties are added via object type `associations` 
* associations with resources are added via `associations` key
* `relations` key will contain: `params`, `label` and related `types` - data are displayed using relation side (left or right),  for example if side is right `inverse_name` is used as `name`

Additional information example:

```json
    "associations": ["Categories"],
    "relations": {
        "has_media": {
            "label": "Has Media",
            "params": null,
            "types": [
                "media"
            ]
        }
    },
```


plugins/BEdita/Core/src/Model/Entity/ObjectType.php
```diff
@@ -315,4 +385,25 @@ protected function _getSchema()
...
     }
+
+    /**
+     * Add internal associations as properties
+     *
+     * @return array
+     */
+    protected function associationProperties(): array
+    {
+        $associations = ['Tags', 'Categories', 'DateRanges'];
```
le0m: Maybe define these as a class property? This way it can be overridden/changed more easily.

stefanorosanelli: Done as class constant for now. 
It's not something we need to change/override at the moment (i think).
Better late than never :wink:

plugins/BEdita/Core/tests/TestCase/Utility/JsonSchemaTest.php
```diff
@@ -227,8 +229,12 @@ public function testGenerate($expected, $name)
         } else {
             static::assertNotEmpty($result);
 
-            $keys = ['definitions', '$id', '$schema', 'type', 'properties', 'required', 'revision'];
-            static::assertEquals($keys, array_keys($result), '', 0, 10, true);
+            $keys = ['definitions', '$id', '$schema', 'type', 'properties', 'required', 'associations', 'relations', 'revision'];
+            if (!in_array($name, ['roles', 'streams'])) {
+                $keys = array_merge($keys, ['associations', 'relations']);
```
batopa: Why do we merge `['associations', 'relations']` that are already in `$keys`?

stefanorosanelli: No reason :wink: - I'll remove the condition

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
```diff
@@ -353,7 +436,26 @@ protected function _getSchema()
...
 
     /**
-     * Check if an object type is child of another object type.
+     * Add internal associations as properties
+     *
+     * @return array
+     */
+    protected function associationProperties(): array
+    {
+        $assocProps = array_intersect(self::ASSOC_PROPERTIES, (array)$this->associations);
```
batopa: Just a note. 
If the intention here is to possibly override `ASSOC_PROPERTIES` constant in an Entity that extends `ObjectEntity` we should take advantage of [late static bindings](https://www.php.net/manual/en/language.oop5.late-static-bindings.php) and use the `static` keyword instead of `self`.

stefanorosanelli: I don't think we will override this constant in an Entity... we will probably modify the constant as new `core` internal resources are added. Anyway it's safer to use `static` so I'm replacing it.

batopa (approved):

PR #1716, opened 2020-09-03
Override static property

stefanorosanelli:
With this PR property types of static properties can be changed.

Consider this example:

* `profiles.street_address` property refers to same column name of type TEXT 
* JSON Schema of this property is by default rich text (string with `"contentMediaType": "text/html"`) 
* you can now change this default with a new record in properties with 
  - `name` : `street_address` 
  - `object_type`: `profiles` 
  - `property_type`: `plain_text` 
  - `is_static`: true, new flag to indicate if a property is a dynamic custom property (default, with false) or a new property type of a static property 
* `street_address` JSON Schema property will be plain text now (string with `"contentMediaType": "text/plain"`)


plugins/BEdita/Core/src/Model/Entity/ObjectType.php
```diff
@@ -315,4 +312,31 @@ protected function _getSchema()
...
+     *
+     * @return array
+     */
+    protected function fetchAllProperties(): array
+    {
+        /** @var \BEdita\Core\Model\Entity\Property[] $allProperties */
+        $allProperties = TableRegistry::getTableLocator()->get('Properties')
+            ->find('objectType', [$this->id])
+            ->toArray();
+        usort($allProperties, function ($a, $b) {
```
batopa: ordering by `is_static` is not enough?

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
```diff
@@ -315,4 +312,31 @@ protected function _getSchema()
 
         return compact('properties', 'required');
     }
+
+    /**
+     * Fetch all properties for this type id.
+     * Static properties at the beginning and dynamic properties at the end.
+     * This way we can override defatul property type of a static property.
```
batopa: typo `defatul`

plugins/BEdita/Core/tests/TestCase/Model/Entity/ObjectTypeTest.php
```diff
@@ -685,4 +685,39 @@ public function testGetSchemaHiddenProperties()
...
+            [
+                'type' => 'null',
+            ],
+            [
+                'type' => 'string',
+            ],
+        ];
+        static::assertEquals($expected, $oneOf);
+
+        // // remove override property type of `street_address`
```
batopa: [iper minor] duplicate `//`

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
stefanorosanelli: good point!

batopa (approved):

PR #1721, opened 2020-09-10
YAML column properties migrations

stefanorosanelli:
This PR follows #1717 and adds database column operations via YAML.
Supported column operations are: creation, change and removal.
Creation of tables, primary keys and indexes must be performed via standard PHP migrations.

Every property in a YAML file will be treated as a column table if `is_column: true` attribute is set.
Using `column_attrubutes` array we can define any column metadata, like `default`, `limit`, `signed` used in migrations.

-------

### YAML file example 

```yaml
create:

  properties:
    - name: custom_prop
      object: foos
      property: text
      is_column: true
      column_attributes: 
         # description is used as column comment, if comment not set
         comment: custom prop
         default: null

    - name: custom_int
      object: foos
      property: integer
      is_column: true
      column_attributes: 
         signed: false
         default: 10
         limit: 5
```


plugins/BEdita/Core/src/Migration/ResourcesMigration.php
```diff
@@ -81,4 +107,133 @@ protected function getConnection(): Connection
...
+     *
+     * @param array $data Migration data.
+     * @return array
+     */
+    protected function tableColumnsActions(array &$data): array
+    {
+        $res = [];
+        foreach ($data as $action => &$value) {
+            $path = 'properties.{n}[is_column=true]';
+            $res[$action] = Hash::extract($value, $path);
```
didoda, line 135: are actions all different? then fine.
otherwise this statement override a possible previous action in the $data array.

stefanorosanelli: yes, `$action` can only be `create`, `remove` or `update` - only once in array

didoda (approved):

batopa (approved):

PR #1724, opened 2020-09-18
feat: create php custom template for migrations

didoda:
This PR provides a custom template for create migrations.

Usage:

```
bin/cake migrations create MyNewMigrations -t plugins/BEdita/Core/src/Template/Phinx/create.php.template [-p MyPlugin]
```


didoda:
This is better handled in https://github.com/bedita/bedita/pull/1724/

PR #1727, opened 2020-09-24
Save relation metadata in `PATCH /folders/{id}/parent` (#1726)

### LINKED ISSUE #1726: Update relation metadata using PATCH
### Expected behavior

Patching a relation should update metadata even if the relation already exists.

### Actual behavior

Metadata are saved if the relation has to be created, ignored if the relation already exists.

1. Create a relation: `PATCH /folder/456/parent`
```json
{
  "data": {
    "id": 123,
    "type": "folders"
  }
}
```
2. Add metadata to the relation: `PATCH /folder/456/parent`
```json
{
  "data": {
    "id": 123,
    "type": "folders",
    "meta": {
      "relation": {
        "menu": false
      }
    }
  }
}
```
3. Get the relation and check metadata: `GET /folder/456/parent`

### Environment information

* PHP version: 7.2.10
* database: MySQL
* web server: Apache

stefanorosanelli:
This PR fixes #1726 

when performing a `PATCH /folder/{id}/parent`

* parent relation metadata is now set via parent folder `_joinData` 
* `belongsTo` association is saved if target has changed (different folder ID or NULL) or if relation metadata have changed

A new `FoldersControllerTest::testPatchParentMeta()` was added.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/SetAssociatedAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SetAssociatedAction.php
@@ -123,6 +123,12 @@ class SetAssociatedAction extends UpdateAssociatedAction
 
         if ($existing === null && $relatedEntity === null) {
             return 0;
+        } elseif ($relatedEntity !== null && $this->Association->getName() !== 'ParentObjects') {
+            $bindingKey = (array)$this->Association->getBindingKey();
+
+            if ($existing !== null && $relatedEntity->extract($bindingKey) == $existing->extract($bindingKey)) {
+                return 0;
+            }
         }
 
         $entity->set($this->Association->getProperty(), $relatedEntity);
--- a/plugins/BEdita/Core/src/Model/Action/SetAssociatedAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SetAssociatedAction.php
@@ -123,12 +123,6 @@ class SetAssociatedAction extends UpdateAssociatedAction
 
         if ($existing === null && $relatedEntity === null) {
             return 0;
-        } elseif ($relatedEntity !== null) {
-            $bindingKey = (array)$this->Association->getBindingKey();
-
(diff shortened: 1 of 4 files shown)


plugins/BEdita/Core/src/Model/Action/SetAssociatedAction.php
```diff
@@ -123,15 +123,30 @@ protected function belongsTo(EntityInterface $entity, EntityInterface $relatedEn
...
+                $relatedEntity->extract($bindingKey) == $existing->extract($bindingKey) &&
+                empty($diff)
+            ) {
                 return 0;
             }
         }
 
         $entity->set($this->Association->getProperty(), $relatedEntity);
+        // set join data anyway, if empty no properties are set
+        $entity->set($joinData);
```
batopa: Since `belongsTo` association doesn't have a join table I think that handling it in this way could bring to unwanted behavior if the Action was used on another `belongsTo` relation (not between `Tree` and `Folder`).

I think that the right place to setup the data on `$entity` should be `UpdateRelatedObjectsAction::prepareData()` https://github.com/bedita/bedita/blob/4-cactus/plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php#L69

There we could add joinData to `$entity`.

Anyway then we would find a way to force the save of  `$entity` in `belongsTo()` method. 
This could be done for example setting a runtime config `$this->setConfig('forceUpdate', true);` always in `UpdateRelatedObjectsAction::prepareData()` and reading it in `SetAssociatedAction::belongsTo()` or maybe checking `$entity->isDirty()` as first thing in `SetAssociatedAction::belongsTo()` (not sure of that)

stefanorosanelli: Ok, seems quite clear.
I'll try that solution with `->isDirty()` as flag

batopa (approved):

PR #1729, opened 2020-09-29
Add endpoint_permissions to resource utility

batopa:
This PR add the feature to handle `endpoint_permissions` used for migrations and yaml migrations.
We can now handle `endpoint_permissions` creation, deleting and updating too via yaml:

```
endpoint_permissions:
  - endpoint: pets
    application: app
    role: boss
    permission: 12
```

It contains a slight refactor to `Resources::loadEntity()` to use a more general `resource` finder instead of `name` finder.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Entity/EndpointPermission.php
+++ b/plugins/BEdita/Core/src/Model/Entity/EndpointPermission.php
@@ -275,7 +275,7 @@ class EndpointPermission extends Entity
     /**
      * Setter for `application` virtual property.
      *
-     * @param string $name The role name
+     * @param string $name The application name
      * @return string|null
      */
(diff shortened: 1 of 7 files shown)


plugins/BEdita/Core/src/Model/Entity/EndpointPermission.php
```diff
@@ -226,4 +227,70 @@ protected function _setWrite($write)
...
+            ->where(['name' => $name])
+            ->firstOrFail();
+
+        return $name;
+    }
+
+    /**
+     * Setter for `application` virtual property.
+     *
+     * @param string $name The role name
```
stefanorosanelli: typo `application`

stefanorosanelli (approved):

PR #1733, opened 2020-10-12
Upgrade to cakephp 3.9

batopa:
Upgrade to CakePHP 3.9


plugins/BEdita/Core/src/Model/Table/ApplicationsTable.php
```diff
@@ -82,19 +82,19 @@ public function validationDefault(Validator $validator)
...
-            ->notEmpty('name')
+            ->notEmptyString('name')
             ->add('name', 'unique', ['rule' => 'validateUnique', 'provider' => 'table'])
 
-            ->allowEmpty('description')
+            ->allowEmptyString('description')
 
             ->boolean('enabled')
-            ->notEmpty('enabled');
+            ->notEmptyString('enabled');
```
stefanorosanelli, line 97: `notEmptyString` for bool? also int?

stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Table/ApplicationsTable.php
batopa: I was in doubt. Maybe it's better to add unit test for it.

PR #1742, opened 2020-11-18
Fix signup when user has no roles but conf require roles

batopa:
This PR fixes some issues when `Signup.roles` is configured.

All logic that checks roles is now in a custom validation method.
Now when:
* `Signup.roles` is empty => signup is allowed for users without roles
* `Signup.roles` not empty => signup requires `roles` and all those passed must be present in conf 

A little refactor was done to set config `Signup` as configuration of the action.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
@@ -96,6 +96,10 @@ class SignupUserAction extends BaseAction implements EventListenerInterface
         if (Configure::check('Signup.activationUrl') && empty($data['data']['activation_url'])) {
             $data['data']['activation_url'] = Router::url(Configure::read('Signup.activationUrl'));
         }
+        $signupRoles = (array)Hash::get($data, 'data.roles', Configure::read('Signup.defaultRoles'));
+        if (!empty($signupRoles)) {
+            $data['data']['roles'] = $signupRoles;
+        }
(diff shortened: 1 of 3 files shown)


plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -181,13 +205,53 @@ protected function validate(array $data)
...
+    {
+        $allowedRoles = (array)$this->getConfig('roles');
+        if (empty($allowedRoles) && !empty($roles)) {
+            return __d('bedita', 'Roles are not allowed on signup');
+        }
+
+        $adminRoleName = $this->Roles->get(RolesTable::ADMIN_ROLE)->get('name');
+
+        $roles = (array)$roles;
+        $message = '{0} not allowed on signup';
```
stefanorosanelli, line 236: Maybe `'Role {0} not allowed on signup'` ?

stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
batopa: I have simplified because the error shown in the exception will contain in the detail

```php
[
    'roles' => [
        'validateRoles' => 'third_role, fourth_role not allowed on signup',
    ],
]
```

that contains already information about "roles".

Moreover I wanted to avoid to distinguish between singular and plural :smile_cat:

PR #1748, opened 2020-12-01
Handle error trying to append child but not parent path is found

nikazzio:
This PR avoid to try to save on `trees ` table when not parent  path was not found.

As bonus some log was added.


bedita-app/models/objects/collection/section.php
```diff
@@ -102,9 +102,13 @@ function afterSave($created) {
 			return;
 		}
 		$tree = ClassRegistry::init('Tree');
+		$tree->create();
 
 		if (!isset($this->data[$this->name]['skipTreeUpdate']) || (isset($this->data[$this->name]['skipTreeUpdate']) && $this->data[$this->name]['skipTreeUpdate'] == false)) {
 			if ($tree->appendChild($this->id, $this->data[$this->name]['parent_id']) === false) {
+				$this->log('Error appendChild on Tree - id: ' . $this->id . ' parent_id: '  . $this->data[$this->name]['parent_id']);
+				$this->log('section->afterSave(): ' . json_encode($this->data, true), 'tree');
```
batopa: `true` in `json_encode()` non ci va

batopa (changes requested):

bedita-app/models/objects/collection/section.php
didoda: colpa mia :)

batopa (approved):

PR #1749, opened 2020-12-15
Fix notice Undefined index url in app controller

didoda:
This PR provides a fix to avoid `Undefined index: url [APP/app_controller.php, line 355]` notice

### DIFF
--- a/bedita-app/app_controller.php
+++ b/bedita-app/app_controller.php
@@ -351,25 +351,40 @@ class AppController extends Controller {
     }
 
     protected function updateHistory() {
-        // save history if configured
-        if ($this->params['url']['url'] != '/') {
-            $historyConf = Configure::read('history');
-            if ( !empty($historyConf) && $this->historyItem !== null && !$this->RequestHandler->isAjax() && !$this->RequestHandler->isFlash()) {
-                $historyModel = ClassRegistry::init('History');
-                $this->historyItem['url'] = ($this->params['url']['url']{0} != '/')? '/' . $this->params['url']['url'] : $this->params['url']['url'];
-                $user = $this->BeAuth->getUserSession();
-                if (!empty($user)) {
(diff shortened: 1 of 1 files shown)


le0m (approved):

bedita-app/app_controller.php
```diff
@@ -351,25 +351,40 @@ private function redirUrl($url) {
...
+        }
+
+        // if ajax or flash request, skip update history
+        if ($this->RequestHandler->isAjax() || $this->RequestHandler->isFlash()) {
+            return;
+        }
+
+        // update history
+        $historyModel = ClassRegistry::init('History');
+        $this->historyItem['url'] = ($this->params['url']['url']{0} != '/')? '/' . $this->params['url']['url'] : $this->params['url']['url'];
```
fquffio: I'd find slightly more readable something like this (or equivalent with string concat rather than `sprintf`):

```suggestion
        $this->historyItem['url'] = sprintf('/%s', ltrim($this->params['url']['url'], '/'); // Prepend slash without duplicating.
```

bedita-app/app_controller.php
```diff
@@ -351,25 +351,40 @@ private function redirUrl($url) {
...
 
+        // if params doesn't contain url.url, skip update history
+        if (Set::check($this->params, 'url.url') === false) {
+            return;
+        }
+
+        // if params.url.url is /, skip update history
+        if ($this->params['url']['url'] === '/') {
+            return;
+        }
```
fquffio: Perhaps we might combine two checks in one. Nothing actually changes, just a cosmetic suggestion — feel free to ignore.

```suggestion
        $url = Set::classicExtract($this->params, 'url.url');
        if (empty($url) || $url === '/') {
            // if params doesn't contain url.url, or url is /, skip update history
            return;
        }
```

fquffio (approved):

PR #1751, opened 2020-12-23
Activate routes cache

stefanorosanelli:
This PR activates routes cache using a new `_bedita_core_` cache config - we may use `default` or a new config


fquffio (approved):

fquffio:
I think we could have used [CakePHP's default](https://github.com/cakephp/app/blob/master/config/app.php#L140) `_cake_routes_ ` instead, but every name is fine actuallly…

batopa (approved):

PR #1759, opened 2021-01-14
DateRanges min/max sort

stefanorosanelli:
This PR introduces some special sort fields on `DateRanges` namely:
``` date_ranges_min_start_date, date_ranges_max_start_date, date_ranges_min_end_date, date_ranges_max_end_date```

These sort fields will enable sorting `events` or other objects having `DateRanges` with `start_date` or `end_date` in ascending/descending order.

A few examples:

* ```GET /events?sort=date_ranges_min_start_date``` - sort all events with ascending min start_date
* ```GET /events?filter[date_ranges][from_date]=2020-11-14&sort=-date_ranges_max_end_date``` - all events with valid date ranges from `2020-11-14` sorted in reverse max end date

When one of the above special sort fields are used an `INNER JOIN` with a subquery having MAX/MIN on `date_ranges.start_date` or `date_ranges.end_date` is added that will be used as `sort` field in the final query.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
@@ -299,11 +299,6 @@ class ObjectsTable extends Table
         if (empty($minMaxField)) {
             return null;
         }
-        if (strpos($minMaxField, 'date_ranges_min_') === 0) {
-            $func = $query->func()->min(substr($minMaxField, 16));
-        } else {
-            $func = $query->func()->max(substr($minMaxField, 16));
-        }
         unset($options[$minMaxField]);
         $finder = 'dateRanges';
         if (empty($options)) {
@@ -312,7 +307,10 @@ class ObjectsTable extends Table
         $subQuery = $this->DateRanges->find($finder, $options)
             ->select([
                 'date_ranges_object_id' => 'object_id',
-                $minMaxField => $func,
+                'date_ranges_min_start_date' => $query->func()->min('start_date'),
+                'date_ranges_max_start_date' => $query->func()->max('start_date'),
+                'date_ranges_min_end_date' => $query->func()->min('end_date'),
+                'date_ranges_max_end_date' => $query->func()->max('end_date'),
             ])
             ->group('object_id');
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: plugins/BEdita/Core/src/Model/Table/ObjectsTable.php [hunk 1]
```diff
@@ -257,11 +270,56 @@ protected function findType(Query $query, array $options)
...
+        }
+        unset($options[$minMaxField]);
+        $finder = 'dateRanges';
+        if (empty($options)) {
+            $finder = 'all';
+        }
+        $subQuery = $this->DateRanges->find($finder, $options)
+            ->select([
+                'date_ranges_object_id' => 'object_id',
+                $minMaxField => $func,
```
fquffio: I would simply always add these four fields:
```suggestion
                'date_ranges_min_start_date' => $query->func()->min('start_date'),
                'date_ranges_max_start_date' => $query->func()->max('start_date'),
                'date_ranges_min_end_date' => $query->func()->min('end_date'),
                'date_ranges_max_end_date' => $query->func()->max('end_date'),
```

The load on MySQL shouldn't change by much… these are already indexed fields, and it should already filter by the relevant object IDs at execution time, so I don't expect it to be a big change in query performance, and it makes this function's code much more readable IMHO and easier to maintain.

stefanorosanelli: Ok, my only concern was query performance, but we can proceed as you suggest

fquffio (approved):

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -257,11 +270,58 @@ protected function findType(Query $query, array $options)
...
+            ])
+            ->group('object_id');
+
+        return $query->distinct([
+                $this->aliasField($this->getPrimaryKey()),
+                $minMaxField,
+            ])
+            ->innerJoin(
+                ['DateBoundaries' => $subQuery],
+                ['DateBoundaries.date_ranges_object_id = ' . $this->aliasField('id')]
```
batopa: Maybe for consistency `$this->aliasField($this->getPrimaryKey())`

batopa (approved):

PR #1761, opened 2021-01-19
Layered cache

le0m:
This PR adds a cache engine with two layers, one persistent and one in-memory, for faster lookup times.

Example config:
```
'fast_cache' => [
    'className' => 'Layered',
    'persistent' => [
        'className' => 'File',
        'path' => CACHE,
        'url' => env('CACHE_DEFAULT_URL', null),
    ],
],
```
Example using a cache already configured:
```
'default' => [
    'className' => 'File',
    'path' => CACHE,
    'url' => env('CACHE_DEFAULT_URL', null),
],
'fast_cache' => [
    'className' => 'Layered',
    'persistent' => 'default',
],
```

### DIFF
--- a/plugins/BEdita/Core/src/Cache/LayeredEngine.php
+++ b/plugins/BEdita/Core/src/Cache/LayeredEngine.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * BEdita, API-first content management framework
- * Copyright 2017 ChannelWeb Srl, Chialab Srl
+ * Copyright 2021 ChannelWeb Srl, Chialab Srl
  *
  * This file is part of BEdita: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
--- a/plugins/BEdita/Core/src/Cache/LayeredEngine.php
+++ b/plugins/BEdita/Core/src/Cache/LayeredEngine.php
@@ -17,8 +17,7 @@ use Cake\Cache\CacheEngine;
 use Exception;
 
 /**
- * This engine uses two layers of cache, one persistent and one in-memory for
- * faster lookup times.
+ * This engine uses two layers of cache, one persistent and one in-memory for faster lookup times.
  */
 class LayeredEngine extends CacheEngine
 {
@@ -40,13 +39,11 @@ class LayeredEngine extends CacheEngine
      * The default config used unless overridden by runtime configuration
      *
      * - `persistent` A cache configuration or an alias, to use as persistent cache
-     * - `memory` A cache configuration or an alias, to use as in-memory cache for faster lookups
      *
      * @var array
      */
     protected $_defaultConfig = [
         'persistent' => ['className' => 'File'],
-        'memory' => ['className' => 'Array'],
     ];
 
 
@@ -59,7 +56,7 @@ class LayeredEngine extends CacheEngine
         parent::init($config);
 
         $this->_persistent = $this->getEngineInstance($this->getConfig('persistent'));
-        $this->_memory = $this->getEngineInstance($this->getConfig('memory'));
+        $this->_memory = $this->getEngineInstance(['className' => 'Array']);
 
         return true;
     }
@@ -77,13 +74,13 @@ class LayeredEngine extends CacheEngine
 
         if (is_string($config)) {
             if (!$registry->has($config)) {
-                throw new Exception("Cache alias {$config} is not defined");
+                throw new Exception("Cache engine alias {$config} is not defined");
             }
 
             $instance = $registry->get($config);
 
             if (!$instance instanceof CacheEngine) {
-                throw new Exception("Cache alias {$config} is not an implementation of CacheEngine");
+                throw new Exception("Cache engine alias {$config} is not an implementation of CacheEngine");
             }
 
             return $instance;
@@ -127,8 +124,6 @@ class LayeredEngine extends CacheEngine
 
     /**
      * {@inheritDoc}
-     *
-     * @return int|false New incremented value, false otherwise
      */
     public function increment($key, $offset = 1)
     {
--- a/plugins/BEdita/Core/src/Cache/LayeredEngine.php
+++ b/plugins/BEdita/Core/src/Cache/LayeredEngine.php
@@ -1,0 +1,179 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2017 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+namespace BEdita\Core;
+
+use Cake\Cache\Cache;
+use Cake\Cache\CacheEngine;
+use Exception;
+
+/**
+ * This engine uses two layers of cache, one persistent and one in-memory for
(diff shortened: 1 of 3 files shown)


plugins/BEdita/Core/src/Cache/LayeredEngine.php
```diff
@@ -0,0 +1,173 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2017 ChannelWeb Srl, Chialab Srl
```
fquffio:
```suggestion
 * Copyright 2021 ChannelWeb Srl, Chialab Srl
```

plugins/BEdita/Core/src/Cache/LayeredEngine.php
```diff
@@ -0,0 +1,173 @@
...
+
+            if (!$instance instanceof CacheEngine) {
+                throw new Exception("Cache engine alias {$config} is not an implementation of CacheEngine");
+            }
+
+            return $instance;
+        }
+
+        if (is_array($config)) {
+            $name = ($config['prefix'] ?? '') . $config['className'];
```
fquffio: Since branch coverage is unavailable for PHP, I generally prefer to avoid `??` and `?:` operators in favour of full `if/else` structures:
```suggestion
            $name = $config['className'];

            if (!empty($config['prefix'])) {
                $name = $config['prefix'] . $name;
            }
```

plugins/BEdita/Core/src/Cache/LayeredEngine.php
```diff
@@ -0,0 +1,173 @@
...
+
+            return $registry->load($name, $config);
+        }
+
+        throw new Exception("Unknown cache configuration");
```
fquffio:
```suggestion
        throw new Exception('Unknown cache configuration');
```

plugins/BEdita/Core/src/Cache/LayeredEngine.php
```diff
@@ -0,0 +1,173 @@
...
+    public function increment($key, $offset = 1)
+    {
+        if ($this->_memory->read($key) === false) {
+            $value = $this->_persistent->read($key);
+            $this->_memory->write($key, $value);
+        }
+
+        $this->_memory->increment($key, $offset);
+
+        return $this->_persistent->increment($key, $offset);
```
fquffio: `ArrayEngine` is not affected by concurrency issues, thus atomicity is not mandatory, but it may be for the persistent storage, especially if it uses a shared engine like Memcached or Redis:

```suggestion
        $value = $this->_persistent->increment($key, $offset);
        $this->_memory->write($key, $value);

        return $value;
```

plugins/BEdita/Core/src/Cache/LayeredEngine.php
```diff
@@ -0,0 +1,173 @@
...
+    public function decrement($key, $offset = 1)
+    {
+        if ($this->_memory->read($key) === false) {
+            $value = $this->_persistent->read($key);
+            $this->_memory->write($key, $value);
+        }
+
+        $this->_memory->increment($key, $offset);
+
+        return $this->_persistent->increment($key, $offset);
```
fquffio: Similarly here:

```suggestion
        $value = $this->_persistent->decrement($key, $offset);
        $this->_memory->write($key, $value);

        return $value;
```

plugins/BEdita/Core/src/Cache/LayeredEngine.php
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * {@inheritDoc}
+     * @throws Exception If the configuration is wrong
+     */
+    public function init(array $config = []): bool
+    {
+        parent::init($config);
+
+        $this->_persistent = $this->getEngineInstance($this->getConfig('persistent'));
+        $this->_memory = $this->getEngineInstance(['className' => 'Array']);
```
fquffio: Why not issuing `new ArrayEngine()` straight away?

le0m: You're right, this should also raise an exception within `$registry->load()` if this engine is used for multiple caches

plugins/BEdita/Core/src/Cache/LayeredEngine.php
```diff
@@ -0,0 +1,170 @@
...
+        if (is_string($config)) {
+            if (!$registry->has($config)) {
+                throw new Exception("Cache engine alias {$config} is not defined");
+            }
+
+            $instance = $registry->get($config);
+
+            if (!$instance instanceof CacheEngine) {
+                throw new Exception("Cache engine alias {$config} is not an implementation of CacheEngine");
+            }
```
fquffio: We could probably raise an exception if `$instance === $this`.

Imagine the scenario `'default' => ['className' => 'BEdita/Core.Layered', 'persistent' => 'default']`. This would end up in an infinite recursion when accessing the `$_persistent` engine.

plugins/BEdita/Core/src/Cache/LayeredEngine.php
```diff
@@ -0,0 +1,170 @@
...
+     * Persistent cache instance.
+     *
+     * @var CacheEngine
+     */
+    protected $_persistent = null;
+
+    /**
+     * In-memory cache instance.
+     *
+     * @var CacheEngine
```
fquffio: [minor] We can be more specific here. Also, in docblock annotations (`@var`, `@param`, `@return`, `@throws`) I usually prefer FQCN rather than aliases, which are more concise but are dependent on current namespace and file `use`d classes.

```suggestion
     * @var \Cake\Cache\Engine\ArrayEngine
```

plugins/BEdita/Core/src/Cache/LayeredEngine.php
```diff
@@ -0,0 +1,170 @@
...
+ * This engine uses two layers of cache, one persistent and one in-memory for faster lookup times.
+ */
+class LayeredEngine extends CacheEngine
+{
+    /**
+     * Persistent cache instance.
+     *
+     * @var CacheEngine
+     */
+    protected $_persistent = null;
```
fquffio: [minor] Using underscore to prefix private and protected class members is common in CakePHP's core, but has been discouraged for a few years. Since there's no potential conflict, I'd go with `$persistent` and `$memory`

fquffio (approved):

plugins/BEdita/Core/src/Cache/LayeredEngine.php
le0m: While writing the tests, I found out that a configuration like the one you suggested would be caught by the previous condition, because the engine itself is not yet instanced:
```
if (!$registry->has($config)) {
    throw new Exception("Cache engine alias {$config} is not defined");
}
```

The only way I could cover the recursion check, is by doing this:
```
Cache::getRegistry()->set('test', new \stdClass());
$layered = new LayeredEngine();
$layered->init(['persistent' => 'test']);
Cache::getRegistry()->set('test', $layered);
Cache::write('random', 4, 'test');
```

Do you think we should cover this?

fquffio: Ok, since this is an edge case, I'd say that we could leave the sanity check in place, but while testing let's just ensure an exception of some sort is raised, without caring too much about its class, message or code. A 100% coverage is a good thing, but if it is only achievable with workarounds, I think it's best to keep a line uncovered.

Doing so, we do test what we _actually_ care about — the fact that some sort of exception is raised if the passed configuration would result in an infinite loop — although the line where we raise the exception will remain uncovered, so it stays there just in case Cake changes the way it handles its registry or whatever other reason, and it acts as a reserve parachute.

stefanorosanelli (approved):

PR #1765, opened 2021-01-24
Db config and applications cache

stefanorosanelli:
This PR enables caching, via `_bedita_core_` configuration, of `applications` and `config` resources that are rarely changed but read from DB on every request.

### DIFF
--- a/plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
+++ b/plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * BEdita, API-first content management framework
- * Copyright 2020 ChannelWeb Srl, Chialab Srl
+ * Copyright 2016 ChannelWeb Srl, Chialab Srl
  *
  * This file is part of BEdita: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
@@ -67,7 +67,7 @@ class DatabaseConfig implements ConfigEngineInterface
     /**
      * Read `$key` parameters group from database using cache.
      *
-     * @param string!null $key The group of parameters to read from database (see `config.context`).
+     * @param string|null $key The group of parameters to read from database (see `config.context`).
      * @return array Parsed configuration values.
      */
     public function read($key): array
@@ -84,8 +84,8 @@ class DatabaseConfig implements ConfigEngineInterface
     }
 
     /**
-     * Read configuration from DB of `$key` parameters group (see database `config.context`)
-     * and return the results as an array.
+     * Read configuration from DB of `$key` parameters group and return the results as an array.
+     * Parameter group is mapped to database column `config.context`.
      *
      * @param string|null $key The group of parameters to read from database (see `config.context`).
      * @return array Parsed configuration values.
--- a/plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
+++ b/plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
@@ -29,7 +29,9 @@ use Cake\Datasource\ModelAwareTrait;
  * DatabaseConfig also manipulates how 'null', 'true', 'false' are handled.
  * These values will be converted to their boolean equivalents or to null.
  *
- * @param \BEdita\Core\Model\Table\ConfigTable $Config
+ * @since 4.0.0
+ *
+ * @property \BEdita\Core\Model\Table\ConfigTable $Config
  */
 class DatabaseConfig implements ConfigEngineInterface
 {
--- a/plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
+++ b/plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
@@ -12,10 +12,10 @@
  */
 namespace BEdita\Core\Configure\Engine;
 
-use Cake\Cache\Cache;
 use Cake\Core\Configure\ConfigEngineInterface;
 use Cake\Database\Exception;
 use Cake\Database\Expression\QueryExpression;
+use Cake\Datasource\ModelAwareTrait;
 use Cake\ORM\TableRegistry;
 
 /**
@@ -30,9 +30,13 @@ use Cake\ORM\TableRegistry;
  *
  * DatabaseConfig also manipulates how 'null', 'true', 'false' are handled.
  * These values will be converted to their boolean equivalents or to null.
+ *
+ * @param \BEdita\Core\Model\Table\ConfigTable $Config
  */
 class DatabaseConfig implements ConfigEngineInterface
 {
+    use ModelAwareTrait;
+
     /**
      * Cache config name.
      *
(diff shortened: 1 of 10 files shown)


plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
```diff
@@ -57,14 +65,33 @@ public function __construct($applicationId = null)
     }
 
     /**
-     * Read from DB (or cache) $key group of paramenters (see database `config.context`)
+     * Read `$key` parameters group from database using cache.
+     *
+     * @param string!null $key The group of parameters to read from database (see `config.context`).
```
fquffio:
```suggestion
     * @param string|null $key The group of parameters to read from database (see `config.context`).
```

plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
```diff
@@ -57,14 +65,33 @@ public function __construct($applicationId = null)
...
+     * @return array Parsed configuration values.
+     */
+    public function read($key): array
```
fquffio:
```suggestion
    public function read(?string $key): array
```

plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
```diff
@@ -57,14 +65,33 @@ public function __construct($applicationId = null)
...
+     * Read configuration from DB of `$key` parameters group (see database `config.context`)
      * and return the results as an array.
      *
-     * @param string $key The group of parameters to read from database (see `config.context`).
+     * @param string|null $key The group of parameters to read from database (see `config.context`).
      * @return array Parsed configuration values.
      * @throws \Cake\Core\Exception\Exception when parameter group is not found
      */
-    public function read($key = null)
+    protected function fetchConfig(?string $key): array
```
fquffio: I would suggest moving the database interaction entirely to the `ConfigTable`. It could be something like:

```php
// `ConfigTable`:
public function fetchConfig(?int $applicationId, ?string $context): array
{
    return $this->find()
        ->select(['name', 'content'])
        ->disableHydration()
        ->where(function (QueryExpression $exp) use ($applicationId, $context): QueryExpression {
            if (!empty($context)) {
                $exp = $exp->eq($this->aliasField('context'), $context);
            }
            if ($applicationId !== null) {
                return $exp->eq($this->aliasField('application_id'), $applicationId);
            }

            return $exp->isNull($this->aliasField('application_id'));
        })
        ->cache(sprintf('config_%s_%s', $applicationId ?: '*', $context ?: '*');
}
```

While this method could become simpler:

```php
public function read(?string $key): array
{
    // Assuming we use `ModelAwareTrait` and `$this->loadModel('Config')` in the constructor.
    return $this->Config->fetchConfig($this->applicationId, $key)
        ->filter(function (array $item): bool {
            return !in_array($item['name'], $this->reservedKeys);  // Could we use a constant `RESERVED_KEYS` instead? We no longer need to support HHVM, so array constants are fine now.
        })
        ->map(function (array $item): array {
            $content = json_decode($item['content'], true);
            if ($content === null && json_last_error() !== JSON_ERROR_NONE) {
                $content = $this->valueFromString($data['content']); // Could this be a static function?
            }

            $data['content'] = $content;

            return $data;
        })
        ->combine('name', 'content')
        ->toArray();
}
```

Doing so would delegate the caching and invalidation logic responsibility _entirely_ to the `ConfigTable`, which makes it easier to track. There would still be some overhead on every request to `json_decode` stuff, but that's totally negligible.

plugins/BEdita/Core/src/State/CurrentApplication.php
```diff
@@ -121,11 +128,15 @@ public static function setApplication(Application $application = null)
...
+     * @throws \Cake\Datasource\Exception\RecordNotFoundException
      */
-    public static function setFromApiKey($apiKey)
+    public static function setFromApiKey(string $apiKey): void
     {
         static::getInstance()->set(
-            TableRegistry::getTableLocator()->get('Applications')->find('apiKey', compact('apiKey'))->firstOrFail()
+            TableRegistry::getTableLocator()->get('Applications')
+                ->find('apiKey', compact('apiKey'))
+                ->cache('app_' . $apiKey, self::CACHE_CONFIG)
```
fquffio: Same here… I would call `->cache(sprintf('app_%s', $apiKey), static::CACHE_CONFIG)` inside `ApplicationsTable::findApiKey()`. Doing so, we would only ever interact with that portion of cache from `ApplicationsTable`, which would make maintenance easier.

le0m (approved):

didoda (approved):

batopa (approved):

plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
```diff
@@ -57,14 +65,33 @@ public function __construct($applicationId = null)
...
+     * Read `$key` parameters group from database using cache.
+     *
+     * @param string|null $key The group of parameters to read from database (see `config.context`).
+     * @return array Parsed configuration values.
+     */
+    public function read($key): array
+    {
+        $cacheKey = sprintf('db_conf_%s_%d', $key, $this->applicationId);
+
+        return (array)Cache::remember(
```
batopa: Maybe it should be better to clear cache when a configuration is saved/deleted in `ConfigTable` via `afterSave()` and `afterDelete()` ~~and take adavantage of `cache()` method of `Query` object~~

plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
stefanorosanelli: I wanted to, but unfortunately we have to implement `ConfigInterface` method  ```public function read($key);``` 
This change would cause a `PHP Fatal error`...

plugins/BEdita/Core/src/Configure/Engine/DatabaseConfig.php
stefanorosanelli: DB operations and caching moved to `ConfigTable` class as you suggested

plugins/BEdita/Core/src/State/CurrentApplication.php
stefanorosanelli: Done

plugins/BEdita/Core/src/Model/Table/ApplicationsTable.php
```diff
@@ -164,7 +171,8 @@ protected function findApiKey(Query $query, array $options)
             ->where([
                 $this->aliasField('api_key') => $options['apiKey'],
                 $this->aliasField('enabled') => true,
-            ]);
+            ])
+            ->cache(sprintf('app_%s', $options['apiKey']), static::CACHE_CONFIG);
```
fquffio, line 196: Should we also invalidate cache in `afterSave()` and `afterDelete()`?

fquffio (approved):

plugins/BEdita/Core/src/Model/Table/ApplicationsTable.php
stefanorosanelli: That's right, I'll do that

stefanorosanelli: Done

PR #1766, opened 2021-01-26
Allow empty relation `label/inverse_label`

stefanorosanelli:
This PR avoids required specification of `label` and `inverse_label` fields in relation creation.

The `humanized` version of relation names and inverse names are used as default for these fields.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/RelationsTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/RelationsTable.php
@@ -19,6 +19,8 @@ use BEdita\Core\ORM\Rule\IsUniqueAmongst;
 use Cake\Cache\Cache;
 use Cake\Database\Expression\QueryExpression;
 use Cake\Database\Schema\TableSchema;
+use Cake\Datasource\EntityInterface;
+use Cake\Event\Event;
 use Cake\ORM\Query;
 use Cake\ORM\RulesChecker;
 use Cake\ORM\Table;
@@ -106,15 +108,13 @@ class RelationsTable extends Table
             ->notEmptyString('name')
             ->add('name', 'unique', ['rule' => 'validateUnique', 'provider' => 'table'])
 
-            ->requirePresence('label', 'create')
-            ->notEmptyString('label')
+            ->allowEmptyString('label')
 
             ->requirePresence('inverse_name', 'create')
             ->notEmptyString('inverse_name')
             ->add('inverse_name', 'unique', ['rule' => 'validateUnique', 'provider' => 'table'])
 
-            ->requirePresence('inverse_label', 'create')
-            ->notEmptyString('inverse_label')
+            ->allowEmptyString('inverse_label')
 
             ->allowEmptyString('description')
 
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/tests/TestCase/Model/Table/RelationsTableTest.php
```diff
@@ -215,6 +222,28 @@ public function testInvalidateCacheAfterSave()
...
+     *
+     * @return void
+     *
+     * @covers ::beforeSave()
+     */
+    public function testBeforeSave()
+    {
+        $entity = $this->Relations->get(1);
+        $entity = $this->Relations->patchEntity($entity, ['description' => 'New description']);
+        $this->Relations->save($entity);
```
fquffio: I assume this aims to cover the "return-early" branch of the `RelationsTable::beforeSave()` method. However, this has no assertions, so it looks like it is just here to increase coverage without testing anything.

I'd prefer two distinct test cases:

```php
/**
 * Test before save callback on an existing entity.
 * 
 * @return void
 *
 * @covers ::beforeSave()
 */
public function testBeforeSaveExisting()
{
    $entity = $this->Relations->get(1);
    $originalLabel = $entity->label;
    $originalInverseLabel = $entity->inverse_label;

    $entity = $this->Relations->patchEntity($entity, [
        'name' => 'foo_bar',
        'inverse_name' => 'bar_foo',
    ]);
    $this->Relations->saveOrFail();

    $entity = $this->Relations->get(1);
    static::assertSame('foo_bar', $entity->name);
    static::assertSame($originalLabel, $entity->label);
    static::assertSame('bar_foo', $entity->inverse_name);
    static::assertSame($originalInverseLabel, $entity->inverse_label);
}

/**
 * Test before save callback with a new entity.
 * 
 * @return void
 *
 * @covers ::beforeSave()
 */
public function testBeforeSave()
{
    $entity = $this->Relations->newEntity([
        'name' => 'some_name',
        'inverse_name' => 'some_inverse_name',
    ]);
    $entity = $this->Relations->saveOrFail($entity);

    static::assertSame('Some Name', $entity->get('label'));
    static::assertSame('Some Inverse Name', $entity->get('inverse_label'));
}

**Review by fquffio — CHANGES_REQUESTED** [review]

File: plugins/BEdita/Core/tests/TestCase/Model/Table/RelationsTableTest.php (hunk 1 above)
Comment (stefanorosanelli) [reply plugins/BEdita/Core/tests/TestCase/Model/Table/RelationsTableTest.php]: Right, try to improve only coverage is a bad habit :smile_cat:
I'll use your suggestion

Comment (stefanorosanelli) [reply plugins/BEdita/Core/tests/TestCase/Model/Table/RelationsTableTest.php]: done

File: plugins/BEdita/Core/src/Model/Table/RelationsTable.php [hunk 2]
```diff
@@ -106,15 +108,13 @@ public function validationDefault(Validator $validator)
             ->notEmptyString('name')
             ->add('name', 'unique', ['rule' => 'validateUnique', 'provider' => 'table'])
 
-            ->requirePresence('label', 'create')
-            ->notEmptyString('label')
+            ->allowEmptyString('label')
```
batopa: maybe it's more safe to allow empty string only on create.

What happens editing a relation setting `label` or `inverse_label` as `''`?

stefanorosanelli: right, something like `->allowEmptyString('label', true, 'create)` ?

stefanorosanelli: done

batopa (approved):

PR #1767, opened 2021-02-03
Status level check

stefanorosanelli:
This PR fixes two possible problems creating new objects on an application having a configured `Status.level`

 *  if `Status.level` is `on` newly created objects must have `on` as default status, otherwise creation via `POST` will fail (object is created, but not read and a `404` error is generated) 
 * if `status` is specified
   - it must be `on` if `Status.level` is `on`
   - cannot be `off` is `Status.level` is `draft`

### DIFF
--- a/plugins/BEdita/Core/src/Model/Behavior/DataCleanupBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/DataCleanupBehavior.php
@@ -24,7 +24,7 @@ use Cake\Utility\Inflector;
  * Data cleanup operations on object creations to allow operations with `dirty` input data
  * and setting basic values per fields, per object type (from configuration 'DefaultValues')
  *
- * This Behavoir acts only on Model.beforeMarshal event
+ * This Behavior acts only on Model.beforeMarshal event
  *
  * @since 4.0.0
  */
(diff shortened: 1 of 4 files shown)


didoda (approved):

plugins/BEdita/Core/src/Model/Behavior/DataCleanupBehavior.php
```diff
@@ -57,17 +57,34 @@ public function beforeMarshal(Event $event, \ArrayObject $data)
...
+     * @param string $type Type name.
+     * @return array
+     */
+    protected function defaultFields(string $type): array
+    {
+        $fields = (array)$this->getConfig('fields');
+        // set default `on` if minimum level is `on`
+        if (Configure::read('Status.level') === 'on') {
+            $fields['status'] = 'on';
+        }
```
fquffio, line 85: This covers the case where default status is `off`/`draft` and status level is `on`. Should we also cover the case where default status is `off` and status level is `draft`?

fquffio (approved):

batopa (approved):

plugins/BEdita/Core/src/Model/Behavior/DataCleanupBehavior.php
stefanorosanelli: I don't think it's necessary. This is just to avoid the use of `draft` global default if no default value is specified in `DefaultValues` configuration.

PR #1773, opened 2021-02-08
Fix `?include` check

fquffio:
This PR fixes an issue that caused `?include` query parameter to be validated on source table rather than the target.

As an example, let's assume a relation "Foo" exists between **Documents** and **Locations**, and a second relation "Bar" exists between **Locations** and **Profiles**.

A request to `GET /documents/:id/foo?include=bar` would be legit. However, it resulted in a **400 Bad Request** with the following message:
> Relationship "bar" does not exist.

This happened because the relationship _Bar_ was being checked on table **Documents** (the source table for the _Foo_ relation) rather than on **Locations** (the target table), therefore it couldn't be found.

## Bonus Track

An extra improvement coming from this refactor is the Table object chosen when a relation allows multiple object types on the "other side". Previously, in case the types allowed were more than one, Objects was always chosen. Now, the closest common ancestor is chosen instead. This means that if a relation **Foo** allows both Images and Videos on the right side (and Documents on the left-hand side), the Media table would be used instead in a request like `GET /documents/42/foo`.

Please note that this is just the first step towards dealing with union and intersection types properly… for instance, if another relation **Bar** exists with both Images and Videos on the left-hand side, the following request would still fail: `GET /documents/24/foo?include=bar`.

### DIFF
--- a/plugins/BEdita/API/src/Controller/FoldersController.php
+++ b/plugins/BEdita/API/src/Controller/FoldersController.php
@@ -15,6 +15,7 @@ namespace BEdita\API\Controller;
 use BEdita\Core\Model\Action\ListRelatedFoldersAction;
 use Cake\Http\Exception\NotFoundException;
 use Cake\ORM\Association;
+use Cake\ORM\Table;
 use Cake\Utility\Hash;
 
 /**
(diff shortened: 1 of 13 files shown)


didoda (approved):

plugins/BEdita/API/src/Controller/FoldersController.php
```diff
@@ -47,14 +48,15 @@ class FoldersController extends ObjectsController
...
+        if ($table === null) {
+            switch ($relationship) {
+                case 'parents':
+                    throw new NotFoundException(__d('bedita', 'Relationship "{0}" does not exist', $relationship));
+                case 'parent':
+                    return $this->Table->getAssociation('Parents');
+            }
         }
 
         return parent::findAssociation($relationship);
```
le0m: shouldn't this pass `$table` along to the parent class?

fquffio: Ouch, I'm afraid you're right!
I'm fixing it and adding an integration test, if I can…

le0m (approved):

batopa (approved):

PR #1774, opened 2021-02-10
`uname` allowed in relationship management

fquffio:
This PR introduces the ability to manage BEdita semantic relations using `uname`s in place of `id`s.

Let **Foo** be a relation between **Documents** and **Users**. The following payload would now be legit:

```http
PATCH /documents/42/relationships/foo

{
    "data": [
        {
            "id": "1",
            "type": "users"
        },
        {
            "id": "some-other-user's-uname",
            "type": "users"
        }
    ]
}
```

Similarly, also `POST /documents/42/relationships/foo` and `DELETE /documents/42/relationships/foo` would admit unames in place of IDs.

### DIFF
--- a/plugins/BEdita/API/src/Model/Action/UpdateRelatedAction.php
+++ b/plugins/BEdita/API/src/Model/Action/UpdateRelatedAction.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2021 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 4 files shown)


le0m (approved):

plugins/BEdita/API/src/Model/Action/UpdateRelatedAction.php
```diff
@@ -0,0 +1,100 @@
...
+use Cake\Database\Expression\QueryExpression;
+use Cake\Datasource\Exception\RecordNotFoundException;
+use Cake\ORM\Association;
+use Cake\ORM\Table;
+use Cake\Utility\Hash;
+
+/**
+ * Command to update relations between objects.
+ *
+ * @since 4.0.0
```
batopa: [minor] `@since` 4.4.0?

stefanorosanelli (approved):

batopa (approved):

plugins/BEdita/API/src/Model/Action/UpdateRelatedAction.php
fquffio: Right!

PR #1776, opened 2021-02-15
Placeholders

fquffio:
This PR introduces a new core relation: `placeholder` / `placeholded`.

Said relation cannot be manually edited, thus `POST|PATCH|DELETE /objects/{id}/relationships/placeholder` will always result in a **403 Forbidden**, while `GET /objects/{id}/relationships/placeholder` will work as expected.

The introduced `PlaceholderBehavior` will parse the configured properties of any entity being saved to extract any reference to other BEdita objects as placeholders.

The default syntax for BEdita placeholders is as an HTML comment:
```html
<!-- BE-PLACEHOLDER.42 --> (Simple placeholder, no parameters)
<!-- BE-PLACEHOLDER.42.Zm9vIGJhcgo= --> (Placeholder with custom parameters. Parameters are in a free form, but they must be base64-encoded)
```

Objects referenced as placeholders are restricted from being deleted. The rationale for this is that when an entity (say, an image) is referenced within a document as a placeholder, is considered to be an integral part of the document itself rather than an attachment to it, therefore the document must be edited in order to remove the placeholder before the referenced image can be deleted.


fquffio:
This feature has been moved to a dedicated plugin: [bedita/placeholders](https://github.com/bedita/placeholders).

PR #1778, opened 2021-02-23
Expose available custom properties getting related object

batopa:
This PR improve the result set of related objects including the related value of available custom properties.

Before this, having a relation as

```
documents <---- relation_name / inverse_realtion_name ----> profiles
```

and having defined custom prop `another_email` on `profiles` type didn't expose that property in the result set unless it wasn't valued.

So calling

```http
GET /documents/{id}/relation_name
```
the result was missing of `another_email` property. 

After the PR the result will contain `"another_email": null`.

### DIFF
--- a/plugins/BEdita/API/tests/IntegrationTest/AssociatedEntitiesTest.php
+++ b/plugins/BEdita/API/tests/IntegrationTest/AssociatedEntitiesTest.php
@@ -212,7 +212,7 @@ class AssociatedEntitiesTest extends IntegrationTestCase
      *
      * @return void
      */
-    public function testRelated()
+    public function testRelated(): void
     {
         $this->configRequestHeaders();
         $this->get('/profiles/4/inverse_test');
--- a/plugins/BEdita/API/tests/IntegrationTest/AssociatedEntitiesTest.php
+++ b/plugins/BEdita/API/tests/IntegrationTest/AssociatedEntitiesTest.php
@@ -208,6 +208,20 @@ class AssociatedEntitiesTest extends IntegrationTestCase
     }
 
     /**
+     * Test that related objects correspond to the pagination count.
+     *
+     * @return void
+     */
+    public function testRelated()
+    {
+        $this->configRequestHeaders();
+        $this->get('/profiles/4/inverse_test');
+        $result = json_decode((string)$this->_response->getBody(), true);
+        static::assertCount(2, $result['data']);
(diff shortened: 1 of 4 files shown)


plugins/BEdita/API/tests/IntegrationTest/AssociatedEntitiesTest.php
```diff
@@ -207,6 +207,20 @@ public function testIncludedRelated()
         static::assertCount(2, $result['included']);
     }
 
+    /**
+     * Test that related objects correspond to the pagination count.
+     *
+     * @return void
+     */
+    public function testRelated()
```
fquffio:
```suggestion
    public function testRelated(): void
```

plugins/BEdita/Core/src/Model/Behavior/CustomPropertiesBehavior.php
```diff
@@ -80,16 +80,16 @@ public function getAvailable()
...
+        if ($objectType === null) {
             return [];
         }
 
+        $properties = TableRegistry::getTableLocator()->get('Properties')->find('type', ['dynamic'])
+            ->find('objectType', [$objectType->id])
+            ->where(['enabled' => true, 'is_static' => false])
+            ->all();
+
         $this->available = collection($properties)->indexBy('name')->toArray();
```
fquffio:
```suggestion
        $this->available = TableRegistry::getTableLocator()->get('Properties')
            ->find('type', ['dynamic'])
            ->find('objectType', [$objectType->id])
            ->where(['enabled' => true, 'is_static' => false])
            ->indexBy('name')
            ->toArray();
```

plugins/BEdita/Core/tests/TestCase/Model/Behavior/CustomPropertiesBehaviorTest.php
```diff
@@ -152,6 +152,31 @@ public function testGetAvailable(array $expected, $tableName)
...
 
+    /**
+     * Test get available properties for related object.
+     *
+     * @return void
+     *
+     * @covers ::getAvailable()
+     * @covers ::objectType()
+     */
+    public function testGetAvailableRelatedObject()
```
fquffio:
```suggestion
    public function testGetAvailableRelatedObject(): void
```

fquffio (approved):

stefanorosanelli (approved):

PR #1780, opened 2021-02-26
External Auth Signup: add `verified`, remove reference on `anonymize`

stefanorosanelli:
This PR solves some issues in `SignupUserAction`

* `verified` timestamp is set upon external auth provider signup - there's no other way to set this meta attribute since no validation link is sent to users signing with external auth (user is actually verified by provider) 
* `external_auth` records are removed in `UsersTable::anonymizeUser` - a user previously anonymized ws unable to signup again due to conflicting record with same `auth_provider_id` and `provider_username`
* deprecated code removed in `getOAuth2Response`

### DIFF
--- a/plugins/BEdita/API/src/Auth/OAuth2Authenticate.php
+++ b/plugins/BEdita/API/src/Auth/OAuth2Authenticate.php
@@ -104,12 +104,12 @@ class OAuth2Authenticate extends BaseAuthenticate
      * @return array Response from an OAuth2 provider
      * @codeCoverageIgnore
      */
-    protected function getOAuth2Response($url, $accessToken)
+    protected function getOAuth2Response($url, $accessToken): array
     {
         /** @var \Cake\Http\Client\Response $response */
         $response = (new Client())->get($url, [], ['headers' => ['Authorization' => 'Bearer ' . $accessToken]]);
 
-        return !empty($response->getJson()) ? $response->getJson() : [];
+        return (array)$response->getJson();
     }
 
     /**
--- a/plugins/BEdita/API/src/Auth/OAuth2Authenticate.php
+++ b/plugins/BEdita/API/src/Auth/OAuth2Authenticate.php
@@ -104,7 +104,7 @@ class OAuth2Authenticate extends BaseAuthenticate
      * @return array Response from an OAuth2 provider
      * @codeCoverageIgnore
      */
-    protected function getOAuth2Response($url, $accessToken): array
+    protected function getOAuth2Response(string $url, string $accessToken): array
     {
         /** @var \Cake\Http\Client\Response $response */
         $response = (new Client())->get($url, [], ['headers' => ['Authorization' => 'Bearer ' . $accessToken]]);
--- a/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
@@ -358,7 +358,7 @@ class SignupUserAction extends BaseAction implements EventListenerInterface
      * @return array Response from an OAuth2 provider
      * @codeCoverageIgnore
      */
-    protected function getOAuth2Response($url, $accessToken): array
+    protected function getOAuth2Response(string $url, string $accessToken): array
     {
         $response = (new Client())->get($url, [], ['headers' => ['Authorization' => 'Bearer ' . $accessToken]]);
 
--- a/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
@@ -357,11 +357,11 @@ class SignupUserAction extends BaseAction implements EventListenerInterface
      * @return array Response from an OAuth2 provider
      * @codeCoverageIgnore
      */
-    protected function getOAuth2Response($url, $accessToken)
+    protected function getOAuth2Response($url, $accessToken): array
     {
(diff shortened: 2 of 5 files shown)


plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -309,15 +310,15 @@ protected function createUserEntity(array $data, $status, $validate)
             ]);
         }
         $action = new SaveEntityAction(['table' => $this->Users]);
+
         $data['status'] = $status;
+        $entity = $this->Users->newEntity();
+        $entityOptions = compact('validate');
+        if (!empty($data['verified'])) {
+            $entityOptions['accessibleFields'] = ['verified' => true];
+        }
```
batopa: In this way passing `$data['verified']` the verification will be bypassed. I think it would be better if `$verified` was passed as fourth parameter

```php
protected function createUserEntity(array $data, $status, $validate, $verified = false)
```
In this way we don't break backward compatibility.

batopa (changes requested):

plugins/BEdita/API/src/Auth/OAuth2Authenticate.php
```diff
@@ -104,12 +104,12 @@ public function getUser(ServerRequest $request)
      * @return array Response from an OAuth2 provider
      * @codeCoverageIgnore
      */
-    protected function getOAuth2Response($url, $accessToken)
+    protected function getOAuth2Response($url, $accessToken): array
```
fquffio: Since we're there…

```suggestion
    protected function getOAuth2Response(string $url, string $accessToken): array
```

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
fquffio: Using this approach, I'd suggest setting the `verified` field directly without going through `patchEntity`:

```php
if ($verified === true) {
    $entity->set('verified', FrozenTime::now());
}
```

Setting fields directly bypasses validation, and we can avoid messing with accessible fields.

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -357,11 +358,11 @@ protected function checkExternalAuth(array $data)
      * @return array Response from an OAuth2 provider
      * @codeCoverageIgnore
      */
-    protected function getOAuth2Response($url, $accessToken)
+    protected function getOAuth2Response($url, $accessToken): array
```
fquffio:
```suggestion
    protected function getOAuth2Response(string $url, string $accessToken): array
```

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -357,11 +358,11 @@ protected function checkExternalAuth(array $data)
      * @return array Response from an OAuth2 provider
      * @codeCoverageIgnore
```
fquffio, line 360: Why is this ignored? If we're unable or not willing to cover a method it's better just to leave it uncovered, rather than cheating… 😉

plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -392,6 +392,8 @@ protected function anonymizeUser(EntityInterface $entity)
         $entity->set('deleted', true);
         $entity->set('locked', true);
 
+        $this->ExternalAuth->deleteAll([$this->ExternalAuth->aliasField('user_id') => $entity->get('id')]);
```
fquffio: `Table::deleteAll()` doesn't invoke any callbacks on deleted entities. While it should be safe, I think we can assume that external auth methods will always be a small number for each user, thus I'd prefer to invoke `Table::delete()` on each row linked to the user being anonymized.

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
stefanorosanelli: Yeah, we should definitely test it... but I agree we should remove the ignore tag until we've done it.

plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
stefanorosanelli: ok, I'd use the `$entity->set()` approach with new `$verified` argument

batopa (approved):

plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -392,7 +392,14 @@ protected function anonymizeUser(EntityInterface $entity)
...
 
-        $this->ExternalAuth->deleteAll([$this->ExternalAuth->aliasField('user_id') => $entity->get('id')]);
+        $condition = [$this->ExternalAuth->aliasField('user_id') => $entity->get('id')];
+        $extAuthList = $this->ExternalAuth->find()->where($condition)->toList();
+        array_walk(
+            $extAuthList,
+            function ($item) {
+                $this->ExternalAuth->deleteOrFail($item);
+            }
+        );
```
fquffio: I'd leave the marshaller do the loading, so we don't need to explicitly write conditions.

```suggestion
        $this->loadInto($entity, ['ExternalAuth']);
        array_walk(
            $entity->external_auths,
            function ($item) {
                $this->ExternalAuth->deleteOrFail($item);
            }
        );
```

fquffio (approved):

PR #1789, opened 2021-03-31
chore: remove custom exception renderer behaviour

nicolocarpignoli:
Remove custom Exception Renderer handling in order to handle default JSON error payload.


fquffio:
+1. Although it was originally designed to support a different exception renderer than the one provided by default with BEdita/API, it is not something we're actually using anywhere as far as I can tell.

This shouldn't affect any installation using the application-based approach, and it would make things a little easier for apps using the plugin-based approach.

fquffio (approved):

stefanorosanelli (approved):

PR #1790, opened 2021-04-01
Endpoints permissions cache

stefanorosanelli:
This PR introduces a cache for `Endpoints` and `EndpointPermissions` resources that are rarely changed in a BEdita application but loaded in `EndpointAuthorize` auth class upon every API call.

* a new `QueryCacheTable` abstract base class to be extended by resource table classes using cache (like `Applications`, `Config`..). Could be also a Trait, but I don't expect to use this same pattern elsewhere. 
* new table methods using cache `EndpointsTable::fetchId()`, `EndpointPermissionsTable::fetchCount()` and EndpointPermissionsTable::fetchPermissions()` have been added    
* `EndpointAuthorize::auhorize()` method has been changed in order to use these new cached methods 
* unit tests have been updated accordingly


batopa:
Overall it's ok, but maybe I would move the cache layer in a `Behavior` instead that in parent class.
We could have a config by default in that behavior but it would easily to setup new cache config

```php
$this->addBehavior('QueryCache', [
    'cacheConfig' => '_my_custom_cache',
]);
```

Maybe we could implement a custom finder, something like

```php
public function findCache(Query $query, array $options): Query
{
    if (empty($options['key'])) {
        return $query;
    }

    return $query->cache($options['key'], $this->getConfig('cacheConfig');
}
```

and use in Table class, for example:

```php
return $this->find('byApplication', compact('applicationId'))
    ->find('byEndpoint', compact('endpointIds'))
    ->find('cache', compact('key'))
    ->count();
```

stefanorosanelli:
> Overall it's ok, but maybe I would move the cache layer in a `Behavior` instead that in parent class.

Use of a new Behavior with `cache` finder seems fine to me. 
@fquffio do you suggest other solutions or is it ok for you too?

fquffio:
Mmh… not quite sure. Caching is easy, but keeping track of what has to be pruned is always the tricky part. I'd be ok in introducing a behavior if it also helped in that task too, but adding a custom finder just to add `->cache($theOptionYouJustPassedToTheFinderItSelf)` doesn't add a great value, IMHO.

Also, adding a finder would mean that caching is exposed via API: `/foos?filter[cache][key]=myOwnClientGeneratedCacheKey` would cache results using a cache key set by the client, which would never be invalidated if it doesn't comply with model's logic… or it would bypass permission checks and return a cached value that was supposed to be served to someone else if the attacker knows enough pieces of information about the system.

Of course, we could deny the `cache` filter in querystring by adding some sort of denylist, but that would require adding clumsy logic all over the place, and I don't think the goal is worth the hassle for now.

plugins/BEdita/Core/src/Model/Behavior/QueryCacheBehavior.php
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+namespace BEdita\Core\Model\Behavior;
+
+use Cake\Cache\Cache;
+use Cake\ORM\Behavior;
+use Cake\ORM\Query;
+
+/**
+ * Abstract base class for table classes using query `cache` in some methods.
```
batopa: The doc for this class should be fixed.

stefanorosanelli: Done

batopa (approved):

PR #1795, opened 2021-05-04
Ensure custom properties results formatter is prepended

fquffio:
This PR fixes an issue that caused a fatal error in case other result formatters were set on the Query object.

This caused the `CustomPropertiesBehavior` to append its own result formatter at the end of the queue, but if the custom formatter doesn't return a CollectionInterface a fatal error is raised due to the anonymous function signature.

Since custom properties are supposed to be handled transparently, it makes sense to always prepend the result formatter rather than appending it.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Behavior/CustomPropertiesBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/CustomPropertiesBehavior.php
@@ -110,7 +110,7 @@ class CustomPropertiesBehavior extends Behavior
      *
      * @param \Cake\Event\Event $event Fired event.
      * @param \Cake\ORM\Query $query Query object instance.
-     * @return void
+     * @return \Cake\ORM\Query
      */
     public function beforeFind(Event $event, Query $query): Query
     {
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Model/Behavior/CustomPropertiesBehavior.php
```diff
@@ -112,13 +112,16 @@ public function getDefaultValues()
...
      */
-    public function beforeFind(Event $event, Query $query)
+    public function beforeFind(Event $event, Query $query): Query
     {
-        $query->formatResults(function (CollectionInterface $results) {
-            return $results->map(function ($row) {
-                return $this->promoteProperties($row);
-            });
-        });
+        return $query->formatResults(
```
stefanorosanelli, line 117: Do we need to `return` the Query object? New formatter is added to `$query` object in use.

Also `@return void` annotation conflicts with `Query` return type.

fquffio: No, it is not strictly needed since it is modified in place. However, I personally prefer to `return` it because it better complies with a functional style — i.e.: it will continue to work even if the `formatResult()` method will start returning a _copy_ of the Query object at some point — which Cake has started adopting in many cases (e.g. Request and Response objects, etc.).

You're right about the annotation. I'm fixing it.

stefanorosanelli (approved):

Issue #1792, opened 2021-04-09
Project model schema 


stefanorosanelli:
Retrieve from a new API endoint (like `/model/project`) a JSON representation of the entire project model schema containing some important resources like:

* object types
* property types
* relations
* properties
* categories
* roles?
* applications?

This schema should not include non portable data like `id` fields or secrets (in `applications` for instance) 
Project schema can be used to synchronize same project model on different environments.

stefanorosanelli:
done via #1798 closing

PR #1798, opened 2021-05-15
Project model endpoint & command

stefanorosanelli:
This PR introduces the `project model` concept and first tools. 
It is a first implementation of #1792

A new utility class `ProjectModel` has been added with two main methods:
* `generate()` - to generate current project model
* `diff()` - to calculate a `diff` between current project model and a new one passed by argument => resulting array will have `create`, `update` or `remove` keys containing the corresponding model items 

The **project model** is (for now) a simple array with 6 main keys where main internal resources are described in the same format used by [resources migrations](https://github.com/bedita/bedita/wiki/YAML-format-to-update-resources):
 * `applications`
 * `roles`
 * `property_types`
 * `object_types`
 * `relations`
 * `properties` (only custom properties for now)

A new `GET /model/project` endpoint is now available with a response containing the above array.
 
A new `ProjectModelCommand` class has been created to sync the current model with a new one:

Usage:
```
bin/cake project_model [-f {/path/to/file}] [-p {Plugin}] [--cache-clear]
``` 
- `/path/to/file` path to JSON file containing new project model
- if no `-f/--file` option is used a default `config/project_model.json` path is used, relative to the app root directory
- if `-p/--plugin` option is used default path `config/project_model.json` will be relative to the plugin root directory
- if `-c/--cache-clear` option is used cache is cleared after resources change

### DIFF
--- a/plugins/BEdita/API/config/routes.php
+++ b/plugins/BEdita/API/config/routes.php
@@ -176,6 +176,11 @@ Router::plugin(
                     ['controller' => 'Schema', 'action' => 'jsonSchema'],
                     ['_name' => 'schema', 'pass' => ['type']]
                 );
+                $routes->get(
+                    '/project',
+                    ['controller' => 'Project', 'action' => 'index'],
+                    'project'
+                );
             }
         );
 
--- a/plugins/BEdita/API/config/routes.php
+++ b/plugins/BEdita/API/config/routes.php
@@ -177,10 +177,10 @@ Router::plugin(
                     ['controller' => 'Schema', 'action' => 'jsonSchema'],
                     ['_name' => 'schema', 'pass' => ['type']]
                 );
-                $routes->get(
+                $routes->connect(
                     '/project',
                     ['controller' => 'Project', 'action' => 'index'],
-                    'project'
+                    ['_name' => 'project']
                 );
             }
         );
--- a/plugins/BEdita/API/src/Controller/Model/ProjectController.php
+++ b/plugins/BEdita/API/src/Controller/Model/ProjectController.php
@@ -16,6 +16,7 @@ namespace BEdita\API\Controller\Model;
 use BEdita\API\Controller\AppController;
 use BEdita\Core\Utility\ProjectModel;
 use Cake\Event\Event;
+use Cake\Http\Response;
 
 /**
  * Controller for `/model/project` endpoint.
@@ -35,7 +36,7 @@ class ProjectController extends AppController
     /**
      * {@inheritDoc}
      */
-    public function initialize()
+    public function initialize(): void
     {
         parent::initialize();
         if ($this->components()->has('JsonApi')) {
@@ -61,7 +62,7 @@ class ProjectController extends AppController
      *
      * @return \Cake\Http\Response
      */
-    public function index()
+    public function index(): Response
     {
         $this->request->allowMethod(['get']);
 
--- a/plugins/BEdita/API/src/Controller/Model/ProjectController.php
+++ b/plugins/BEdita/API/src/Controller/Model/ProjectController.php
@@ -16,7 +16,7 @@ namespace BEdita\API\Controller\Model;
 use BEdita\API\Controller\AppController;
 use BEdita\Core\Utility\ProjectModel;
 use Cake\Event\Event;
-use Cake\Http\Response;
+use Cake\Http\Exception\NotAcceptableException;
 
 /**
  * Controller for `/model/project` endpoint.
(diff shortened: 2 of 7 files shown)


plugins/BEdita/API/config/routes.php
```diff
@@ -177,6 +177,11 @@ function (RouteBuilder $routes) use ($modelingControllers, $resourcesRoutes) {
                     ['controller' => 'Schema', 'action' => 'jsonSchema'],
                     ['_name' => 'schema', 'pass' => ['type']]
                 );
+                $routes->get(
```
batopa: In this way a request as `POST /model/project` obtains a `404 Not Found` as response. I would prefer a `405 Method Not Allowed` response in these cases. So I suggest to use `connect()`

plugins/BEdita/API/src/Controller/Model/ProjectController.php
```diff
@@ -0,0 +1,77 @@
...
+            $this->components()->unload('JsonApi');
+        }
+        $this->viewBuilder()->setClassName('Json');
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * Intentionally left blank to override parent method.
+     * Avoid content-type negotiation checks based on `Accept` header.
```
batopa: Why do we need it to escape it? I think `application/json` must be accepted but for example `text/html` (and also `application/vnd.api+json`) should be refused.

If it is correct I would replace `$this->viewBuilder()->setClassName('Json')` in `initialize()` with 

```php
$this->RequestHandler->setConfig('viewClassMap.json', 'Json');
```
That component will be responsible to set the right content type and `JsonView` class to use.

To complete the negotiation `beforeFilter()` method could contain

```php
if (!$this->request->is(['json'])) {
        throw new NotAcceptableException(
            __d('bedita', 'Bad request content type "{0}"', $this->request->getHeaderLine('Accept'))
        );
}
```

plugins/BEdita/API/src/Controller/Model/ProjectController.php
```diff
@@ -0,0 +1,77 @@
...
+    public function index(): Response
+    {
+        $this->request->allowMethod(['get']);
+
+        $model = ProjectModel::generate();
+
+        $this->set($model);
+        $this->set('_serialize', true);
+
+        return $this->render()
```
batopa: Applying content negotiation we don't need it.

plugins/BEdita/Core/src/Command/ProjectModelCommand.php
```diff
@@ -0,0 +1,121 @@
...
+
+            return self::CODE_SUCCESS;
+        }
+
+        Resources::save($diff);
+        $io->success('Project model updated');
+
+        foreach (Cache::configured() as $cache) {
+            Cache::clear(false, $cache);
+        }
```
batopa: It should be equal to `Cache::clearAll()`

plugins/BEdita/Core/tests/TestCase/Command/ProjectModelCommandTest.php
```diff
@@ -0,0 +1,183 @@
+<?php
+namespace BEdita\Core\Test\TestCase\Command;
```
batopa, line 13: Missing license and copyright disclaimer

batopa (changes requested):

plugins/BEdita/API/config/routes.php
stefanorosanelli: ok

plugins/BEdita/API/src/Controller/Model/ProjectController.php
stefanorosanelli: In some cases, like `GET /model/schema` we skipped content-negotiation just to have a simpler integration.
We could require `application/json` here, it will depend on the use case... but it's ok

plugins/BEdita/Core/src/Command/ProjectModelCommand.php
stefanorosanelli: Right, but we could probably skip it and delegate it to the use or deploy process..

What do you think?

batopa: We could introduce an option `--clear-cache` (default `false`?).
I think it can be useful to clear cache if you want to launch by hand this command.

stefanorosanelli: ok, let's do it with an option

batopa (approved):

Issue #1591, opened 2019-02-18
Locked objects contraints


stefanorosanelli:
Objects with `locked` == `true` should have this behavior:

* not `trashable` - put in trashcan not allowed
* `uname` and `status` are not changeable

In all previous cases a `403 Forbidden` error should be used as response.

PR #1800, opened 2021-06-23
Locked constraint logic

stefanorosanelli:
This PR solves #1591 

Upon `ObjectsTable::beforeSave()` an additional check is performed: if `locked` is true and `uname`, `status` or `deleted` have been changed (checked via `$entity->isDirty()`) a `403 Forbidden` error is raised.

Other changes:
* `UniqueNameBehavior::uniqueName()` has been modified => `$entity->isDirty('uname')` always returned **true** even if `uname` property was unchanged and formally correct.
* many tests have been changed because one of the most used object in test fixtures - document with `id` 2 -  has `locked` set to `true` and some operations are now forbidden...

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
@@ -208,7 +208,7 @@ class ObjectsTable extends Table
      */
     protected function checkLocked(EntityInterface $entity): void
     {
-        if (empty($entity->get('locked'))) {
+        if (empty($entity->get('locked')) || $entity->isDirty('locked')) {
             return;
         }
         if ($entity->isDirty('status') || $entity->isDirty('uname') || $entity->isDirty('deleted')) {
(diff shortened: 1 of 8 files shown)


plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -196,6 +198,24 @@ protected function checkLangTag(EntityInterface $entity)
         }
     }
 
+    /**
+     * Check `locked` attribute.
+     * If `locked` is true `status`, `uname` and `deleted` cannot be changed.
+     *
+     * @param \Cake\Datasource\EntityInterface $entity Entity being saved.
+     * @return void
+     * @throws \Cake\Http\Exception\BadRequestException
```
batopa:
```suggestion
     * @throws \Cake\Http\Exception\ForbiddenException
```

batopa (approved):

Issue #1191, opened 2017-04-10
Actions to lock/unlock objects and roles, block/unblock users


stefanorosanelli:
`objects.locked` and `roles.unchangeable` and `users.blocked` are special meta properties that cannot normally be changed

only user with `administrator` role shoudl be able to apply changes to those fields with a special action.
Endpoint could be something like `/objects/{id}/lock` or could be PATCH action on a special property like
```json
{
"data": {
        "id": 123,
        "type": "users",
        "attributes": {
            "locked" : "true",
        }
    }
}
```

or

```json
{
"data": {
        "id": 123,
        "type": "users",
        "meta": {
            "locked" : "true",
        }
    }
}
```

PR #1802, opened 2021-07-01
Lock/unlock actions

stefanorosanelli:
This PR solves #1191 

With this PR users with `admin` role are allowed to change two metadata (for now):
 * `locked` common to all object types - see the logic implemented in #1800 
 * `blocked` flag for `Users`

This change can be performed via regular `PATCH` using `meta` section like this:

```
PATCH /users/123
{
"data": {
        "id": 123,
        "type": "users",
        "meta": {
            "locked" : "true",
        }
    }
}
```

### DIFF
--- a/plugins/BEdita/API/src/Controller/ObjectsController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectsController.php
@@ -270,7 +270,7 @@ class ObjectsController extends ResourcesController
             $action = new SaveEntityAction(['table' => $this->Table, 'objectType' => $this->objectType]);
 
             $data = (array)$this->request->getData();
-            $entityOptions = $this->patchEntityOptions($data);
+            $entityOptions = $this->saveEntityOptions($data);
             $entity = $action(compact('entity', 'data', 'entityOptions'));
         }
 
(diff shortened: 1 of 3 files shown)


plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -67,6 +68,13 @@ class ObjectsController extends ResourcesController
         ],
     ];
 
+    /**
+     * Meta properties accessible for admins
+     *
+     * @var array
+     */
+    protected const ADMIN_META_ACCESSIBLE = ['blocked', 'locked'];
```
batopa: [minor] Since `blocked` is a users flag maybe it's better to define `['locked']` overwriting the constant in `UsersController`.

plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
```diff
@@ -2744,4 +2744,79 @@ public function testPrepareFilter(): void
...
+        ];
+    }
+
+    /**
+     * Test `saveEntityOptions()`
+     *
+     * @param bool $expected Expected result
+     * @param string $id Test object ID
+     * @param array $meta Meta data
+     * @param array $user USer data
```
batopa:
```suggestion
     * @param array $user User data
```

plugins/BEdita/API/src/Controller/ObjectsController.php
stefanorosanelli: Done

PR #1810, opened 2021-07-14
Upgrade jQuery

fquffio:
This PR updates jQuery and jQuery UI to the latest available versions at the time of this writing: 3.6.0 and 1.12.1 respectively.

This change is aimed to upgrade the libraries to versions that are unaffected by any known vulnerability.
This PR also removes a floating-thead plugin, which can nowadays be achieved using `position: sticky` CSS rule, and adds an `autocomplete="off"`, which is ignored by all modern browsers, but "may cause problems obtaining PCI compliance".

These vulnerabilities have obviously been highlighted by a semi-automated tool without actual evaluation onf the impact on this specific application, but the CVEs are there and upgrading to a newer version isn't a bad idea generally speaking, so…


didoda:
obsolete

PR #1823, opened 2021-09-10
Fix/v4/generate unique uname

le0m:
This PR changes `UniqueNameBehavior`'s behavior when the user has explicitly set an uname. Previously the behavior was to fallback to `title` (or configured) field and append a random suffix, now it preserves the uname value set by the user and appends a random suffix to that.

Furthermore, for `Folders` objects the `UniqueNameBehavior` was prevented from working because of a _validation_ rule set on uname field (see edit to `ObjectsValidator`). This _validation_ rule has been removed, and now the behavior runs before [rules](https://book.cakephp.org/3/en/orm/validation.html#applying-application-rules) are evaluated (the `FoldersTable` has a rule for uniqueness of the uname field).

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
@@ -30,11 +30,12 @@ class ObjectsControllerTest extends IntegrationTestCase
      * @var array
      */
     public $fixtures = [
-        'plugin.BEdita/Core.Locations',
-        'plugin.BEdita/Core.ObjectRelations',
-        'plugin.BEdita/Core.Streams',
         'plugin.BEdita/Core.DateRanges',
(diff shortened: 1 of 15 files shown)


plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
```diff
@@ -1208,41 +1208,6 @@ public function testEditConflict()
...
 
-    /**
-     * Test edit method with invalid data.
-     *
-     * @return void
-     *
-     * @covers ::resource()
-     * @covers ::initialize()
-     */
-    public function testEditInvalid()
```
stefanorosanelli, line 1219: I'd like to keep this test - maybe with another case causing a `400` response (like a user with an existing `email`?) - without usgin `uname`

Also keep the 404 response on PATCH, even if trivial

stefanorosanelli (approved):

PR #1825, opened 2021-09-21
Fix: UpdateRelatedObjectsActions::prepareData, on multiple relatedEntities

didoda:
This PR fixes a bug introduced with https://github.com/bedita/bedita/pull/1727.

# buggy behavior

When `$data['relatedEntities']` is an array of more than 1 element, an exception is thrown in `plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php:68` (`$relatedEntities` is considered as an entity, but it's an array).

# expected behavior

When `$data['relatedEntities']` is an array of more than 1 element, no errors in `prepareData`, skip `set('_join_data', ...)`.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
@@ -63,7 +63,7 @@ abstract class UpdateRelatedObjectsAction extends UpdateAssociatedAction
         if (is_array($relatedEntities) && count($relatedEntities) === 1) {
             $relatedEntities = reset($relatedEntities);
         }
-        if (!empty($relatedEntities)) {
+        if (!empty($relatedEntities) && $relatedEntities instanceof \Cake\Datasource\EntityInterface) {
             /** @var EntityInterface $relatedEntities */
             $joinData = (array)$relatedEntities->get('_joinData');
             // set join data properties in Tree entity, on empty array no properties are set
--- a/plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
@@ -58,7 +58,7 @@ abstract class UpdateRelatedObjectsAction extends UpdateAssociatedAction
 
         $relatedEntities = $data['relatedEntities'];
         if (is_array($relatedEntities) && count($relatedEntities) > 1) {
-            throw new BadRequestException('Parents association for folders allows at most one related entity');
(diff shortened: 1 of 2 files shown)


stefanorosanelli (approved):
Seems ok, a test case could be useful (but not mandatory)

plugins/BEdita/Core/tests/TestCase/Model/Action/AddRelatedObjectsActionTest.php
```diff
@@ -110,6 +110,26 @@ public function invocationProvider()
                     ],
                 ],
             ],
+            'update, more than one related entities' => [
+                [2, 3],
+                'Profiles',
```
batopa: :thinking: I think the test passes without the modify too.

`prepareData()` immediately return if

```php
if (empty($data['entity']) || !($data['entity'] instanceof Folder) || $this->Association->getName() !== 'Parents') {
    $this->setupPriority($data);
    return $data;
}
```

So I think the test should be done with `Folders` as main object and `parents` as relation.
But in this way the save on folders should be fail due to validation rule https://github.com/bedita/bedita/blob/4-cactus/plugins/BEdita/Core/src/Model/Table/FoldersTable.php#L82

If the save works we have a problem. In this case we should throw an Exception if `$relatedEntities` is an array with more than one value.

didoda: changing this as suggested, an exception is thrown (as you said, @batopa ).
```
BEdita\Core\Test\TestCase\Model\Action\AddRelatedObjectsActionTest::testInvocation with data set "update, more than one related entities" (array(2, 3), 'Folders', 'parents', 4, array(array(2, 1, null), array(1, 1, array('value'))))
Cake\Database\Exception: Cannot describe trees. It has 0 columns.
```

didoda: I changed this part, to throw an exception on "more than one parent provided", for folders.

plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
```diff
@@ -55,11 +56,15 @@ protected function prepareData(array $data)
             return $data;
         }
 
+        $relatedEntities = $data['relatedEntities'];
+        if (is_array($relatedEntities) && count($relatedEntities) > 1) {
+            throw new BadRequestException('Parents association for folders allows at most one related entity');
```
batopa: We use localized string with domain `bedita`

```php
__d('bedita', 'string to translate')
```

batopa (changes requested):

batopa (approved):

PR #1827, opened 2021-09-24
get /model/schema/:type, all relations

didoda:
This PR fixes an unexpected behavior on `get /model/schema/:type, when relations use inheritance (abstract objects or "parent" objects references). 

The problem cause resides in the core entity `ObjectType`, method `getSchema`.

# buggy behavior

In the response, `relations` contains only "direct" relations, and not inherited ones.

# expected behavior

In the response, `relations` contains all relations (direct and inherited).

### DIFF
--- a/plugins/BEdita/API/tests/TestConstants.php
+++ b/plugins/BEdita/API/tests/TestConstants.php
@@ -27,7 +27,7 @@ class TestConstants
         'applications' => '2244706479',
         'documents' => '4059696127',
         'events' => '1528552691',
-        'files' => '290039203',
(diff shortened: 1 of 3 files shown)


fquffio (approved):

fquffio:
If it isn't too much of a burden, I'd add a unit test specifically targeting this behavior to prevent any future regression.

Right now, tests would only fail due to the "schema revision" that is a bit of an opaque value, and one might be tempted to change the constant value without checking for any difference…

stefanorosanelli (approved):

PR #1830, opened 2021-10-15
Streams with `private` URL

stefanorosanelli:
This PR introduces `private` URL for Streams. Use case: files, usually with sensitive data, that cannot have a public accessible URL. 

 * a new `streams.private_url` boolean DB field and property has been added, with `false` as default value and not accessible/modifiable, to indicate if the stream URL can be accessed or not
 * private URL for streams can only be defined upon upload using a `private_url` query string like `POST /streams/upload/{filename}?private_url=true` or `POST /files/upload/{filename}?private_url=true` 
 * to retrieve file content a stream download route has been introduced => `GET /streams/download/{uuid}`
 * stream file visibility is changed to 'private` via Flysystem `MountManager` if `private_url` is `true` -> this will work best on storage systems like S3 where `private` visibility is actually implemented


plugins/BEdita/API/src/Controller/Component/UploadComponent.php
```diff
@@ -76,6 +76,7 @@ public function upload($fileName, int $objectId = null): EntityInterface
             'contents' => $request->getBody(),
         ];
         $entity->set('object_id', $objectId);
+        $entity->set('private_url', (bool)$request->getQuery('private_url', false));
```
batopa: Since `(bool)'false'` return `true`, a better way to cast the query string as boolean is

```php
$entity->set('private_url', filter_var($request->getQuery('private_url', false), FILTER_VALIDATE_BOOLEAN));
```

batopa (changes requested):

didoda (approved):

batopa (approved):

PR #1833, opened 2021-10-26
Add association categories to folders

didoda:
This PR provides a migration to add association `Categories` to core model `Folders`.


stefanorosanelli:
Not needed anymore since we will use `Categories` tree feature to group categories.

PR #1840, opened 2021-11-25
`password_modified` read-only property

stefanorosanelli:
This PR adds a new column and property `users.password_modified` to have the timestamp of the last password modification. 
This is a readonly property that will appear in `meta` section for `users`. Value is automatically set upon `password` create or update.


didoda (approved):
OK for me

batopa (approved):

PR #1844, opened 2021-12-09
Improve OAuth2 check

stefanorosanelli:
This PR improves OAuth2 external provider check.

Scenario: upon signup or login via OAuth2 external providers a check on `access_token` validity is performed. An URL of the auth provider using the access token is called to get some user data. Then a match between the provider user ID and the same provider user ID in the original request is done to validate the token.

Most of the providers use an `Authorization Bearer {accessToken}` header in this OAuth2 requests, but some (like Facebook) use the access token via query string. A new `OAuth2` utility class has been added to perform this check request in a configurable way.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
@@ -18,11 +18,11 @@ use BEdita\Core\Model\Entity\User;
 use BEdita\Core\Model\Table\RolesTable;
 use BEdita\Core\Model\Validation\Validation;
 use BEdita\Core\Utility\LoggedUser;
+use BEdita\Core\Utility\OAuth2;
 use Cake\Core\Configure;
 use Cake\Event\Event;
 use Cake\Event\EventDispatcherTrait;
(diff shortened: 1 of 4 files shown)


plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -356,14 +361,13 @@ protected function checkExternalAuth(array $data)
...
      * @return array Response from an OAuth2 provider
      * @codeCoverageIgnore
      */
-    protected function getOAuth2Response(string $url, string $accessToken): array
+    protected function getOAuth2Response(string $url, string $accessToken, array $options = []): array
     {
-        $response = (new Client())->get($url, [], ['headers' => ['Authorization' => 'Bearer ' . $accessToken]]);
-
-        return (array)$response->getJson();
+        return (new OAuth2())->response($url, $accessToken, $options);
```
batopa, line 370: Myabe I would pass the configuration in the class constructor, but anyway it's ok for me

batopa (approved):

PR #1847, opened 2021-12-23
Tags lazy loading in filters

batopa:
With this PR the tags lazy loading in filters is implemented.

When the select is opened the tags are loaded via ajax. Pagination of results is used implementing infinite scroll.
Searching tag will filter results.

### DIFF
--- a/bedita-app/app_controller.php
+++ b/bedita-app/app_controller.php
@@ -1232,9 +1232,6 @@ abstract class ModulesController extends AppController {
             }
         }
 
-        // get Tags
-        $listTags = ClassRegistry::init("Category")->getTags(array("cloud" => false));
-
         // template data
         $this->set('tree', $tree);
         $this->set('sectionSel',$sectionSel);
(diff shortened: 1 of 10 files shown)


bedita-app/app_controller.php
```diff
@@ -1232,9 +1232,6 @@ protected function paginatedList($id, $filter, $order, $dir, $page, $dim) {
             }
         }
 
-        // get Tags
-        $listTags = ClassRegistry::init("Category")->getTags(array("cloud" => false));
-
         // template data
```
didoda, line 1235: I'd remove or deprecate `Category::getTags`

bedita-app/webroot/js/beditaUI.js
```diff
@@ -733,6 +733,59 @@ $(document).ready(function(){
                 }
             }
         });
+
+    $('input[name="filter[tag]"]').select2({
+        dropdownAutoWidth: true,
+        allowClear: true,
+        initSelection: function(element, callback) {
+            var id = $(element).val();
+            if (id !== '') {
```
didoda: or... `if (!id) { return; }`

didoda (approved):

bedita-app/app_controller.php
batopa: It is used elsewhere.

PR #1854, opened 2022-01-17
fix: use DateTimeInterface and enhance checks

nicolocarpignoli:
On `DateRangesTable`, use `DateTimeInterface` instead of `Time` objects and enhance checks.


plugins/BEdita/Core/src/Model/Table/DateRangesTable.php
```diff
@@ -180,16 +181,17 @@ protected function findDateRanges(Query $query, array $options)
     /**
      * Create Time object from $time string
      *
-     * @param string $time Input time.
-     * @return null|\Cake\I18n\Time
+     * @param string|\DateTimeInterface $time Input time.
+     * @return null|\DateTimeInterface
```
fquffio:
```suggestion
     * @param \DateTimeInterface|string|null $time Input time.
     * @return \DateTimeInterface|null
```

fquffio (approved):

PR #1856, opened 2022-01-18
feat: add command to refresh streams metadata in database

le0m:
This PR adds a command to `StreamsShell` to refresh streams metadata in database (file size, width, height, file hashes and EXIF metadata).

Usage: `bin/cake streams refresh_metadata [--force]`

By default the command refreshes only streams that have empty `file_size`, `width` or `height`; to refresh all streams use `--force` flag.


plugins/BEdita/Core/src/Model/Entity/Stream.php
```diff
@@ -289,9 +296,24 @@ protected function readFileMetadata($resource): void
...
+        // Filter non-standard sections
+        $exif = array_intersect_key($exif, array_flip(static::EXIF_SECTIONS));
+        // Filter undefined tags, which may break JSON encoding
+        $exif = array_map(function ($section) {
+            return array_filter($section, function ($key) {
+                return strpos($key, 'UndefinedTag:') === false;
+            }, ARRAY_FILTER_USE_KEY);
+        }, $exif);
+        // Check if data is JSON-encodable
+        if (json_decode(json_encode($exif)) === null) {
```
fquffio: To check if `$exif` can be successfully JSON-encoded, it should be sufficient to check if `json_encode()` returns `false`:

```suggestion
        if (json_encode($exif) === false) {
```

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -77,4 +93,83 @@ public function removeOrphans()
...
+                $this->Streams->aliasField('width') . ' IS' => null,
+                $this->Streams->aliasField('height') . ' IS' => null,
+            ],
+        ];
+        $force = (bool)$this->param('force');
+        if ($force) {
+            $conditions = [];
+        }
+
+        $count = $this->Streams->find('all', compact('conditions'))->count();
```
fquffio:
```suggestion
        $query = $this->Streams->find('all');
        if ((bool)$this->param('force') === false) {
            $query = $query->where(function (QueryExpression $exp): QueryExpression {
                return $exp->_or(function (QueryExpression $exp): QueryExpression {
                    return $exp
                        ->isNull($this->Streams->aliasField('file_size'))
                        ->isNull($this->Streams->aliasField('width'))
                        ->isNull($this->Streams->aliasField('height'));
                });
            });
        }

        $count = $query->count();
```

Then, the same `$query` object can be safely passed to the `streamsGenerator` method, which can clone it and use to iterate over all streams.

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -77,4 +93,83 @@ public function removeOrphans()
...
+        $this->info(sprintf('Checking %d streams', $count));
+
+        // Set custom error handler because Stream calls some PHP functions that use old error handling
+        set_error_handler(
+            function (int $code, string $message, string $filename, int $lineNumber): void {
+                throw new \ErrorException($message, $code, LOG_ERR, $filename, $lineNumber);
+            }
+        );
```
fquffio: It is very dangerous to override the error handler at this point. I'd frankly rather have it as close as possible to the call to `exif_read_data`, to avoid handling more errors than we actually want to catch… Ideally, I'd place this in the `Stream` entity itself.

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -77,4 +93,83 @@ public function removeOrphans()
...
+        foreach ($this->streamsGenerator($conditions) as $stream) {
+            $this->updateStreamMetadata($stream);
+        }
+        restore_error_handler();
```
fquffio: This is rather unsafe. What we want is to restore the previous error handler in any scenario, at this point. But the execution might be abruptly halted due to an exception, for instance, which would cause this to be skipped.

The pattern should be:

```php
set_error_handler($myHandler);
try {
    // the code which triggers old-style errors
    // as little code as possible should be placed here
} finally {
    restore_error_handler();
}

File: plugins/BEdita/Core/src/Shell/StreamsShell.php [hunk 5]
```diff
@@ -77,4 +93,83 @@ public function removeOrphans()
...
+    protected function streamsGenerator(array $conditions = [], int $pageSize = 100): \Generator
+    {
+        $query = $this->Streams->find('all', compact('conditions'));
+        $pageCount = ceil($query->count() / $pageSize);
+
+        for ($page = 1; $page <= $pageCount; $page++) {
+            yield from $query->page($page, $pageSize)->all();
+            $this->info(sprintf('  processed %d streams', min($page * $pageSize, $query->count())));
+        }
+    }
```
fquffio: This skips a lot of results, because after retrieving page 1 with streams [A B C], if the streams returned are processed successfully, the same query does not return [A B C D E F G H I] anymore, but only [D E F G H I]. So, after the first page has been processed, page 2 becomes [G H I], and [D E F] are skipped entirely.

If we refactor this method to accept a `Query` object rather than an array of conditions, we could:

```suggestion
    protected function streamsGenerator(Query $query, int $pageSize = 100): \Generator
    {
        // Although `id` is not a monotonically increasing field, we will at most skip the streams that are created
        // AFTER we launch the script, and whose UUID is lexicographically less than the ID we are currently
        // checking — but we still cover all streams created before our script starts!
        $query = $query->sortAsc($this->Streams->aliasField('id'));
        $q = clone $query;
        do {
            $results = $q->all();
            if ($results->isEmpty()) {
                break;
            }

            yield from $results;

            $last = $results->last();
            $q = clone $query;
            $q = $q->where(function (QueryExpression $exp) use ($last): QueryExpression {
                return $exp->gt($this->Streams->aliasField('id'), $last->id);
            });
        }
    }
```

fquffio (changes requested):

plugins/BEdita/Core/src/Shell/StreamsShell.php
le0m: I agree that the approach taken does not work, but we're hitting another problem: `streams` does not have an `id` column, only `uuid` which is a random string.

We could ignore all of it and yield directly from `->all()`:
```php
    protected function streamsGenerator(Query $query, int $pageSize = 100): \Generator
    {
        yield from $query->all();
    }
```

le0m: I committed the changes, could you look at it again please?

plugins/BEdita/Core/src/Model/Entity/Stream.php
```diff
@@ -276,11 +284,13 @@ protected function readFileMetadata($resource): void
     {
         if (preg_match('/image\//', $this->mime_type) && function_exists('getimagesizefromstring')) {
             rewind($resource);
-            $size = getimagesizefromstring(stream_get_contents($resource));
-
-            if (!empty($size)) {
-                $this->width = $size[0];
-                $this->height = $size[1];
+            $content = stream_get_contents($resource);
```
fquffio, line 297: This loads the entire image contents in memory as a string. For very large images, this may get the process killed.
Furthermore, it is really done only to obtain image size.

I'd strongly advise revisiting this part. EXIF already returns tons of metadata which includes image size, so perhaps it'd be wiser to use that instead of this approach. It would also have the benefit of reading the stream only once rather than twice.

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -77,4 +95,100 @@ public function removeOrphans()
...
+                    return $exp
+                        ->isNull($this->Streams->aliasField('file_size'))
+                        ->isNull($this->Streams->aliasField('width'))
+                        ->isNull($this->Streams->aliasField('height'));
+                });
+            });
+        }
+
+        $count = $query->count();
+        $this->info(sprintf('Checking %d streams', $count));
```
fquffio: [minor] This number may change if new streams are saved while the script is running, or if streams are deleted, fixed by other means, …

I'd argue that this count may be useless, but if it is actually desired to give an idea of how much data is to be processed, at least I'd point out explicitly that this is an approximation:

```suggestion
        $this->info(sprintf('Approximately %d streams to be processed', $count));
```

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -77,4 +95,100 @@ public function removeOrphans()
...
+            // Read current file's content...
+            $content = $stream->contents;
+            if ($content === null) {
+                $this->warn(sprintf('  stream %s (object %d) is empty or could not be read', $stream->uuid, $stream->object_id));
+
+                return false;
+            }
+
+            // ...and write it back, triggering Stream model's methods to read metadata from file
+            $stream->contents = $content;
```
fquffio, line 149: Since the method to read exif metadata executes `rewind()` on the Stream, that stream will need to be seekable.

Streams returned by some filesystem adapters (e.g., S3) may not be seekable. However, I'm not sure whether the `Stream::_setContents()` accessor makes a copy of the incoming stream to make sure it is always seekable.

If it does NOT, it is our responsibility here to make such copy:

```php
$copy = fopen('php://temp', 'r+');
stream_copy_to_stream($contents, $copy);
fclose($contents); // Free resource.
$stream->contents = $copy;
```

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -77,4 +95,100 @@ public function removeOrphans()
...
+            if ($stream->isDirty() && !$this->Streams->save($stream)) {
```
fquffio: Is it necessary to check whether `$stream->isDirty()` here?

fquffio:
For testing, I'd rather simply run a `$this->Streams->updateAll(['width' => null])`, execute the command, and check if the `width` has been re-populated, with some assertions on command output.

It may also be a good idea to test whether pagination works (set limit to 1) and check that command is a no-op without the `--force` flag with media that already have their metadata populated.

plugins/BEdita/Core/src/Shell/StreamsShell.php
le0m: The setter calls `Stream::createStream()`, which handles rewinding the original resource (if possible) and makes a copy in a new stream

plugins/BEdita/Core/src/Shell/StreamsShell.php
le0m: Completely unnecessary since `Stream::_setContents()` always sets a bunch of properties as "dirty".

plugins/BEdita/Core/src/Model/Entity/Stream.php
le0m: I did some test with various types of image and found out that png, bmp and webp all errored out when using `exif_read_data`.
It seems like only jpeg/jpg and tiff images have EXIF data, so I changed the code to use `getimagesizefromstring` or `exif_read_data` depending on the MIME type.

nicolocarpignoli:
@fquffio tests added

plugins/BEdita/Core/src/Shell/StreamsShell.php
```diff
@@ -76,4 +95,100 @@ public function removeOrphans()
...
+                return false;
+            }
+
+            // ...and write it back, triggering Stream model's methods to read metadata from file
+            $stream->contents = $content;
+            if (!$this->Streams->save($stream)) {
+                $this->err(sprintf('  error updating stream %s (object %d): %s', $stream->uuid, $stream->object_id, print_r($stream->getErrors(), true)));
+
+                return false;
+            }
```
fquffio: Since we're already catching errors, I'd use `saveOrFail` to avoid this condition:

```suggestion
            $this->Streams->saveOrFail($stream));
```

plugins/BEdita/Core/tests/TestCase/Shell/StreamsShellTest.php
```diff
@@ -64,6 +64,49 @@ public function removeOrphansProvider()
...
 
+    /**
+     * Test `refreshMetadata` method
+     *
+     * @return void
+     * @covers ::refreshMetadata()
+     * @covers ::updateStreamMetadata()
+     * @covers ::streamsGenerator()
+     */
+    public function testRefreshMetadata()
```
fquffio:
```suggestion
    public function testRefreshMetadata(): void
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -51,19 +51,19 @@ class StreamTest extends TestCase
     ];
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
```
fquffio: This change MUST be reverted in order to work with the latest PHP CS version:

```suggestion
     * @inheritDoc
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -51,19 +51,19 @@ class StreamTest extends TestCase
...
+    public function setUp()
     {
         parent::setUp();
         $this->filesystemSetup();
         $this->Streams = TableRegistry::getTableLocator()->get('Streams');
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
```
fquffio: This change MUST be reverted in order to work with the latest PHP CS version:

```suggestion
     * @inheritDoc
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -129,12 +129,13 @@ public function filesystemPathProvider()
      * @param string $filesystem Filesystem name.
      * @param int $subLevels Number of sub-levels.
      * @return void
+     *
```
fquffio: This change MUST be reverted in order to work with the latest PHP CS version:

```suggestion
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -51,19 +51,19 @@ class StreamTest extends TestCase
     ];
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
      */
-    public function setUp(): void
+    public function setUp()
```
fquffio: This change MUST be reverted in order to work with the latest PHUnit version:

```suggestion
    public function setUp(): void
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -51,19 +51,19 @@ class StreamTest extends TestCase
...
         $this->filesystemSetup();
         $this->Streams = TableRegistry::getTableLocator()->get('Streams');
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
      */
-    public function tearDown(): void
+    public function tearDown()
```
fquffio: This change MUST be reverted in order to work with the latest PHUnit version:

```suggestion
    public function tearDown(): void
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -146,6 +147,7 @@ public function testFilesystemPath($expected, array $data, $filesystem = 'defaul
      * Test getter of contents.
      *
      * @return void
+     *
```
fquffio: This change MUST be reverted in order to work with the latest PHP CS version:

```suggestion
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -164,6 +166,7 @@ public function testGetContents()
      * Test getter of contents for a missing file.
      *
      * @return void
+     *
```
fquffio: This change MUST be reverted in order to work with the latest PHP CS version:

```suggestion
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -179,11 +182,12 @@ public function testGetContentsFileNotReadable()
      * Test getter of contents for a file not uploaded.
      *
      * @return void
+     *
```
fquffio: This change MUST be reverted in order to work with the latest PHP CS version:

```suggestion
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -261,6 +265,7 @@ public function setContentsProvider()
      * @param \Exception|string $expected Expected stream contents.
      * @param mixed $contents Contents.
      * @return void
+     *
```
fquffio: This change MUST be reverted in order to work with the latest PHP CS version:

```suggestion
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -309,6 +314,7 @@ public function getUrlProvider()
      * @param string|null $expected Expected result.
      * @param string $uuid UUID of stream to test.
      * @return void
+     *
```
fquffio: This change MUST be reverted in order to work with the latest PHP CS version:

```suggestion
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -328,6 +334,7 @@ public function testGetUrl($expected, $uuid)
      * Test URL getter with private url.
      *
      * @return void
+     *
```
fquffio: This change MUST be reverted in order to work with the latest PHP CS version:

```suggestion
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -339,8 +346,7 @@ public function testGetUrlPrivate()
 
     /**
      * Read data from image if is possible
-     *
-     * @param Stream $stream stream entity
+     * @param EntityStream $stream stream entity
```
fquffio: This change MUST be reverted in order to work with the latest PHP CS version:
```suggestion
     *
     * @param \BEdita\Core\Model\Entity\Stream $stream Stream entity.
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -179,11 +182,12 @@ public function testGetContentsFileNotReadable()
      * Test getter of contents for a file not uploaded.
      *
      * @return void
+     *
      * @covers ::_getContents()
      */
     public function testGetContentsNotUploaded()
     {
-        $stream = $this->Streams->newEntity([]);
+        $stream = $this->Streams->newEntity();
```
fquffio, line 186:
```suggestion
        $stream = $this->Streams->newEmptyEntity();
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -273,7 +278,7 @@ public function testSetContents($expected, $contents)
             $this->expectExceptionMessage($expected->getMessage());
         }
 
-        $stream = $this->Streams->newEntity([]);
+        $stream = $this->Streams->newEntity();
```
fquffio, line 276:
```suggestion
        $stream = $this->Streams->newEmptyEntity();
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -377,13 +383,13 @@ public function testReadFileMetadata()
         $imageTest = new Stream($path . '/a4fbe302-3d5b-4774-a9df-18598def690e-image-metadata.jpeg', 'r');
         $gifTest = new Stream($path . '/6aceb0eb-bd30-4f60-ac74-273083b921b6-bedita-logo-gray.gif', 'r');
 
-        $stream = $this->Streams->newEntity([]);
+        $stream = $this->Streams->newEntity();
```
fquffio, line 380:
```suggestion
        $stream = $this->Streams->newEmptyEntity();
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -377,13 +383,13 @@ public function testReadFileMetadata()
...
         $stream->mime_type = 'image/jpeg';
         $stream->contents = $imageTest;
 
         $this->readDataFromImage($stream);
         // mime type not allowed
-        $stream = $this->Streams->newEntity([]);
+        $stream = $this->Streams->newEntity();
```
fquffio, line 386:
```suggestion
        $stream = $this->Streams->newEmptyEntity();
```

plugins/BEdita/Core/tests/TestCase/Model/Entity/StreamTest.php
```diff
@@ -399,7 +405,7 @@ public function testFailedReadFileMetadata()
     {
         $path = Configure::read('Filesystem.default.path');
         $gifTest = new Stream($path . '/6aceb0eb-bd30-4f60-ac74-273083b921b6-bedita-logo-gray.gif', 'r');
-        $stream = $this->Streams->newEntity([]);
+        $stream = $this->Streams->newEntity();
```
fquffio, line 402:
```suggestion
        $stream = $this->Streams->newEmptyEntity();
```

fquffio (changes requested):

PR #1863, opened 2022-02-08
Mark join entity as new when it is a new association for source entity

batopa:
This PR fixes an issue we have trying to add related objects loaded from another entity.

Before this PR doing

```php
$Documents = \Cake\ORM\TableRegistry::get('Documents');

// assuming to have an existing document with id = 5 with "attach" relation
$doc = $Documents->get(5);
$Documents->loadInto($doc, ['Attach']);
$relatedEntities = $doc->get('attach');

$entity = $Documents->newEntity(['title' => 'My title']);
$entity = $Documents->save($entity);

$action = new \BEdita\Core\Model\Action\AddRelatedObjectsAction(compact('association'));
$action(compact('entity', 'relatedEntities'));
```

save `$entity` with no errors but the related entities were not linked to `$entity`.

The reason was in the way we hydrate related entities in source entity. The join entity is updated with the right primary keys (`left_id`, `right_id`, `relation_id`), but since that join entity was loaded from another entity it results already persisted and not saved.

Now it is marked as new to ensure to store it.

### DIFF
--- a/plugins/BEdita/Core/tests/TestCase/Model/Action/AddRelatedObjectsActionTest.php
+++ b/plugins/BEdita/Core/tests/TestCase/Model/Action/AddRelatedObjectsActionTest.php
@@ -54,7 +54,7 @@ class AddRelatedObjectsActionTest extends TestCase
     /**
      * {@inheritDoc}
      */
-    public function setUp()
+    public function setUp(): void
     {
         parent::setUp();
 
@@ -64,7 +64,7 @@ class AddRelatedObjectsActionTest extends TestCase
     /**
      * {@inheritDoc}
      */
-    public function tearDown()
+    public function tearDown(): void
     {
         parent::tearDown();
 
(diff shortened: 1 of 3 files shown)


fquffio (approved):

plugins/BEdita/Core/tests/TestCase/Model/Action/AddRelatedObjectsActionTest.php
```diff
@@ -47,8 +48,29 @@ class AddRelatedObjectsActionTest extends TestCase
         'plugin.BEdita/Core.Users',
         'plugin.BEdita/Core.Roles',
         'plugin.BEdita/Core.RolesUsers',
+        'plugin.BEdita/Core.History',
     ];
 
+    /**
+     * {@inheritDoc}
+     */
+    public function setUp()
```
stefanorosanelli: [super minor] `: void` return type. also in `tearDown()`

stefanorosanelli (approved):

PR #1867, opened 2022-02-17
Draft: Feat/exact search

nicolocarpignoli:
Introduce exact search if "exact" and "string" params are provided.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
@@ -118,7 +118,13 @@ class SearchableBehavior extends Behavior
      */
     public function findQuery(Query $query, array $options)
     {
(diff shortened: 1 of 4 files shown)


fquffio:
Just a quick note: from an API's perspective, the existing functionality is preserved and expanded.

The following requests are now possible:

- `GET /objects?filter[query]=foo+bar` — existing functionality
- `GET /objects?filter[query][string]=foo+bar&filter[query][exact]=0` — same as existing functionality (`exact` is off by default, and may be omitted)
- `GET /objects?filter[query][string]=foo+bar&filter[query][exact]=1` — expanded functionality: the query string is searched exactly as provided, without searching word by word

The "exact" search performs an exact match with the whole query string, rather than splitting words and searching them one by one.

For instance, when searching "foo bar", a document whose title is "some bar are foo" will not match.

PR #1872, opened 2022-02-22
Cake 4 Core compatibility changes

stefanorosanelli:
In this PR some CakePHP 4 compatibility changes are applied to `BEdita/Core` plugin 

 * in `bootstrap.php` `I18n::translators()->registerLoader()` must return a `Package` object, not a callable
 * Query: call `all()` to retrieve results first when resultset methods are called directly on query instances
 * `unsetProperty($item)` entity method has been replaced by `unset($item)` - use `unset($entity[$item])` 
 * use `newEntity([])` instead of `newEntity()` table method since signature has changed, at least one argument is needed
 * missing return types are added to various methods
 * use `setNew($bool)` instead of `isNew($bool)` in entities
 * `QueryExpression` methods like `and_()` and `or_()` are deprecated => use `and()` or `or()` instead
 * don't use special chars like `*` in cache key strings => use `config_any_any` instead of  `config_*_*`

### DIFF
--- a/plugins/BEdita/Core/src/Command/FixHistoryCommand.php
+++ b/plugins/BEdita/Core/src/Command/FixHistoryCommand.php
@@ -96,7 +96,7 @@ class FixHistoryCommand extends Command
      *
      * @return void
      */
-    public function initialize()
+    public function initialize(): void
     {
         $this->History = $this->Objects->getBehavior('History')->Table;
         $application = TableRegistry::getTableLocator()->get('Applications')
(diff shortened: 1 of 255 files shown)


plugins/BEdita/Core/src/Command/FixHistoryCommand.php
```diff
@@ -227,7 +227,7 @@ protected function fixHistoryUpdate(ObjectEntity $object): void
      */
     protected function historyEntity(ObjectEntity $object): History
     {
-        $history = $this->History->newEntity();
+        $history = $this->History->newEntity([]);
```
batopa, line 231: To make happy scrutinizer:

```suggestion
       /** @var \BEdita\Core\Model\Entity\History $history */
        $history = $this->History->newEntity([]);
```

batopa (approved):

plugins/BEdita/Core/src/Command/FixHistoryCommand.php
stefanorosanelli: done

PR #1880, opened 2022-03-17
Fix/v4/thumb generator ext

paolosantarsiero:
This PR fixes converted file extension of thumb generator and add check extensions allowed.


plugins/BEdita/Core/src/Filesystem/Thumbnail/GlideGenerator.php
```diff
@@ -54,10 +56,15 @@ class GlideGenerator extends ThumbnailGenerator
      * @param \BEdita\Core\Model\Entity\Stream $stream Stream entity instance.
      * @param array $options Thumbnail options.
      * @return string
+     * @throws BadRequestException
```
fquffio: It's generally advisable to use fully-qualified class names in docblock annotations:

```suggestion
     * @throws \Cake\Http\Exception\BadRequestException
```

plugins/BEdita/Core/src/Filesystem/Thumbnail/GlideGenerator.php
```diff
@@ -54,10 +56,15 @@ class GlideGenerator extends ThumbnailGenerator
...
      */
     protected function getFilename(Stream $stream, array $options = [])
     {
         $ext = pathinfo($stream->file_name, PATHINFO_EXTENSION);
+        $fm = Hash::get($options, 'fm', 'jpg');
+        if (!in_array($fm, ['jpg', 'pjpg', 'png', 'gif', 'webp', 'avif'])) {
+            throw new BadRequestException(__d('bedita', 'Encoded the image to a specific format is not available.'));
```
fquffio: I'd include the invalid format in the exception message: it may be simpler to detect the misconfiguration:

```suggestion
            throw new BadRequestException(__d('bedita', 'Invalid thumbnail format: {0}', $fm));
```

plugins/BEdita/Core/src/Filesystem/Thumbnail/GlideGenerator.php
```diff
@@ -54,22 +56,24 @@ class GlideGenerator extends ThumbnailGenerator
...
         $ext = pathinfo($stream->file_name, PATHINFO_EXTENSION);
-        if ($ext) {
-            $ext = '.' . $ext;
+        $fm = Hash::get($options, 'fm', 'jpg');
+        if (!in_array($fm, ['jpg', 'pjpg', 'png', 'gif', 'webp', 'avif'])) {
+            throw new BadRequestException(__d('bedita', 'Invalid thumbnail format: {0}', $fm));
         }
         $filesystem = $this->getConfig('cache', 'thumbnails');
-        $base = $stream->filesystemPath($filesystem);
+        $base = str_replace('.' . $ext, '', $stream->filesystemPath($filesystem));
```
batopa: I would leave `$base` with original file extension

plugins/BEdita/Core/src/Filesystem/Thumbnail/GlideGenerator.php
```diff
@@ -156,7 +160,8 @@ public function exists(Stream $stream, array $options = [])
     public function delete(Stream $stream)
     {
         $filesystem = $this->getConfig('cache', 'thumbnails');
-        $base = $stream->filesystemPath($filesystem);
+        $ext = pathinfo($stream->file_name, PATHINFO_EXTENSION);
+        $base = str_replace('.' . $ext, '', $stream->filesystemPath($filesystem));
```
batopa: I would leave $base with original file extension

batopa (changes requested):

plugins/BEdita/Core/src/Filesystem/Thumbnail/GlideGenerator.php
```diff
@@ -54,19 +56,20 @@ class GlideGenerator extends ThumbnailGenerator
...
+     * @throws \Cake\Http\Exception\BadRequestException
      */
     protected function getFilename(Stream $stream, array $options = [])
     {
-        $ext = pathinfo($stream->file_name, PATHINFO_EXTENSION);
-        if ($ext) {
-            $ext = '.' . $ext;
+        $ext = Hash::get($options, 'fm', 'jpg'); // jpg default
+        if (!in_array($ext, ['jpg', 'pjpg', 'png', 'gif', 'webp', 'avif'])) {
+            throw new BadRequestException(__d('bedita', 'Invalid thumbnail format: {0}', $ext));
```
fquffio: Where possible, I'd avoid using these HTTP-specific exceptions outside Controllers, Middlewares and Components.

I'd rather use the SPL `\InvalidArgumentException`, which fits the purpose better in my opinion. On the other hand, this would cause a **500 Internal Server Error** if uncaught, whereas the `\Cake\Http\Exception\BadRequestException` would cause a **400 Bad Request**.

Therefore, if a **400** is desired, one would need to catch this exception in the controller action, and re-throw the desired HTTP-specific one.

fquffio (changes requested):

plugins/BEdita/Core/src/Filesystem/Thumbnail/GlideGenerator.php
stefanorosanelli: What about the new `InvalidDataException` for now? Still not optimal, but maybe we can close this issue. @batopa @fquffio

fquffio: Seems the best option to me! 👍🏼

stefanorosanelli (approved):

batopa (approved):

PR #1881, opened 2022-03-17
Reset admin status on setup

edoardocavazza:
Reset user admin status when running the setup command.

### DIFF
--- a/plugins/BEdita/Core/src/Shell/Task/SetupAdminUserTask.php
+++ b/plugins/BEdita/Core/src/Shell/Task/SetupAdminUserTask.php
@@ -113,6 +113,7 @@ class SetupAdminUserTask extends Shell
         $user->username = $this->param('admin-username');
         $user->password = $this->param('admin-password');
         $user->blocked = false;
+        $user->status = 'on';
 
         try {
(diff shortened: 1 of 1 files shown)


batopa (approved):

plugins/BEdita/Core/src/Shell/Task/SetupAdminUserTask.php
```diff
@@ -113,6 +113,7 @@ public function main()
         $user->username = $this->param('admin-username');
         $user->password = $this->param('admin-password');
         $user->blocked = false;
+        $user->status = 'on';
```
stefanorosanelli, line 116: This is normally not needed because the `admin` user with ID 1 is created `locked` and with `status` on by the migrations. 

If for some reason `status` is not `on` we should first set `locked` to `false`, then save `status` `on` and then set `locked` to `true` again.

edoardocavazza: I see. I think this was caused by a manually corrupted state, so we can ignore this change. I'll close the PR. Thanks!

PR #1884, opened 2022-03-23
Coding Style - Cake 4 compatibility

stefanorosanelli:
This PR introduces various coding style changes originated from Cake 4 coding style stricter rules that are still valid in `4.x` codebase.
The goal is to minimize changes between `4-cactus` branch and the next `5-###` Cake 4 branch and to improve code quality.


plugins/BEdita/Core/src/Model/Action/UpdateRelatedObjectsAction.php
```diff
@@ -68,7 +67,7 @@ protected function prepareData(array $data)
             $relatedEntities = reset($relatedEntities);
         }
         if (!empty($relatedEntities)) {
-            /** @var EntityInterface $relatedEntities */
+            /** @var \BEdita\Core\Model\Action\EntityInterface $relatedEntities */
```
batopa: It should be
```suggestion
            /** @var \Cake\Datasource\EntityInterface $relatedEntities */
```

plugins/BEdita/Core/src/ORM/Inheritance/Table.php
```diff
@@ -226,7 +226,7 @@ public function callFinder($type, CakeQuery $query, array $options = []): CakeQu
     /**
      * {@inheritDoc}
      *
-     * @param bool $inheritedFields Should fields from inherited tables be considered?
+     * @param bool $field Should fields from inherited tables be considered?
```
batopa: I think `$inheritedFields` is right

stefanorosanelli: Right, `phpcbf` was wrong here

batopa (approved):

PR #1886, opened 2022-03-25
Introduce InvalidDataException

stefanorosanelli:
This PR solves a problem with exception signatures in Cake 4

`BadRequestException` and other `HTTP` exceptions accept only strings as first constructor argument `$message` 
Since we are using quite often an array as first argument, specially in `BEdita/Core` a new  `InvalidDataException` has been introduced that can be safely used in `BEdita/Core`.

`ExceptionRenderer` class has also been modified to print previous exception message in `details` if present.


plugins/BEdita/Core/src/Exception/InvalidDataException.php
```diff
@@ -0,0 +1,34 @@
...
+     * {@inheritDoc}
+     *
+     * Default error code 400
+     *
+     * @codeCoverageIgnore
+     */
+    public function __construct($message, $code = 400, $previous = null)
+    {
+        parent::__construct($message, $code, $previous);
+    }
```
fquffio: Since we're now overriding the exception class, I'd take this chance to promote `details` to a proper class member:

```suggestion
class InvalidDataException extends Exception
{
    /**
     * Additional details on the cause of this exception.
     *
     * @var mixed
     */
    protected $details;

    /**
     * {@inheritDoc}
     *
     * Default error code 400
     *
     * @param mixed $details Additional details on the cause of this exception.
     *
     * @codeCoverageIgnore
     */
    public function __construct($message, $code = 400, $previous = null, $details = null)
    {
        parent::__construct($message, $code, $previous);

        $this->details = $details;
    }

    /**
     * Get exception details.
     *
     * @return mixed
     */
    public function getDetails()
    {
        return $this->details;
    }
```

I think this would be both more elegant and more robust:

1. It's only a matter of time before either Cake or PHP itself add a strict `string` type on Exceptions `$message`, whether it's the constructor argument or the return type of `Exception::getMessage()`.
2. Arrays with keys as options are reeeeeally old fashioned, they are obscure, hard to typecast and easy to mistype (detail vs details).

stefanorosanelli: I don't like the `array` signature either, but we are currently using Cake base Exception `$_attributes` array in `ExceptionRenderer` to display additional information in JSON-API response.
In Cake 5 `$_attributes` array is stll there.

I'd suggest a different approach. Textual details could be probably added to the main `string` message but structured error information may need an additional details array.
Sometimes we also need an application error code like `be_token_expired` or `be_user_exists` that should be passed using a custom exception like `ExpiredTokenException`. We always need an HTTP status code and sometimes an additional application error code.

To sum up we could adopt the solution you propose using `array` type for `$details` attribute. 
If we want to get rid of Cake `$_attributes` we should also update `ExceptionRenderer` and check every exception raised in plugins & apps. 
We have then to decide how to use app error codes like `be_token_expired` - maybe an `$appErrorCode` attribute if we don't want to use `$_attributes`.

That said I think we should implement this refactor in the new Cake 4 to avoid bad side effects in current implementations and stick with the _ugly_ array format for now.

stefanorosanelli: Another solution for this specific exception that doesn't need a change in `ExceptionRenderer` 

```php
public function __construct($message, $code = 400, $previous = null, ?array $details = null)
{
     parent::__construct($message, $code, $previous);
	
     $this->$_attributes['detail'] = $details;
}
```

stefanorosanelli: @fquffio @batopa what about a solution like this one?  

```php
class InvalidDataException extends Exception
{
    public function __construct(string $message = 'Invalid Data', ?array $details = null)
    {
        parent::__construct($message, 400);
        $this->_attributes['detail'] = $details;
    }
}
```

This way we can handle all cases except the one with `be_user_exists` application error code, that we could handle with a new `UserExistsException`. 
Also no changed in `ExceptionRenderer` are needed.
The signature could be extended with other arguments like `?Throwable $previous = null` but I prefer a cleaner and simpler signature until we have some real use cases where an extension is needed.

stefanorosanelli: @fquffio @batopa what about a solution like this one?  

```php
class InvalidDataException extends Exception
{
    public function __construct(string $message = 'Invalid Data', ?array $details = null)
    {
        parent::__construct($message, 400);
        $this->_attributes['detail'] = $details;
    }
}
```

This way we can handle all current cases except the one with `be_user_exists` application error code, that we could handle with a new `UserExistsException`. 
Also no changes in `ExceptionRenderer` are needed.
The signature could be extended with other arguments like `?Throwable $previous = null` but I prefer a cleaner and simpler signature until we have some real use cases where an extension is needed.

batopa: For me it's ok. 
Since we're working on CakePHP 4 migration,  I would keep changes limited as possible. 
A refactor could be done later we complete the migratation.

batopa (approved):

PR #1887, opened 2022-03-25
Add user_preferences field to users table

batopa:
This PR adds `user_preferences` field (as JSON) to `users` table.


plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -167,6 +168,20 @@ public function buildRules(RulesChecker $rules): RulesChecker
...
+     *
+     * @codeCoverageIgnore
+     */
+    protected function _initializeSchema(TableSchema $schema): TableSchema
+    {
+        if ($schema->getColumnType('user_preferences') !== 'json') {
+            $schema->setColumnType('user_preferences', 'json');
+        }
+
+        return $schema;
```
fquffio: Would there be any undesirable side effect in setting `json` right away?

```suggestion
        return $schema
            ->setColumnType('user_preferences', 'json');
```

fquffio (approved):

fquffio (commented):
I'd also add the `@property` annotation on the `User` entity, if possible, and I'm wondering whether it should be made `$_accessible`, or `$_hidden`… but I suspect it heavily depends on the intended use case.

stefanorosanelli (approved):

batopa:
Right for the `@property` annotation, thanks!
For the property visibility I'm undecided. As you say the use case determines how to use that field, so maybe we can leave accessible for now.

PR #1891, opened 2022-03-29
Better coverage

fquffio:
This PR uses `phpdbg` to generate coverage in place of PCov, which requires a clobber.

Additionally, Composer autoloader and authoritative classmap is dumped before running unit tests to catch usages of deprecated aliased classes, and Codecov will now wait for all jobs to complete before notifying about a coverage decrease.


didoda:
obsolete

PR #1892, opened 2022-04-01
Refactor before cake4 migration

batopa:
Another PR to prepare to CakePHP 4 migration.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Behavior/PriorityBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/PriorityBehavior.php
@@ -88,7 +88,12 @@ class PriorityBehavior extends Behavior
     {
         $conditions = [];
         foreach ($scope as $item) {
-            $conditions[$this->getTable()->aliasField($item)] = $entity->get($item);
+            $valueField = $entity->get($item);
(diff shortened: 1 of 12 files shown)


plugins/BEdita/Core/src/Model/Behavior/PriorityBehavior.php
```diff
@@ -88,7 +88,12 @@ protected function maxValue(EntityInterface $entity, string $field, array $scope
     {
         $conditions = [];
         foreach ($scope as $item) {
-            $conditions[$this->getTable()->aliasField($item)] = $entity->get($item);
+            $valueField = $entity->get($item);
```
stefanorosanelli: [minor] we could use automatic `IS NULL` creation https://book.cakephp.org/3/en/orm/query-builder.html#automatic-is-null-creation 

```php
$keyField = sprintf('%s IS', $this->getTable()->aliasField($item));
$conditions[$keyField] = $entity->get($item);
```

stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Behavior/PriorityBehavior.php
batopa: Nice!

PR #1893, opened 2022-04-08
Ensure left and right fields are update when removing a tree entity

edoardocavazza:
When deleting two (or more) child nodes from a folder, BEdita performs the deletion by removing the nodes from the parent/children association.

These nodes are already hydrated at the time of removal, with a `tree_left` and a `tree_right` that correspond to the position of the nodes in the tree. However, when the first node association is deleted, Cake does not take into account the new shifted position of the subsequent nodes, and performs updates using the outdated `tree_left` and `tree_right`. This may lead to corrupted tree state.

This PR introduces a change in the `beforeDelete` hook of the TreeBehavior to make sure that that position is up to date before the delete is performed.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Behavior/TreeBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/TreeBehavior.php
@@ -18,6 +18,7 @@ use Cake\Database\Expression\IdentifierExpression;
 use Cake\Database\Expression\QueryExpression;
 use Cake\Database\Query;
 use Cake\Datasource\EntityInterface;
+use Cake\Event\Event;
 use Cake\ORM\Behavior\TreeBehavior as CakeTreeBehavior;
 use Cake\ORM\Table;
 
@@ -42,6 +43,19 @@ class TreeBehavior extends CakeTreeBehavior
         ];
 
         parent::__construct($table, $config);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function beforeDelete(Event $event, EntityInterface $entity)
+    {
+        // ensure to use actual left and right fields
+        unset($entity[$this->getConfig('left')]);
+        unset($entity[$this->getConfig('right')]);
+        $this->_ensureFields($entity);
+
+        parent::beforeDelete($event, $entity);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Model/Behavior/TreeBehavior.php
```diff
@@ -44,6 +45,19 @@ public function __construct(Table $table, array $config = [])
...
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function beforeDelete(Event $event, EntityInterface $entity)
+    {
+        // ensure to use actual left and right fields
+        unset($entity[$this->getConfig('left')]);
+        unset($entity[$this->getConfig('right')]);
```
fquffio:
```suggestion
        unset($entity[$this->getConfig('left')], $entity[$this->getConfig('right')]);
```

plugins/BEdita/Core/tests/TestCase/Model/Behavior/TreeBehaviorTest.php
```diff
@@ -361,4 +367,28 @@ public function testCheckIntegritySiblingsOverlapping(): void
...
+            ->firstOrFail();
+
+        $this->Table->ChildCategories->unlink($parentNode, [
+            $parentNode->child_categories[0],
+            $parentNode->child_categories[2],
+        ]);
+
+        $errors = $this->Table->checkIntegrity();
+
+        $this->assertEmpty($errors);
```
fquffio:
```suggestion
        static::assertEmpty($errors);
```

fquffio (approved):

plugins/BEdita/Core/src/Model/Behavior/TreeBehavior.php
```diff
@@ -44,6 +45,18 @@ public function __construct(Table $table, array $config = [])
         parent::__construct($table, $config);
     }
 
+    /**
+     * {@inheritDoc}
```
stefanorosanelli: [superminor] it's `@inheritDoc` without `{}` in this case
` {@inheritDoc}` is used when annotations details are added

stefanorosanelli (approved):

PR #1900, opened 2022-05-09
fix: avoid changing results cardinality when using `?filter[ancestors]`

nicolocarpignoli:
This PR fixes a problem when using `?filter[ancestors]`: results cardinality is affected by multiple positions of the object on a specific tree (e.g. object is on three, different position of the tree, the API call shows the object three times on results data instead of one).

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
@@ -429,7 +429,7 @@ class ObjectsTable extends Table
                         return $exp
                             ->gt($this->TreeNodes->aliasField('tree_left'), $parentNode->get('tree_left'))
                             ->lt($this->TreeNodes->aliasField('tree_right'), $parentNode->get('tree_right'));
-                    }),
+                    })
             );
         });
     }
--- a/plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
@@ -420,15 +420,18 @@ class ObjectsTable extends Table
             ])
             ->firstOrFail();
 
-        return $query
-            ->innerJoinWith('TreeNodes', function (Query $query) use ($parentNode) {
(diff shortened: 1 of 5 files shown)


plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -420,15 +420,18 @@ protected function findAncestor(Query $query, array $options)
...
+        return $query->where(function (QueryExpression $exp) use ($parentNode): QueryExpression {
+            return $exp->in(
+                $this->aliasField('id'),
+                $this->TreeNodes->find()
+                    ->select(['object_id'])
+                    ->where(function (QueryExpression $exp) use ($parentNode) {
+                        return $exp
+                            ->gt($this->TreeNodes->aliasField('tree_left'), $parentNode->get('tree_left'))
+                            ->lt($this->TreeNodes->aliasField('tree_right'), $parentNode->get('tree_right'));
+                    }),
```
fquffio:
```suggestion
                    })
```

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectsTableTest.php
```diff
@@ -540,7 +540,7 @@ public function testFindAncestor()
         $objects = $this->Objects->find('ancestor', [11])->toArray();
         static::assertNotEmpty($objects);
         $ids = Hash::extract($objects, '{n}.id');
-        static::assertEquals([12, 4, 2], $ids);
+        static::assertEquals([12, 2, 4], $ids);
```
nicolocarpignoli: This test keeps failing: it seems that different runs give different results. Does it make any sense? How can we fix this? @fquffio

fquffio: It should be sufficient to enforce sorting by ID:

```php
        $objects = $this->Objects->find('ancestor', [11])
            ->sortAsc($this->Objects->aliasField('id'))
            ->toArray();
```

Then:

```suggestion
        static::assertEquals([2, 4, 12], $ids);
```

PR #1915, opened 2022-06-16
Update exported project model format

stefanorosanelli:
This PR introduces some changes to the `project_model` format exported via `/model/project` endpoint in order to be compatible with format used by `\BEdita\Core\Utility\Resources` engine, same used by resources migrations.

 Bonus track: `cakephp/plugin-installer` added to composer require


plugins/BEdita/Core/src/Utility/ProjectModel.php
```diff
@@ -122,11 +123,23 @@ protected static function relations(): array
                 $right = (array)Hash::extract($row, 'right_object_types.{n}.name');
                 sort($left);
                 sort($right);
-                unset($row['id']);
-                $row->set('left_object_types', $left);
-                $row->set('right_object_types', $right);
+                $row->unset(['id', 'left_object_types', 'right_object_types']);
+                $row->setAccess(['left', 'right'], true);
+                $row->set(compact('left', 'right'));
```
batopa: Maybe instead of changing property accessibility we could simply do
```php
$row->set('left', $left);
$row->set('right', $right);
```

plugins/BEdita/Core/src/Utility/ProjectModel.php
```diff
@@ -140,11 +153,19 @@ protected static function properties(): array
...
+                $hidden = [
+                    'id',
+                    'created',
+                    'modified',
+                    'label',
+                    'is_static',
+                    'object_type_name',
+                    'property_type_name',
+                ];
+                $row->setAccess(['object', 'property'], true);
```
batopa: We can remove  `$row->setAccess(['object', 'property'], true);` because we set a single property at the time.

batopa (changes requested):

plugins/BEdita/Core/src/Utility/ProjectModel.php
stefanorosanelli: ok, the problem was the `guard` option set to `true` on a mass assignment using `compact()` 

done

plugins/BEdita/Core/src/Utility/ProjectModel.php
stefanorosanelli: ok, done

batopa (approved):

PR #1922, opened 2022-07-04
Introducing StatusBehavior

edoardocavazza:
This PR introduces a new `StatusBehavior` in order to handle common methods and finders based on the `status` property.

It enables to apply the `statusLevel` query to `translations` finder.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Behavior/StatusBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/StatusBehavior.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * BEdita, API-first content management framework
- * Copyright 2018 ChannelWeb Srl, Chialab Srl
+ * Copyright 2022 ChannelWeb Srl, Chialab Srl
  *
  * This file is part of BEdita: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
@@ -35,8 +35,6 @@ class StatusBehavior extends Behavior
      */
     protected $_defaultConfig = [
         'implementedMethods' => [
-            'getStatusLevel' => 'getStatusLevel',
-            'setStatusLevel' => 'setStatusLevel',
             'checkStatus' => 'checkStatus',
         ],
         'implementedFinders' => [
@@ -44,27 +42,6 @@ class StatusBehavior extends Behavior
         ],
         'field' => 'status',
     ];
-
-    /**
-     * Get the configured status level.
-     *
-     * @return string
-     */
-    public function getStatusLevel()
-    {
-        return Configure::read('Status.level', 'all');
-    }
-
-    /**
-     * Set the status level configuration.
-     *
-     * @param string|null $level A valid status level.
-     * @return bool
-     */
-    public function setStatusLevel(?string $level)
-    {
-        return Configure::write('Status.level', $level);
-    }
 
     /**
      * Check that `status` is consistent with `level` configuration.
@@ -75,10 +52,10 @@ class StatusBehavior extends Behavior
      */
     public function checkStatus(EntityInterface $entity): void
     {
-        if ($entity->isNew() || !$entity->isDirty('status')) {
+        if ($entity->isNew() || !Configure::check('Status.level') || !$entity->isDirty('status')) {
             return;
         }
-        $level = $this->getStatusLevel();
+        $level = Configure::read('Status.level');
         $status = $entity->get('status');
         if (($level === 'on' && $status !== 'on') || ($level === 'draft' && $status === 'off')) {
             throw new BadRequestException(__d(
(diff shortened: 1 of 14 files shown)


plugins/BEdita/Core/src/Model/Behavior/StatusBehavior.php
```diff
@@ -0,0 +1,129 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2018 ChannelWeb Srl, Chialab Srl
```
stefanorosanelli: [superminor] it's 2022, unfortunately :wink:

plugins/BEdita/Core/tests/TestCase/Model/Behavior/StatusBehaviorTest.php
```diff
@@ -0,0 +1,221 @@
...
+        'plugin.BEdita/Core.ObjectTags',
+        'plugin.BEdita/Core.History',
+    ];
+
+    /**
+     * setUp method
+     *
+     * @return void
+     */
+    public function setUp()
```
stefanorosanelli: [minor] `: void` return type

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -88,6 +87,7 @@ public function initialize(array $config): void
 
         $this->addBehavior('BEdita/Core.ObjectModel');
         $this->addBehavior('BEdita/Core.Categories');
+        $this->addBehavior('BEdita/Core.Status');
```
stefanorosanelli: I don't think we need the behavior here if it's already in `ObjectModelBehavior`

plugins/BEdita/Core/src/Model/Behavior/StatusBehavior.php
```diff
@@ -0,0 +1,129 @@
...
+        ],
+        'field' => 'status',
+    ];
+
+    /**
+     * Get the configured status level.
+     *
+     * @return string
+     */
+    public function getStatusLevel()
```
stefanorosanelli: I'm not sure if it's a good idea to have `getStatusLevel`/`setStatusLevel` methods. 
I'd prefer to keep using `Configure::read/write` explicitly throughout the code. But i's a matter of taste I guess, it is clearer for me.

Since we are changing a global configuration a call like `$Documents->setStatusLevel()` don''t simply affect the single documents table class but every object table...

fquffio: I agree.

While I don't like the `Configure::read()` approach (in my opinion it should only be used at startup, then read it through some other mean afterwards), interacting with a global setting through a behavior, which will be instantiated multiple times, is set to be more harm than good, and is likely to cause a lot of confusion.

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -500,7 +439,12 @@ protected function findStatusLevel(Query $query, array $options)
     protected function findTranslations(Query $query, array $options)
     {
         return $query->contain('Translations', function (Query $query) use ($options) {
-            return $query->where(['Translations.lang' => $options['lang']]);
+            $query = $query->find('statusLevel', [Configure::read('Status.level')]);
```
stefanorosanelli: use default to avoid missing config errors
```php
[Configure::read('Status.level', 'all')]
```

plugins/BEdita/Core/src/Model/Table/ObjectsTable.php
```diff
@@ -520,17 +464,15 @@ protected function findAvailable(Query $query): Query
...
      *
      * @param \Cake\ORM\Query $query Query object instance.
      * @return \Cake\ORM\Query
      */
     protected function findPublishable(Query $query): Query
     {
-        if (Configure::check('Status.level')) {
-            $query = $query->find('statusLevel', [Configure::read('Status.level')]);
-        }
+        $query = $query->find('statusLevel', [Configure::read('Status.level')]);
```
stefanorosanelli: [same as above] use default to avoid missing config errors
```php
[Configure::read('Status.level', 'all')]
```

plugins/BEdita/Core/tests/TestCase/Model/Table/ObjectsTableTest.php
```diff
@@ -887,7 +780,8 @@ public function findPublishableProvider(): array
      * Test `findPublishable()`.
      *
      * @param int $expected Expected results.
-     * @param string $config Configuration to write.
+     * @param string $$statusLevel Status level.
```
stefanorosanelli: missing param

stefanorosanelli (approved):

PR #1925, opened 2022-07-07
`/config` endpoint refactor

stefanorosanelli:
This PR changes the behavior of `/config` endpoint. 
It is now used to store application related configurations: only `config` resources with `application_id` matching current application and fixed `'app'` `context` can now be created/updated and removed by the client application.

In `/admin/config` all config resources are still available.


plugins/BEdita/Core/src/Model/Table/AppConfigTable.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function newEmptyEntity(): EntityInterface
+    {
+        $entity = parent::newEmptyEntity();
+        $entity->set('context', static::DEFAULT_CONTEXT, ['guard' => false]);
+        $entity->set('application_id', CurrentApplication::getApplicationId(), ['guard' => false]);
```
batopa: Setting a property doesn't require `guard` option. It would need if you set both properties at the same time.

```suggestion
        $entity->set('context', static::DEFAULT_CONTEXT);
        $entity->set('application_id', CurrentApplication::getApplicationId());
```

plugins/BEdita/Core/tests/TestCase/Model/Table/AppConfigTableTest.php
```diff
@@ -0,0 +1,155 @@
...
+        parent::tearDown();
+    }
+
+    /**
+     * Test initialization.
+     *
+     * @return void
+     * @coversNothing
+     */
+    public function testInitialization()
```
batopa:
```suggestion
    public function testInitialization(): void
```

plugins/BEdita/Core/tests/TestCase/Model/Table/AppConfigTableTest.php
```diff
@@ -0,0 +1,155 @@
...
+        $this->assertEquals('config', $this->AppConfig->getTable());
+        $this->assertEquals('id', $this->AppConfig->getPrimaryKey());
+    }
+
+    /**
+     * Data provider for `testValidation` test case.
+     *
+     * @return array
+     */
+    public function validationProvider()
```
batopa:
```suggestion
    public function validationProvider(): array
```

plugins/BEdita/Core/tests/TestCase/Model/Table/AppConfigTableTest.php
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Test validation.
+     *
+     * @param bool $expected Expected result.
+     * @param array $data Data to be validated.
+     * @return void
+     * @dataProvider validationProvider
+     * @covers ::validationDefault()
+     */
+    public function testValidation($expected, array $data)
```
batopa:
```suggestion
    public function testValidation($expected, array $data): void
```

plugins/BEdita/Core/tests/TestCase/Model/Table/AppConfigTableTest.php
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    /**
+     * Test `all` finder
+     *
+     * @return void
+     * @covers ::findAll()
+     */
+    public function testFindAll()
```
batopa:
```suggestion
    public function testFindAll(): void
```

plugins/BEdita/Core/tests/TestCase/Model/Table/AppConfigTableTest.php
```diff
@@ -0,0 +1,155 @@
...
+        CurrentApplication::setApplication($app);
+    }
+
+    /**
+     * Test `newEmptyEntity` method
+     *
+     * @return void
+     * @covers ::newEmptyEntity()
+     */
+    public function testNewEmptyEntity()
```
batopa:
```suggestion
    public function testNewEmptyEntity(): void
```

batopa (changes requested):

batopa (approved):

PR #1940, opened 2022-09-29
refactor: use BEdita TableLocator instead of default one when creatin…

le0m:
This PR changes the utility class `Resources` to use BEdita's `TableLocator` instead of the default one, which requires removing and readding the requested table and does not handle casting arrays to strings.

### DIFF
--- a/plugins/BEdita/Core/src/Utility/ResourcesBase.php
+++ b/plugins/BEdita/Core/src/Utility/ResourcesBase.php
@@ -13,9 +13,9 @@
 
 namespace BEdita\Core\Utility;
 
+use BEdita\Core\ORM\Locator\TableLocator;
 use Cake\Http\Exception\BadRequestException;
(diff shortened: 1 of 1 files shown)


plugins/BEdita/Core/src/Utility/ResourcesBase.php
```diff
@@ -47,8 +47,8 @@ protected static function getTable(string $type, array $options = []): Table
             );
         }
         $type = Inflector::camelize($type);
-        TableRegistry::getTableLocator()->remove($type);
+        $tl = new TableLocator();
```
batopa, line 50: Maybe it should be better to set our `TableLocator` in `TableRegistry` as we do in the core bootstrap.

```php
if (!TableRegistry::getTableLocator() instanceof TableLocator) {
    TableRegistry::setTableLocator(new TableLocator());
}
```

PR #1942, opened 2022-10-03
Fix signup error for `draft` users

stefanorosanelli:
This PR solves a problem in `/signup` logic when a status level filter is active, namely with `'Status.level' configuration is `on`.

Upon successful signup, at the end of `SignupUserAction` the newly created user is returned via `GetObjectAction`. If a user requires a validation (email verification usually) its user object can be `draft` and a 404 Not Found error will be raised if `'Status.level' conf is `on`. Solution: get rid of status level filter in this particular case.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
@@ -154,7 +154,7 @@ class SignupUserAction extends BaseAction implements EventListenerInterface
 
         $user = (new GetObjectAction(['table' => $this->Users]))->execute(['primaryKey' => $user->id, 'contain' => 'Roles']);
         Configure::write('Status.level', $statusLevel);
-        
+
         return $user;
     }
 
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -147,6 +148,8 @@ public function execute(array $data = [])
 
             throw $e;
         }
+        // Remove `Status.level` filter here to allow user object with `draft` or any other status
+        Configure::delete('Status.level');
 
         return (new GetObjectAction(['table' => $this->Users]))->execute(['primaryKey' => $user->id, 'contain' => 'Roles']);
```
batopa: Maybe it would be better to restore the status level after getting user to consider it in any later queries 
```suggestion
        $statusLevel = Configure::read('Status.level');
        Configure::delete('Status.level');

        $user = (new GetObjectAction(['table' => $this->Users]))->execute(['primaryKey' => $user->id, 'contain' => 'Roles']);
        Configure::write('Status.level', $statusLevel);
        
        return $user;
```

batopa (changes requested):

batopa (approved):

PR #1943, opened 2022-10-13
Regex to validate internal resource names

stefanorosanelli:
This PR introduces a new regex to validate some internal resources names.

Regex is ```/^[a-z][a-z0-9_]{1,200}$/```
 * starts with a lowercase letter
 * composed by lowercase letters, numbers and underscores
 * length between 2 and 200 characters

This new validation regexp applies to: 
 * `auth_providers.name`
 * `endpoints.name`
 * `object_types.name`
 * `object_types.singular`
 * `properties.name`
 * `property_types.name`
 * `relations.name`
 * `relations.inverse_name`
 * `roles.name`

### DIFF
--- a/plugins/BEdita/Core/src/Model/Validation/Validation.php
+++ b/plugins/BEdita/Core/src/Model/Validation/Validation.php
@@ -30,7 +30,7 @@ class Validation
 {
     /**
      * Regular expression to validate resource names:
-     *  - starts with a lowercase latter
+     *  - starts with a lowercase letter
      *  - lowercase letters, numbers and underscores
      *  - length between 2 and 200 characters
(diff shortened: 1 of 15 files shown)


plugins/BEdita/Core/src/Model/Validation/Validation.php
```diff
@@ -28,6 +28,16 @@
  */
 class Validation
 {
+    /**
+     * Regular expression to validate resource names:
+     *  - starts with a lowercase latter
```
batopa:
```suggestion
     *  - starts with a lowercase letter
```

batopa (approved):

PR #1944, opened 2022-10-17
Add new mime_type : model/gltf-binary for application

andresbautista:
Add new `mime_type`: `model/gltf-binary` for object_type application

### DIFF
--- a/bedita-app/models/objects/content/stream.php
+++ b/bedita-app/models/objects/content/stream.php
@@ -284,7 +284,7 @@ class Stream extends BEAppModel
 
         $mime_type = $this->getMimeTypeByExtension($filename);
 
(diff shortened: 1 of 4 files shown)


bedita-app/models/objects/content/stream.php
```diff
@@ -284,7 +284,8 @@ public function getMimeType($path, $filename=null) {
 
         $mime_type = $this->getMimeTypeByExtension($filename);
 
-        if (function_exists("finfo_open") && file_exists($path)) {
+        // Not change mime_type of glb file (binary gltf) with magic_mime_type because it is wrong
```
le0m:
```suggestion
        // Don't use finfo_file for GLB files because it returns a wrong MIME type
```

le0m (approved):

PR #1946, opened 2022-10-25
History include filter and related endpoints

edoardocavazza:
This PR introduces related endpoints for history entities: `/history/{id}/user`. It also fixes `include=user` filter for index method.

### DIFF
--- a/plugins/BEdita/API/src/Controller/HistoryController.php
+++ b/plugins/BEdita/API/src/Controller/HistoryController.php
@@ -12,53 +12,52 @@
  */
 namespace BEdita\API\Controller;
 
-use BEdita\Core\Model\Action\ListEntitiesAction;
-use BEdita\Core\Utility\JsonApiSerializable;
-use Cake\Core\Configure;
-use Cake\ORM\TableRegistry;
+use Cake\Http\Exception\NotFoundException;
+use Cake\ORM\Association;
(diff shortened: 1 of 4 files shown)


stefanorosanelli (approved):

plugins/BEdita/API/src/Controller/HistoryController.php
```diff
@@ -12,53 +12,35 @@
...
-        parent::initialize();
-
-        $historyTable = (string)Configure::read('History.table', 'History');
-        $this->HistoryTable = TableRegistry::getTableLocator()->get($historyTable);
-    }
+    protected $_defaultConfig = [
+        'allowedAssociations' => [
+            'user' => ['users'],
+        ],
+    ];
```
stefanorosanelli:
```suggestion
     * @inheritDoc
     */
    protected $_defaultConfig = [
        'allowedAssociations' => [
            'user' => ['users'],
        ],
    ];

    /**
     * @inheritDoc
     */
    public function initialize(): void
    {
        $this->modelClass = (string)Configure::read('History.table', 'History');

        parent::initialize();
    }
```

stefanorosanelli: This way we keep the possibility to use a different table via `History.table` config.

PR #1953, opened 2022-11-03
Fix `json` prop type as pure JSON + additional JSON types

stefanorosanelli:
This PR fixes a problem on core property types: the `json` prop type should represent any valid JSON for which the JSON SCHEMA is `{}` and not current `{"type": "object"}` that represents a JSON object. 

Some useful JSON core properties have also been added: 

 * `json_object` => with schema  `{"type": "object"}`
 * `string_array` => a simple array of strings 
 * `string_map` => a simple key/value map with string values


plugins/BEdita/Core/src/Model/Entity/Property.php
```diff
@@ -210,6 +210,9 @@ public function getSchema($accessMode = null)
             // Booleans are valid schemas, though they're quite uncommon.
             return $schema;
         }
+        if (empty($schema)) {
+            $schema = new \stdClass();
```
batopa: What happen if `$schema = new \stdClass();` and is not nullable?
At row `230` there is an array assignement `$schema['$id'] = sprintf('/properties/%s', $this->name);` that would give an error as `Cannot use object of type stdClass as array.`

stefanorosanelli: Right, in case of empty array as `$schema` we should probably have just 

```null
  $schema = [
                'oneOf' => [
                    [
                        'type' => 'null',
                    ],
                    new \stdClass(),
                ],
   ];
```

When reading a property schema if there's no `type` directive we should treat the property as JSON

batopa (approved):

Issue #1956, opened 2022-11-30
Endpoint to get and save trees.children_order


didoda:
After https://github.com/bedita/bedita/issues/1954 and https://github.com/bedita/bedita/pull/1955, children order customization is saved in `trees.children_order`.

We need to provide an endpoint to get and to update `trees.children_order` for a specified folder.

For instance:

```
GET /folders/{{id}} // children_order in response meta

PATCH /folders/{{id}}
{
    "data": {
        "id": "{{id}}",
        "type": "folders",
        "meta": {
            "children_order": "-title"
        }
    }
}
```

or

```
GET /folders/{{id}}/children_order // children_order in response { "children_order": "-title" }

PATCH /folders/{{id}}/children_order
{
    "children_order": "-title"
}
```

didoda:
Done with #1957 and #1959

PR #1957, opened 2022-12-01
Folders children_order property and ListAssociatedAction custom sort for Children

### LINKED ISSUE #1954: Children sorting management per single object
### Sort folder children by a specified objects field (title, modified)

Every folder should have a default sorting logic available for related `GET /folders/:id/children`.
When a default sort field is set, all `GET /folders/:folderId/children` return data sort by that field.

I.e.: we want to sort by `-modified` field the children of 19123 folder. The `trees` record for id 19123 could have a field `children_order` with value `-modified`. In this example `-modified` means "ORDER BY objects.modified DESC".  `GET /folders/19123/children` will return data order by `-modified`.

didoda:
This PR fixes #1954 and https://github.com/bedita/bedita/issues/1956 (that becomes obsolete) using a new folders property "children_order".

### DIFF
--- a/plugins/BEdita/Core/config/Migrations/20221201092959_AddChildrenOrder.php
+++ b/plugins/BEdita/Core/config/Migrations/20221201092959_AddChildrenOrder.php
@@ -54,17 +54,15 @@ class AddChildrenOrder extends AbstractMigration
      */
     public function up()
     {
+        $connection = $this->getAdapter()->getCakeConnection();
         Resources::save(
             ['create' => $this->create],
-            ['connection' => $this->getAdapter()->getCakeConnection()]
+            ['connection' => $connection]
         );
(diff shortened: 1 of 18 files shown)


plugins/BEdita/Core/config/Migrations/20221201092959_AddChildrenOrder.php
```diff
@@ -0,0 +1,74 @@
...
+
+use BEdita\Core\Utility\Resources;
+use Migrations\AbstractMigration;
+
+class AddChildrenOrder extends AbstractMigration
+{
+    protected $create = [
+        'property_types' => [
+            [
+                'name' => 'objects_order',
```
stefanorosanelli: I'd use `children_order` as property type name - this way we could handle the property automagically in the Manager app

stefanorosanelli (approved):

batopa (approved):

PR #1959, opened 2022-12-06
Folders children_order property and ListAssociatedAction custom sort for Children

### LINKED ISSUE #1954: Children sorting management per single object

didoda:
This PR fixes https://github.com/bedita/bedita/issues/1954 and https://github.com/bedita/bedita/issues/1956 (that becomes obsolete) using a new folders property "children_order".

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/ListAssociatedAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/ListAssociatedAction.php
@@ -221,7 +221,8 @@ class ListAssociatedAction extends BaseAction
             $query = $query->select($this->Association->junction());
         }
         if ($this->Association instanceof BelongsToMany || $this->Association instanceof HasMany) {
-            $query = $query->order($this->Association->getSort());
+            $sort = $this->sort($this->Association, $primaryKey);
+            $query = $query->order($sort);
         }
 
         $primaryKeyConditions = $this->primaryKeyConditions($inverseAssociation->getTarget(), $primaryKey);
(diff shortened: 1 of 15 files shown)


plugins/BEdita/Core/src/Model/Action/ListAssociatedAction.php
```diff
@@ -286,4 +286,21 @@ public function execute(array $data = [])
...
+
+    /**
+     * Get association sort by association and primary key.
+     * When association name is "Children", use Folders.getSort($primaryKey).
+     *
+     * @param mixed $primaryKey Primary key
+     * @param \Cake\ORM\Association $association Association
+     * @return array
+     */
+    protected function sort(Association $association, $primaryKey): array
```
stefanorosanelli, line 299: This method seems to be not covered
@didoda can you confirm?

didoda: `ListAssociatedActionTest::testSort` should cover it

didoda: added explicit `@covers` in `ListAssociatedActionTest`

stefanorosanelli (approved):

PR #1963, opened 2022-12-21
Add Enqueue support to async jobs

stefanorosanelli:
This PR introduces [Enqueue](https://php-enqueue.github.io/) support to BEdita async jobs via [Cake/Queue plugin](https://github.com/cakephp/queue).

Main changes:
* a new  `QueueJob` class has been added to run every async job using the `ServiceRegistry` 
* as soon as a new `async` job is created a new `QueueJob` is pushed onto the queue for immediate processing

Currently only `Processor::ACK` and `Processor::REJECT` results are handled.
For now `Processor::REQUEUE` cases can be handled via usual unix cron jobs to retry previously failed jobs having attempts still available.


plugins/BEdita/Core/tests/TestCase/Job/QueueJobTest.php
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        return $service;
+    }
+
+    /**
+     * Create queue message
+     *
+     * @param array $body Message body
+     * @return Message
```
batopa: FQN

```suggestion
     * @return \Cake\Queue\Job\Message
```

composer.json
```diff
@@ -25,14 +25,16 @@
         "cakephp/cakephp": "~4.4.1",
         "cakephp/plugin-installer": "^1.3.1",
         "guzzlehttp/psr7": "^2.2.1",
-        "wikimedia/composer-merge-plugin": "^2.0.1"
+        "wikimedia/composer-merge-plugin": "^2.0.1",
+        "cakephp/queue": "^0.1.11"
```
batopa: I would move in `BEdita/Core` `composer.json`

batopa (changes requested):

batopa (approved):

PR #1973, opened 2023-02-07
Avoid injection of URLs or tags in `name`, `surname` or `username`

stefanorosanelli:
This PR contains a security fix on users/profiles data to avoid insertion of possible malicious elements like tags, scripts or URLs in fields like `username`, `name` or `surname` of Profiles and Users. 

* in `ProfilesValidator` a new rule has been added to avoid insertion of tags or urls in `name` and `username` 
* in `UsersValidator` a new rule has been added to allow the use of email format as `surname`, but to avoid insertion of possible tags/urls 
* (bonus) allow `_` usage in category names

### DIFF
--- a/plugins/BEdita/Core/src/Model/Validation/UsersValidator.php
+++ b/plugins/BEdita/Core/src/Model/Validation/UsersValidator.php
@@ -14,7 +14,7 @@
 namespace BEdita\Core\Model\Validation;
 
 use Cake\ORM\TableRegistry;
-use Cake\Validation\Validation;
+use Cake\Validation\Validation as CakeValidation;
 
 /**
  * Validator for users.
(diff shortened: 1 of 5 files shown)


plugins/BEdita/Core/src/Model/Validation/UsersValidator.php
```diff
@@ -45,4 +47,37 @@ public function __construct()
...
+        if (!preg_match(ProfilesValidator::NAME_REGEX, $value, $matches)) {
+            return false;
+        }
+
+        // check for domain name presence
+        if (!preg_match(ProfilesValidator::NO_DOMAIN_REGEX, $value, $matches)) {
+            return false;
+        }
+
+        return true;
```
batopa:
```suggestion
        return parent::validName($value);
```

batopa (changes requested):

batopa (approved):

PR #1975, opened 2023-02-10
Handle `translatable` objects and properties

stefanorosanelli:
This PR introduces some logic to handle `translatable` object and properties

* a new column and property `object_types.is_translatable` (default `true`) to determine if an object is translatable or not
* a new `translatable` array on object JSON SCHEMA to list `translatable` properties 
* a `translatable` virtual getter on dynamic and static properties - i.e. `Property` and `StaticProperty` entities - to list actually translatable properties - properties that contain some text like text/string columns/properties instead of numbers, dates etc...    
* a new column and property `object_types.translation_rules` (default `null`) to define exceptions on the current rules implied by the `translatable` getter above - this is done via a JSON object with this simple structure 

```json
{
    "prop1": true,
    "prop2": false,
}
```
Every key is a property name and every value forces the desired outcome: `true` if we want to make a property translatable (even if theoretically it isn't), `false` if we want to make a property not translatable (even if it is a textual property). In the example above `prop1` will be `translatable` and `prop2` not, regardless of their features.

* a new `notTranslatable` internal array in `ObjectEntity` to list textual properties that are not translatable like `uname` and `lang` with two methods:
  - `addNotTranslatable` to define additional not translatable properties 
  - `isFieldTranslatable` to see if a particular property is listed in `notTranslatable` array


didoda (approved):

batopa:
I think we would add as not translatable other entity fields. 
`Profile` , `Location`, `Publication` have "not translatable" properties,  for example `Profile.name`, `Publication.public_url`, `Location.coords`, etc...

stefanorosanelli:
> I think we would add as not translatable other entity fields. `Profile` , `Location`, `Publication` have "not translatable" properties, for example `Profile.name`, `Publication.public_url`, `Location.coords`, etc...

@batopa you're right: can you provide a more complete list?

batopa:
I'd say, for `Profile` entity:

```
name
surname
email
company_name
street_address
zipcode
phone
website
national_id_number
vat_number
pseudonym
 ```

For `Location` entity:

```
coords
address (not sure about it)
postal_code
```
 For `Publication` entity

```
public_url
staging_url
stats_code
```

stefanorosanelli:
> I'd say, for `Profile` entity:

@batopa pushed a commit to use `addNotTranslatable` on listed fields

batopa (approved):

PR #1976, opened 2023-02-13
Category and tag name validation: allow name starts with a lowercase letter or a number

didoda:
This PR updates `CATEGORY_NAME_REGEX` used in `tags` and `categories` save, to allow "starts with a lowercase letter or a number".

### DIFF
--- a/plugins/BEdita/Core/src/Model/Validation/Validation.php
+++ b/plugins/BEdita/Core/src/Model/Validation/Validation.php
@@ -30,14 +30,14 @@ class Validation
 {
     /**
      * Regular expression to validate category names:
-     *  - starts with a lowercase letter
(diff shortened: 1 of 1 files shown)


plugins/BEdita/Core/src/Model/Validation/Validation.php
```diff
@@ -30,14 +30,14 @@ class Validation
 {
     /**
      * Regular expression to validate category names:
-     *  - starts with a lowercase letter
+     *  - starts with a lowercase letter or a number
      *  - lowercase letters, numbers and hyphens
```
stefanorosanelli: - lowercase letters, numbers, hyphens and underscores as accepted characters

stefanorosanelli (approved):

PR #1979, opened 2023-02-21
Handle `num_login_err` and `last_login_err` on authentication failure

stefanorosanelli:
In this PR `users.num_login_err` (consecutive number of login failures) and `users.last_login_err` (last login failure timestamp) fields are updated upon authentication failure with `grant_type` password.

A new event `Authentication.failure` has been introduced for this use case.

`'Authentication.afterIdentify' => 'login'` item has been removed from `implementedEvents()` because the method was called twice (there is also an explicit `EventManager::instance()->on(..)` call)

### DIFF
--- a/plugins/BEdita/Core/src/Model/Table/UsersTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/UsersTable.php
@@ -236,9 +236,8 @@ class UsersTable extends Table
         /** @var \Cake\Http\ServerRequest|null $request */
         $request = $event->getData('request');
         if (
-            empty($request) ||
-            (string)$request->getData('grant_type') !== 'password' ||
-            !$request instanceof ServerRequest
+            !$request instanceof ServerRequest ||
+            (string)$request->getData('grant_type') !== 'password'
         ) {
             return;
         }
--- a/plugins/BEdita/Core/src/Model/Table/UsersTable.php
+++ b/plugins/BEdita/Core/src/Model/Table/UsersTable.php
@@ -111,7 +111,7 @@ class UsersTable extends Table
         ]);
 
         EventManager::instance()->on('Authentication.afterIdentify', [$this, 'login']);
-        EventManager::instance()->on('Authentication.failure', [$this, 'failure']);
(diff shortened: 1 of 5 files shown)


plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -216,13 +216,39 @@ public function login(EventInterface $event)
...
+     * Update fields on authentication failure.
+     *
+     * @param \Cake\Event\EventInterface $event Dispatched event.
+     * @return void
+     */
+    public function authFailure(EventInterface $event): void
+    {
+        /** @var \Cake\Http\ServerRequest|null $request */
+        $request = $event->getData('request');
+        if (empty($request) || (string)$request->getData('grant_type') !== 'password') {
```
batopa: Maybe it's better to check if `$request` is instance of `ServerRequest`

batopa (changes requested):

plugins/BEdita/Core/src/Model/Table/UsersTable.php
stefanorosanelli: done

plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -216,13 +217,43 @@ public function login(EventInterface $event)
...
+     */
+    public function authFailure(EventInterface $event): void
+    {
+        /** @var \Cake\Http\ServerRequest|null $request */
+        $request = $event->getData('request');
+        if (
+            empty($request) ||
+            (string)$request->getData('grant_type') !== 'password' ||
+            !$request instanceof ServerRequest
+        ) {
```
batopa: It would be enough

```suggestion
        if (
            !$request instanceof ServerRequest ||
            (string)$request->getData('grant_type') !== 'password'
        ) {
```
I suggest to place first the instance check to avoid a possible warning calling `getData()`.

batopa (approved):

PR #1988, opened 2023-03-21
Add role priority

didoda:
This PR provides `priority` field for Roles to handle a privilege escalation issue.

Priority for `admin` is 0 and its the strongest possible. Default priority is 100. Lower value means more power for the role.

User roles can be modified only by a user that has "enough power": its roles min `priority` MUST BE less than or equal to the role priority involved in the user change. 

E.g: a user with role priority `10` can add or remove role association to another user only if the involved roles have a priority value of `10` or more. Otherwise the role association API call will fail with a `403 Forbidden` response.


stefanorosanelli (approved):

plugins/BEdita/Core/src/Model/Table/RolesUsersTable.php
```diff
@@ -87,17 +90,62 @@ public function buildRules(RulesChecker $rules): RulesChecker
...
+     * Logged user roles min priority should be less or equal to the role priority.
+     *
+     * @param int $roleId The role ID to check againt logged user roles priorities
+     * @return bool
+     */
+    protected function canHandle(int $roleId): bool
+    {
+        $user = LoggedUser::getUser();
+        $roles = (array)Hash::get($user, 'roles');
+        $ids = (array)Hash::extract($roles, '{n}.id');
```
batopa:
```suggestion
        $ids = (array)Hash::extract($user, 'roles.{n}.id');
```

batopa (approved):

PR #1995, opened 2023-03-23
Fix signup: create user with role admin

didoda:
This PR fixes a problem generated by recent introduction of `roles.priority`.

During signup process, `admin` user creates the new user. Admin role is necessary to provide a `roles.priority` change.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
@@ -271,7 +271,8 @@ class SignupUserAction extends BaseAction implements EventListenerInterface
     protected function createUser(array $data)
     {
         if (!LoggedUser::getUser()) {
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Model/Action/SignupUserAction.php
```diff
@@ -271,7 +271,8 @@ public function validateRoles($roles)
     protected function createUser(array $data)
     {
         if (!LoggedUser::getUser()) {
-            LoggedUser::setUser(['id' => 1]);
+            // use user 1 (admin) role 1 (admin / unchangeable)
+            LoggedUser::setUser(['id' => 1, 'roles' => [['id' => 1]]]);
```
batopa: Introdurrei le costanti `UsersTable::ADMIN_USER` e `RolesTable::ADMIN_ROLE`

batopa (approved):

PR #1999, opened 2023-03-27
refactor: change objects description and body to be mediumtext

le0m:
This PR increases `objects.description` and `objects.body` size to medium-text.


plugins/BEdita/Core/src/Model/Validation/ObjectsValidator.php
```diff
@@ -57,8 +57,10 @@ public function __construct()
             ->allowEmptyString('title')
 
             ->allowEmptyString('description')
+            ->maxLengthBytes('description', (2 << 23) - 1) // 2^24 - 1 bytes = 16MiB
```
fquffio: Rather than hardcoding these values, could we read length from cached schema?

```suggestion
            ->maxLengthBytes('description', $this->getSchema()->getColumn('description')['length'])
```

plugins/BEdita/Core/config/Migrations/20230327104919_ChangeObjectsDescriptionBodySize.php
```diff
@@ -0,0 +1,55 @@
...
+     * Up Method.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        $limit = null;
+        if ($this->getAdapter()->getAdapterType() === 'mysql') {
+            $limit = MysqlAdapter::TEXT_MEDIUM;
+        }
```
fquffio: I know this "well-known" field length is specific to MySQL, but I can't see why it shouldn't work for other engines too. Couldn't we just use the same length for all?

fquffio (approved):

plugins/BEdita/Core/src/Model/Validation/ObjectsValidator.php
le0m: The schema's length for sqlite and Postgress is empty, so I used `MysqlAdapter::TEXT_MEDIUM` which is a numeric costant

stefanorosanelli (approved):

batopa (approved):

PR #2000, opened 2023-03-29
Introduce object permissions in meta

didoda:
This updates `object_permissions` table and provides perms data in `meta`, i.e.:

```json
"meta": {
    "perms": {
        "roles": ["manager", "frontend"],
        "inherited": false
    }
}
```

Permissions logic available only if `folders` model has association `Permissions`.


plugins/BEdita/Core/src/Model/Entity/ObjectEntity.php
```diff
@@ -351,4 +361,48 @@ public function isFieldTranslatable(string $name): bool
...
+
+        if (!$this->hasProperty('permissions')) {
+            $this->getTable()->loadInto($this, ['Permissions']);
+        }
+
+        if (empty($roleIds)) {
+            return [];
+        }
+
+        $roles = TableRegistry::getTableLocator()->get('Roles')
```
stefanorosanelli: We should probably introduce some sort of cache for `Roles` - resources rarely updated and easy to invalidate cache. 
See `ObjectTypesTable` for instance, `get`, `afterSave` and `afterDelete` methods

batopa: Maybe I would introduce cache in another specific PR

fquffio: Rather than introducing a new cache, I'd change `->loadInto($this, ['Permissions.Roles'])`, which uses a straightforward INNER JOIN.

The amount of queries executed would remain the same even when cache will be introduced, and right now would decrease by 1.
And I bet that as long as the number of Roles remains reasonable, that added join would add close to nothing to the complexity of the query executed by `loadInto`.

Once that has been done, it would be easy to extract the name of the relevant roles:

```php
$roles = array_unique(Hash::extract((array)$this->permissions, '{n}.role.name'));
```

batopa: I had though about it but if `Permissions` was already loaded we are not sure that `Roles` are loaded too, so for every `Permission` entity we have to check that roles are loaded.
We could go with something similar to:
```php
$roles = array_unique(Hash::extract((array)$this->permissions, '{n}.role.name'));

// ensure permission and roles are loaded
if (empty($roles)) {
    $this->getTable()->loadInto($this, ['Permissions.Roles']);
    $roles = array_unique(Hash::extract((array)$this->permissions, '{n}.role.name'));
}
```
that reload the permissions entities if `roles` were missing from original permissions, but maybe we could consider it as an extreme case

fquffio: I agree that we have no guarantee about Roles being loaded, but I'd treat it as a gentlemen's agreement and assume that if Permissions are loaded, their Roles are present too… at least as a first implementation, surely the snippet you wrote would be the safest way to go.

batopa: Ok changed.... even if we are gentlemen it is better to be safe :grin:

PR #2001, opened 2023-04-13
Fix docker entrypoint perms for sqlite db

didoda:
This PR fixes a permissions problem on sqlite database by using `chown www-data:www-data` on sqlite database file.

The folder that houses the database file must be writeable (https://www.php.net/manual/en/ref.pdo-sqlite.php#57356), so this applies proper perms on database directory too.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -24,7 +24,8 @@ if [ ! -z "${DATABASE_URL}" ]; then
         DATABASE_PATH=$(php -r "echo substr(parse_url(preg_replace('/^([\\w\\\\\\]+)/', 'file', getenv('DATABASE_URL')), PHP_URL_PATH), 1);")
         echo "=====> Path: ${DATABASE_PATH}"
         chmod a+rwx ${DATABASE_PATH}
-        chown www-data:www-data ${DATABASE_PATH}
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -24,6 +24,8 @@ if [ ! -z "${DATABASE_URL}" ]; then
         DATABASE_PATH=$(php -r "echo substr(parse_url(preg_replace('/^([\\w\\\\\\]+)/', 'file', getenv('DATABASE_URL')), PHP_URL_PATH), 1);")
         echo "=====> Path: ${DATABASE_PATH}"
         chmod a+rwx ${DATABASE_PATH}
+        chown www-data:www-data ${DATABASE_PATH} ${DATABASE_DIR}
+        chmod g+w ${DATABASE_DIR}
```
fquffio, line 29: Might also work without `DATABASE_DIR` env variable being set manually in `Dockerfile` inferring its value using `dirname`:

```suggestion
        DATABASE_DIR="$(dirname ${DATABASE_PATH})"
        chown www-data:www-data ${DATABASE_PATH} ${DATABASE_DIR}
        chmod g+w ${DATABASE_DIR}
```

PR #2004, opened 2023-04-18
Descendant folder permissions in `meta.perms.descendant_perms_granted`

le0m:
This PR adds information in `meta` about if the current user has permissions for a descendant folder.

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
@@ -2863,7 +2863,7 @@ class ObjectsControllerTest extends IntegrationTestCase
         $expected = [
             'roles' => ['second role'],
             'inherited' => false,
-            'descendant_have_perms' => false,
+            'descendant_perms_granted' => false,
         ];
 
         static::assertEquals($expected, Hash::get($result, 'data.0.meta.perms'));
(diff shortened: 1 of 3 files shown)


plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
```diff
@@ -2863,6 +2863,7 @@ public function testParentsPermissions(): void
         $expected = [
             'roles' => ['second role'],
             'inherited' => false,
+            'descendant_have_perms' => false,
```
didoda: Not sure that `descendant_have_perms` can have a straightforward meaning... It could be confused with the meaning "the descendant folders have permissions set". We want a label that tells "the authenticated user roles grants full access to descendant folders" 
Maybe we could use another label, as `descendant_perms_granted`.

plugins/BEdita/Core/src/Model/Entity/Folder.php
```diff
@@ -101,6 +105,44 @@ protected function getInheritedRolesPermissions(): array
...
+     * @return bool
+     */
+    protected function descendantHavePermissions(): bool
+    {
+        $user = LoggedUser::getUser();
+        if (empty($user)) {
+            return false;
+        }
+
+        $Trees = TableRegistry::getTableLocator()->get('Trees');
```
didoda, line 124: if user has role 'admin' you don't need the query, just return true.

didoda (changes requested):

plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
le0m: Completely agree, naming things is hard

didoda (approved):

PR #2005, opened 2023-04-20
Introduce object_permissions endpoint

batopa:
With this PR a new endpoint `object_permissions` is added.
The endpoint supports creation and deletion (not edit) of object_permissions.


plugins/BEdita/API/tests/TestCase/Controller/ObjectPermissionsControllerTest.php
```diff
@@ -0,0 +1,184 @@
...
+        'plugin.BEdita/Core.ObjectPermissions',
+    ];
+
+    /**
+     * Test index method.
+     *
+     * @return void
+     * @coversNothing
+     */
+    public function testIndex()
```
stefanorosanelli: [superminor] add `: void` return type, in other methods too

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
```diff
@@ -377,6 +379,39 @@ protected function _getSchema()
         return $event->getResult() ?: $schema;
     }
 
+    /**
+     * Virtual property to get info about permissions availability.
+     *
+     * @return bool
+     */
+    protected function _getPermissionsEnabled(): bool
```
stefanorosanelli: Instead of a virtual prop can we use methods like 
`hasAssoc(string $name)` and `addAssoc(string $name)` `removeAssoc(string $name)` ? 

This way we don't expose a new property in API calls and we don't need to hide it if we don't want it

The permissions enabled/disabled information is already in the `associations` property

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
```diff
@@ -377,6 +379,39 @@ protected function _getSchema()
...
+     */
+    protected function _setPermissionsEnabled(bool $enable): bool
+    {
+        if ($enable) {
+            $this->associations = array_unique(array_merge((array)$this->associations, ['Permissions']));
+
+            return $enable;
+        }
+
+        if (empty($this->associations)) {
```
stefanorosanelli: some lines are not covered by tests in `_setPermissionsEnabled`

fquffio (approved):

plugins/BEdita/Core/src/Model/Table/ObjectPermissionsTable.php
```diff
@@ -86,4 +106,69 @@ public function buildRules(RulesChecker $rules): RulesChecker
...
+        }
+    }
+
+    /**
+     * Check if logged user can create, edit or delete the object permission.
+     *
+     * Object permission is editable if one of below condition is true:
+     * - user belongs to admin role
+     * - no permission is set on the related object
+     * - permissions are set on the related object and user belongs to role that with permission
```
stefanorosanelli, line 144: Could we use also use `role` priority for that? 
F.i. if a permission is set, it is editable if:

* user has `admin` role
* user has the same role of the permission
* user has a lower priority role than the permission role

fquffio: I think this would make the logic even more knotty and mazy than it already is…

Furthermore it might just lead to an undesired effect in cases where there are independent hierarchies of roles for different sectors.

Say you have an instance with two sectors Foo and Bar. Each of these has a "Manager" and "Editor" role. You'd probably assign the permissions as follows:

```
 - Folder "Foo" (permissions to "Manager Foo")
   - Folder "Foo Posts" (permissions to "Manager Foo" and "Editor Foo")
   - (more folders)
 - Folder "Bar" (permissions to "Manager Bar")
   - Folder "Bar Posts" (permissions to "Manager Bar" and "Editor Bar")
   - (more folders)
```

If roles "Manager Foo" and "Manager Bar" have priority 10 and "Editor Foo" and "Editor Bar" have priority 20, which would make sense in this scenario, implementing the logic you suggested would lead to "Manager Foo" having access to "Bar Posts" and vice versa, which is unexpected and plain wrong.

plugins/BEdita/Core/src/Model/Entity/ObjectType.php
batopa: :thinking: I like to have in `meta` an explicit information about permission enabled/disabled rather than having to figure it out from associations.

batopa: After some thinking I'm going to remove `permissions_enabled` virtual property as suggested.

fquffio (approved):

plugins/BEdita/Core/src/Model/Table/ObjectPermissionsTable.php
stefanorosanelli: Perhaps my comment was not clear enough. 
I was not referring to the permissions rules itself (can i modify this object?) but to the permission editing (can I edit this permission?). 
We could use role `priority` to see if a certain role has enough power to edit permission for another role.
But we can implement this later, if necessary.

stefanorosanelli (approved):

PR #2008, opened 2023-05-02
Fix: full path name for service get

didoda:
This fixes `ServiceRegistry::get(string $name)` when `$name` is a fullpath, i.e. `\\MyService\Full\Path\ClassName`.

### DIFF
--- a/plugins/BEdita/Core/src/Job/ServiceRegistry.php
+++ b/plugins/BEdita/Core/src/Job/ServiceRegistry.php
@@ -56,7 +56,7 @@ class ServiceRegistry
 
         $className = $plugin . Inflector::camelize($name);
         if (strpos($name, '\\') === 0) {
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Job/ServiceRegistry.php
```diff
@@ -55,6 +55,9 @@ public static function get(string $name): JobService
         }
 
         $className = $plugin . Inflector::camelize($name);
+        if (strpos($name, '\\') === 0) {
+            $className = Inflector::camelize($name);
```
batopa: In this case we don't need to camelize the `$name` since we expect a FQN class.

batopa (changes requested):

batopa (approved):

PR #2009, opened 2023-05-08
Introduce policy for object permissions

batopa:
This PR introduce in the API a new `ObjectPolicy` used to  check if a BEdita object can be updated from the authenticated identity.

### DIFF
--- a/plugins/BEdita/API/src/Policy/ObjectPolicy.php
+++ b/plugins/BEdita/API/src/Policy/ObjectPolicy.php
@@ -1,0 +1,74 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2023 ChannelWeb Srl, Chialab Srl
+ *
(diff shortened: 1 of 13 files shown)


plugins/BEdita/API/src/Policy/ObjectPolicy.php
```diff
@@ -0,0 +1,74 @@
...
+        $permsRoles = Hash::extract((array)$object->perms, 'roles');
+        if (empty($permsRoles)) { // no permission set
+            return true;
+        }
+
+        $userRolesNames = Hash::extract($identity->getOriginalData(), 'roles.{n}.name');
+        if (empty($userRolesNames)) {
+            $userRolesNames = $this->fetchTable('Roles')
+                ->find('list')
+                ->where(['id IN' => (array)Hash::extract($identity->getOriginalData(), 'roles.{n}.id')])
```
le0m: I'd check that `roles.{n}.id` is also not empty (ex. user without any role)

le0m (approved):

PR #2010, opened 2023-05-08
Add column results to async_jobs

didoda:
This PR adds a column `results` to `async_jobs` table.

The field `results` is a json representing the process results. Its internal structure:
```json
[
    {
        "attempt_number": 1, // number of job retry
        "success": true, // true or false
        "data": { // this could be custom, depending on the job service
            "messages": ["process completed with no errors"] // an array of messages
        }
    },
    // ... can have more
]
```


plugins/BEdita/Core/src/Model/Table/AsyncJobsTable.php
```diff
@@ -328,4 +332,26 @@ protected function findPriority(Query $query, array $options)
...
+    {
+        $this->getConnection()->transactional(function () use ($entity, $success, $messages) {
+            $results = (array)$entity->get('results');
+            $attempt = count($results) + 1;
+            $data = compact('messages');
+            $result = compact('data', 'success');
+            $result['attempt_number'] = $attempt;
+            $results[] = $result;
+            $entity->set('results', $results);
+            $this->saveOrFail($entity);
```
le0m:
```suggestion
            $this->saveOrFail($entity, ['atomic' => false]);
```
To avoid running a transaction within a transaction.

le0m (approved):

PR #2011, opened 2023-05-16
Avoid login for draft user by default [v4]

didoda:
This PR fix an issue that allowed user to authenticate having status `draft`. 
The new default is only `on` and it's configurable to allow `draft` in some cases.

This PR is the v4 version of https://github.com/bedita/bedita/pull/1969


didoda:
Obsolete

PR #2012, opened 2023-05-19
Add `async_jobs` endpoint for get and post

didoda:
This PR provides a new `/async_jobs` endpoint for `GET` and `POST`.

This endpoint will be available for "non admin" users.
Admins can use `/admin/async_jobs` for `GET`, `PATCH`, `POST`, `DELETE`.

### DIFF
--- a/plugins/BEdita/API/src/Controller/AsyncJobsController.php
+++ b/plugins/BEdita/API/src/Controller/AsyncJobsController.php
@@ -34,10 +34,9 @@ class AsyncJobsController extends ResourcesController
      */
     public function initialize(): void
     {
-        parent::initialize();
-
         if (!in_array($this->request->getMethod(), ['GET', 'POST'])) {
             throw new MethodNotAllowedException();
         }
+        parent::initialize();
     }
 }
--- a/plugins/BEdita/API/src/Controller/AsyncJobsController.php
+++ b/plugins/BEdita/API/src/Controller/AsyncJobsController.php
@@ -1,7 +1,9 @@
 <?php
+declare(strict_types=1);
+
 /**
  * BEdita, API-first content management framework
- * Copyright 2021 ChannelWeb Srl, Chialab Srl
+ * Copyright 2023 Atlas Srl, Chialab Srl
  *
  * This file is part of BEdita: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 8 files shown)


plugins/BEdita/API/src/Controller/AsyncJobsController.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+namespace BEdita\API\Controller;
+
+/**
+ * Controller for `/async_jobs` endpoint.
+ *
+ * @property \BEdita\Core\Model\Table\AsyncJobsTable $AsyncJobs
+ */
+class AsyncJobsController extends AppController
```
batopa: The controller should extend `ResourceController`. It would allow to limit controller code to 

```php
/**
 * @inheritDoc
 */
public $defaultTable = 'AsyncJobs';

/**
 * @inheritDoc
 */
public function initialize(): void
{
    parent::initialize();

    if (!in_array($this->request->getMethod(), ['GET', 'POST'])) {
        throw new MethodNotAllowedException();
    }
}
```

batopa (changes requested):

stefanorosanelli (commented):
Code is ok

I have only some security concerns: this way every user will be able to create async jobs

Can we introduce a migration to allow only `admin` role to `POST /async_jobs` ? 
In a specific instance we can after change this permission rule.


Something like

```yaml
create:
  endpoints:
    - name: async_jobs

  endpoint_permissions:
    - endpoint_name: async_jobs
      role_name: admin
      application_name: null
      read: true
      write: true
```

didoda:
> Code is ok
> 
> I have only some security concerns: this way every user will be able to create async jobs
> 
> Can we introduce a migration to allow only `admin` role to `POST /async_jobs` ? In a specific instance we can after change this permission rule.
> 
> Something like
> 
> ```yaml
> create:
>   endpoints:
>     - name: async_jobs
> 
>   endpoint_permissions:
>     - endpoint_name: async_jobs
>       role_name: admin
>       application_name: null
>       read: true
>       write: true
> ```

This means:

 - core add a migration to allow `POST /async_jobs` to admins only
 - every customer api that uses Bedita Manager add a migration to allow `POST /async_jobs` to all users and application manager

I understand the security concerns, but it seems to me overkill. Though I don't have an alternative idea to provide more security

batopa (approved):

batopa:
> every customer api that uses Bedita Manager add a migration to allow POST /async_jobs to all users and application manager

Since until now async jobs were only created and read  from users with `admin` role, just instances that require to expose async jobs to different roles should update `async_jobs` endpoint permissions.

For other instances all will continue to work as expected.

stefanorosanelli (approved):

PR #2013, opened 2023-05-24
Authorize object locked by folder

batopa:
This PR adds authorization check to see if resource is locked by folder i.e. if exists at least a parent of resource entity forbidden for authenticated user.

In this case:
* the resource can't be deleted
* the resource can be modified but `status` and `uname` fields
* the resource can't be removed from protected folder

### DIFF
--- a/plugins/BEdita/API/src/Controller/ObjectsController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectsController.php
@@ -13,7 +13,6 @@
 namespace BEdita\API\Controller;
 
 use BEdita\API\Model\Action\UpdateRelatedAction;
-use BEdita\Core\Exception\LockedResourceException;
 use BEdita\Core\Model\Action\ActionTrait;
 use BEdita\Core\Model\Action\AddRelatedObjectsAction;
 use BEdita\Core\Model\Action\DeleteObjectAction;
@@ -335,7 +334,7 @@ class ObjectsController extends ResourcesController
                 }
 
                 if ($entity->get($field) !== Hash::get($data, $field)) {
-                    throw new LockedResourceException(
+                    throw new ForbiddenException(
                         __d(
                             'bedita',
                             'Cannot change "{0}" field since object {1} is locked by parent',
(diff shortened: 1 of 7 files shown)


stefanorosanelli (approved):

plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -308,6 +305,49 @@ public function resource($id)
...
+            if ($entity->get($field) !== Hash::get($data, $field)) {
+                throw new ForbiddenException(
+                    __d(
+                        'bedita',
+                        'Cannot change "{0}" field since object {1} is locked by parent',
+                        [$field, $entity->id]
+                    )
+                );
+            }
+        }
```
fquffio: Since `->can($entity, 'updateParents')` is a rather expensive check, I'd try to flip the logic in this method to actually run that check only if the request is trying to change a locked field (which is a rather fast check):

```suggestion
        $protectedFieldsChanged = array_filter(
            ['uname', 'status'],
            fn (string $field): bool => Hash::check($data, $field) && $entity->get($field) !== Hash::get($data, $field),
        );
        if (empty($protectedFieldsChanged) || $this->Authorization->can($entity, 'updateParents')) {
            return;
        }

        throw new ForbiddenException(
            __d( // Perhaps join all changed protected fields, although the message needs to be rewritten.
                'bedita',
                'Cannot change "{0}" field since object {1} is locked by parent',
                [$protectedFieldsChanged[0], $entity->id]
            )
        );
```

plugins/BEdita/API/src/Policy/ObjectPolicy.php
```diff
@@ -64,6 +66,42 @@ public function canUpdate(IdentityInterface $identity, ObjectEntity $object): bo
...
+        if ($object instanceof Folder) {
+            return $this->canUpdate($identity, $object);
+        }
+
+        $parents = $this->fetchTable('Folders')
+            ->find('available')
+            ->contain(['Permissions.Roles'])
+            ->innerJoinWith('Children', function (Query $q) use ($object) {
+                return $q->where(['Children.id' => $object->id]);
+            })
```
fquffio: [minor] Arrow functions should be supported since PHP 7.4. It would make this condition slightly more compact
```suggestion
            ->innerJoinWith('Children', fn (Query $q) => $q->where(['Children.id' => $object->id]))
```

fquffio (approved):

fquffio (approved):

fquffio (approved):

PR #2019, opened 2023-06-13
feat: search registry and adapter

le0m:
This is the first PR of a series to add the ability to have different adapters for query searches.

This PR adds a registry for search adapters and the first adapter implementation.

### DIFF
--- a/plugins/BEdita/Core/src/Search/Adapter/SimpleAdapter.php
+++ b/plugins/BEdita/Core/src/Search/Adapter/SimpleAdapter.php
@@ -1,0 +1,150 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2023 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+namespace BEdita\Core\Search\Adapter;
+
+use BEdita\Core\Exception\BadFilterException;
+use BEdita\Core\Search\BaseAdapter;
+use Cake\Database\Expression\FunctionExpression;
+use Cake\Database\Expression\QueryExpression;
+use Cake\Datasource\EntityInterface;
+use Cake\ORM\Query;
+use Cake\Utility\Hash;
+use Cake\Validation\Validator;
+
+/**
+ * Simple search adapter.
+ * Execute search via `LIKE` query.
+ *
+ * @since 5.14.0
+ */
+class SimpleAdapter extends BaseAdapter
+{
+    /**
+     * Get validator.
+     *
+     * @param array $config Search configuration
+     * @return \Cake\Validation\Validator
+     */
+    public function getValidator(array $config = []): Validator
+    {
+        $validator = new Validator();
+
+        return $validator
+            ->isArray('words')
+            ->add('words', 'checkWords', [
+                'rule' => function ($value) use ($config) {
+                    $value = (array)$value;
+                    $minLength = Hash::get($config, 'minLength');
+                    $maxWords = Hash::get($config, 'maxWords');
+                    if (count($value) === 0) {
+                        return __d('bedita', 'query strings must be at least {0} characters long', $minLength);
+                    }
+
+                    if ($maxWords > 0 && count($value) > $maxWords) {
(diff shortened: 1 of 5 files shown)


plugins/BEdita/Core/src/Search/Adapter/SimpleAdapter.php
```diff
@@ -0,0 +1,150 @@
...
+ */
+class SimpleAdapter extends BaseAdapter
+{
+    /**
+     * Get validator.
+     *
+     * @param array $config Search configuration
+     * @return \Cake\Validation\Validator
+     */
+    public function getValidator(array $config = []): Validator
```
fquffio: Since this method is not part of the public interface, I'd make it protected instead.

```suggestion
    protected function getValidator(array $config = []): Validator
```

plugins/BEdita/Core/src/Search/Adapter/SimpleAdapter.php
```diff
@@ -0,0 +1,150 @@
...
+                    ['%', '\\'],
+                    ['\\%', '\\\\'],
+                    mb_strtolower($word)
+                );
+            },
+            array_filter( // Filter out words that are too short.
+                $words,
+                function ($word) use ($minLength) {
+                    return mb_strlen($word) >= $minLength;
+                }
```
fquffio:
```suggestion
                fn (string $word): bool => mb_strlen($word) >= $minLength,
```

plugins/BEdita/Core/src/Search/Adapter/SimpleAdapter.php
```diff
@@ -0,0 +1,150 @@
...
+            $words = preg_split('/\W+/', $text); // Split words.
+        }
+        $words = array_unique(array_map( // Escape `%` and `\` characters in words.
+            function ($word) {
+                return str_replace(
+                    ['%', '\\'],
+                    ['\\%', '\\\\'],
+                    mb_strtolower($word)
+                );
+            },
```
fquffio: Although this is taken from the previous implementation, I'd skip lowercasing the string here. The database already compares string respectfully of the collation set on the table: if a case-insensitive match is desired, a case-insensitive collation should be used.

Additionally, we could use an arrow-function here:

```suggestion
            fn (string $word): string => str_replace(
                ['%', '\\'],
                ['\\%', '\\\\'],
                $word,
            ),
```

plugins/BEdita/Core/src/Search/Adapter/SimpleAdapter.php
```diff
@@ -0,0 +1,150 @@
...
+        $table = $query->getRepository();
+        foreach (array_keys($tableFields) as $field) {
+            $fields[] = $query->func()->coalesce([
+                $table->aliasField($field) => 'identifier',
+                '',
+            ]);
+            $fields[] = ' '; // Add a spacer.
+        }
+        array_pop($fields); // Remove last spacer.
+        $field = new FunctionExpression('LOWER', [$query->func()->concat($fields)]);
```
fquffio: Similarly to my previous consideration, I think wrapping with `LOWER()` is redundant:

```suggestion
        $field = $query->func()->concat($fields);
```

plugins/BEdita/Core/tests/TestCase/Search/Adapter/SimpleAdapterTest.php
```diff
@@ -0,0 +1,208 @@
...
+     * @covers ::search()
+     * @covers ::prepareText()
+     * @covers ::getValidator()
+     */
+    public function testSearch($expected, string $text, array $options = [], $tableName = 'FakeAnimals', array $fields = [])
+    {
+        if ($expected instanceof \Exception) {
+            $this->expectException(get_class($expected));
+            $this->expectExceptionCode($expected->getCode());
+            $this->expectExceptionMessage($expected->getMessage());
```
fquffio:
```suggestion
            $this->expectExceptionObject($expected);
```

fquffio (approved):

plugins/BEdita/Core/src/Search/Adapter/SimpleAdapter.php
```diff
@@ -0,0 +1,146 @@
...
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+namespace BEdita\Core\Search\Adapter;
+
+use BEdita\Core\Exception\BadFilterException;
+use BEdita\Core\Search\BaseAdapter;
+use Cake\Database\Expression\FunctionExpression;
```
fquffio: Unused after review. Sorry.

```suggestion
```

PR #2023, opened 2023-06-26
Use search adapters in SearchableBehavior

batopa:
This PR integrates the use of search adapters in `SearchableBehavior`.

To keep code clean `SearchableBehavior` isn't initialized with previous configuration (`fields`, ...) anymore because that configuration was valid for simple search only. That configuration was moved in `SimpleAdapter`.

Now  `fields` and other configurations are set taking adavantage of `SearchAdapter.initialize` event fired from `SearchableBehavior` when search adapters are loaded.

For backward compatibility the `SearchableBehavior` can be initialized with the old config yet but it's discouraged.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
@@ -118,7 +118,7 @@ class SearchableBehavior extends Behavior
      */
     protected function getAdapter(?string $name = null): BaseAdapter
     {
-        $name ??= 'default';
+        $name ??= (string)Configure::read('Search.use', 'default');
         $searchRegistry = $this->getSearchRegistry();
         if ($searchRegistry->has($name)) {
             return $searchRegistry->get($name);
@@ -159,8 +159,7 @@ class SearchableBehavior extends Behavior
 
         unset($options[0], $options['string']);
 
-        return $this->getAdapter(Configure::read('Search.use'))
-            ->search($query, $text, $options);
+        return $this->getAdapter()->search($query, $text, $options);
     }
 
     /**
@@ -187,7 +186,7 @@ class SearchableBehavior extends Behavior
             // `fields` key in SimpleAdapter is changed.
             // It is now a list of fields without unused priority.
             if ($key === 'fields') {
-                deprecationWarning('"fields" must be a list of fields. Unused priorities have been removed.');
+                deprecationWarning('"fields" must be a list of strings. Unused priorities have been removed.');
                 $conf = array_keys($conf);
             }
 
--- a/plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
+++ b/plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
@@ -1,4 +1,6 @@
 <?php
+declare(strict_types=1);
+
 /**
  * BEdita, API-first content management framework
  * Copyright 2017 ChannelWeb Srl, Chialab Srl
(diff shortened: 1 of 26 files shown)


plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
```diff
@@ -29,84 +32,93 @@
...
+     * @return void
      */
-    public function initialize(array $config): void
+    public function afterSave(EventInterface $event, EntityInterface $entity): void
     {
-        foreach (['columnTypes', 'fields'] as $key) {
-            if (isset($config[$key])) {
-                $this->setConfig($key, $config[$key], false);
-            }
+        foreach (array_keys((array)Configure::read('Search.adapters')) as $name) {
```
fquffio: Since this might become a common pattern should we extract a method `public function getAdapters(): string[]`?

Or perhaps a method to iterate through adapters altogether?

```php
/**
 * @return iterable<string, \BEdita\Core\Search\BaseAdapter>
 */
public function getAdapters(): iterable
{
    foreach (array_keys((array)Configure::read('Search.adapters')) as $name) {
        yield $name => $this->getAdapter($name);
    }
}
```

plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
```diff
@@ -29,84 +32,93 @@
...
     /**
-     * Get searchable fields and their priorities.
+     * Get an adapter by name.
      *
-     * @return array Array where keys are columns, and values are priorities.
+     * @param string $name The adapter name
+     * @return \BEdita\Core\Search\BaseAdapter
      */
-    public function getFields()
+    protected function getAdapter($name): BaseAdapter
```
fquffio:
```suggestion
    protected function getAdapter(string $name): BaseAdapter
```

I also wonder if we should allow `null` and fallback to the "default" adapter, for searching purposes:

```php
$myAdapter = $this->getAdapter('foo');
$defaultAdapter = $this->getAdapter(); // or $this->getAdapter(null);
```

plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
```diff
@@ -121,78 +133,50 @@ public function findQuery(Query $query, array $options)
...
+        $adapter->setConfig($config);
+
+        // Config keys that must be overridden
+        foreach (['columnTypes', 'fields'] as $key) {
+            $conf = $this->getConfig($key);
+            if (!is_array($conf)) {
+                continue;
+            }
+
+            // `fields` key in SimpleAdapter is changed.
```
fquffio, line 186: Maybe we could raise a `deprecationWarning('blahblahblah')` when we reach this point?

fquffio (approved):

plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
batopa: Good idea!

plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
```diff
@@ -29,84 +32,107 @@
...
-        $allFields = $this->getAllFields($this->table());
-
-        $fields = array_intersect_key($fields, array_flip($allFields));
-        if ($wildCard !== null) {
-            // If wildcard `*` is present, all other fields have default priority.
-            $fields += array_diff_key(
-                array_fill_keys($allFields, $wildCard),
-                $fields
-            );
+        $name ??= 'default';
```
fquffio:
```suggestion
        $name ??= (string)Configure::read('Search.use', 'default');
```

fquffio (approved):

plugins/BEdita/Core/src/Model/Behavior/SearchableBehavior.php
batopa: Right! We can also modify in `findQuery()`

```php
return $this->getAdapter(Configure::read('Search.use'))
            ->search($query, $text, $options);
 ```
 
 with 
 
 ```php
 return $this->getAdapter()
            ->search($query, $text, $options);
 ```
:+1:

fquffio (approved):

PR #2031, opened 2023-08-02
Admin sysinfo endpoint

didoda:
This PR provides a new `/admin/sysinfo` endpoint (admin only).

`GET /admin/sysinfo` response example:
```
{
    "links": {
        "self": "https://api.example.com/admin/sysinfo",
        "home": "https://api.example.com/home"
    },
    "meta": {
        "info": {
            "Url": "https://api.example.com",
            "Version": "5.13.9",
            "CakePHP": "4.4.13",
            "PHP": "8.2.8",
            "Operating System": "Linux 5.19.0-50-generic #50-Ubuntu SMP PREEMPT_DYNAMIC Mon Jul 10 18:24:29 UTC 2023 x86_64",
            "PHP Server API": "apache2handler",
            "Extensions": [
                "Core",
                "date",
                "libxml",
                "openssl",
                "pcre",
                "zlib",
                "filter",
                "hash",
                "json",
                "random",
                "Reflection",
                "SPL",
                "session",
                "standard",
                "sodium",
                "apache2handler",
                "mysqlnd",
                "PDO",
                "xml",
                "bcmath",
                "bz2",
                "calendar",
                "ctype",
                "curl",
                "dom",
                "mbstring",
                "FFI",
                "fileinfo",
                "ftp",
                "gd",
                "gettext",
                "iconv",
                "igbinary",
                "intl",
                "exif",
                "mysqli",
                "pdo_mysql",
                "pdo_sqlite",
                "Phar",
                "posix",
                "readline",
                "redis",
                "shmop",
                "SimpleXML",
                "sockets",
                "sqlite3",
                "sysvmsg",
                "sysvsem",
                "sysvshm",
                "tokenizer",
                "xmlreader",
                "xmlwriter",
                "xsl",
                "zip",
                "Zend OPcache",
                "xdebug"
            ],
            "Extensions info": [
                "Zend OPcache",
                "Xdebug"
            ],
            "Memory limit": "128M",
            "Post max size": "50M",
            "Upload max size": "40M"
        }
    }
}
```

### DIFF
--- a/plugins/BEdita/API/config/routes.php
+++ b/plugins/BEdita/API/config/routes.php
@@ -171,7 +171,7 @@ return function (RouteBuilder $routes) {
         $routes->connect(
             '/admin/sysinfo',
             ['controller' => 'Sysinfo', 'action' => 'index'],
-            ['_name' => 'sysinfo']
+            ['_name' => 'admin:sysinfo']
         );
 
         // Modeling endpoints.
--- a/plugins/BEdita/API/config/routes.php
+++ b/plugins/BEdita/API/config/routes.php
@@ -171,7 +171,7 @@ return function (RouteBuilder $routes) {
                 $routes->connect(
                     '/sysinfo',
                     ['controller' => 'Sysinfo', 'action' => 'index'],
-                    ['_name' => 'admin:sysinfo']
+                    ['_name' => 'sysinfo']
                 );
             }
         );
--- a/plugins/BEdita/API/config/routes.php
+++ b/plugins/BEdita/API/config/routes.php
@@ -64,13 +64,6 @@ return function (RouteBuilder $routes) {
             ['_name' => 'status']
         );
 
-        // Sysinfo.
-        $routes->connect(
-            '/sysinfo',
-            ['controller' => 'Sysinfo', 'action' => 'index'],
-            ['_name' => 'sysinfo']
-        );
-
         // Login.
         $routes->post(
             '/auth',
@@ -173,6 +166,12 @@ return function (RouteBuilder $routes) {
                 '_namePrefix' => 'admin:',
             ],
             $resourcesRoutes($adminControllers)
+        );
+
+        $routes->connect(
+            '/admin/sysinfo',
+            ['controller' => 'Sysinfo', 'action' => 'index'],
+            ['_name' => 'sysinfo']
         );
 
         // Modeling endpoints.
--- a/plugins/BEdita/API/config/routes.php
+++ b/plugins/BEdita/API/config/routes.php
@@ -165,13 +165,15 @@ return function (RouteBuilder $routes) {
             [
                 '_namePrefix' => 'admin:',
             ],
-            $resourcesRoutes($adminControllers)
-        );
-
-        $routes->connect(
-            '/admin/sysinfo',
-            ['controller' => 'Sysinfo', 'action' => 'index'],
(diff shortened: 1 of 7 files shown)


fquffio:
Exposing such detailed information about the underlying system would not be a good thing in a production environment, generally speaking, as it would provide details to those looking for a possible surface of attack (e.g.: one might look up the known CVEs for the current PHP version and OS).

Although "security by obscurity" is not actually _security_ by itself, it usually helps and is considered a good idea not to make your system details public.

While I see that this endpoint could be useful, I think it should either be enabled only upon request (i.e. explicitly enabling a configuration in cases where APIs are not publicly reachable, and can only be queried by trusted software, like the manager app) or require an authenticated user with role `admin`. Leaving this open or requiring only an API key doesn't look quite right to me…

A good explanation of restricting access to an URL that exposes sensible information about your system can be found for the [PHP-FPM status endpoint](https://www.php.net/manual/en/fpm.status.php), which falls more or less in the same category.

didoda:
> Exposing such detailed information about the underlying system would not be a good thing in a production environment, generally speaking, as it would provide details to those looking for a possible surface of attack (e.g.: one might look up the known CVEs for the current PHP version and OS).
> 
> Although "security by obscurity" is not actually _security_ by itself, it usually helps and is considered a good idea not to make your system details public.
> 
> While I see that this endpoint could be useful, I think it should either be enabled only upon request (i.e. explicitly enabling a configuration in cases where APIs are not publicly reachable, and can only be queried by trusted software, like the manager app) or require an authenticated user with role `admin`. Leaving this open or requiring only an API key doesn't look quite right to me…
> 
> A good explanation of restricting access to an URL that exposes sensible information about your system can be found for the [PHP-FPM status endpoint](https://www.php.net/manual/en/fpm.status.php), which falls more or less in the same category.

good point. "admin only" endpoint sounds great

plugins/BEdita/API/config/routes.php
```diff
@@ -64,6 +64,13 @@
             ['_name' => 'status']
         );
 
+        // Sysinfo.
+        $routes->connect(
```
stefanorosanelli: Why not `/admin/sysinfo` ? The `/admin` prefix makes it more clear that this is admin stuff

`SysInfoController` should then be moved to `Admin` namespace

plugins/BEdita/API/src/Controller/SysinfoController.php
```diff
@@ -0,0 +1,64 @@
...
+ * @since 5.13.9
+ */
+class SysinfoController extends AppController
+{
+    /**
+     * @inheritDoc
+     */
+    public function initialize(): void
+    {
+        $htmlRequest = (Configure::read('debug') || Configure::read('Accept.html')) && $this->request->is('html');
```
stefanorosanelli: I guess this has been inspired by `/status` controller, but since this endpoint is not meant to be easily integrated (`/status` may be used by external monitoring tools) - this lines are probably not necessary

We could keep just 

```php
        parent::initialize();

        $this->request = $this->request->withAttribute(EndpointPolicy::ADMINISTRATOR_ONLY, true);
```

plugins/BEdita/API/config/routes.php
didoda: ok for `/admin/sysinfo`

stefanorosanelli (approved):

PR #2034, opened 2023-09-18
`build_search_index` command

didoda:
This PR introduces the `BuildSearchIndexCommand` / `build_search_index`.

Usage
--------

```bash
$ bin/cake build_search_index --help
Command to reindex objects for search.

Usage:
cake build_search_index [options]

Options:

--adapter      Reindex only using one or more specific adapters.
--help, -h     Display this help.
--id           Reindex only one or more specific objects by ID.
--quiet, -q    Enable quiet output.
--type         Reindex only objects from one or more specific types.
--uname        Reindex only one or more specific objects by uname.
--verbose, -v  Enable verbose output.
```

Usage examples follow:

```bash
$ bin/cake build_search_index
$ bin/cake build_search_index --type documents,profiles
$ bin/cake build_search_index --id 1742,5012
$ bin/cake build_search_index --uname dummy-article,something-else
$ bin/cake build_search_index --adapter "BEdita\Core\Search\Adapter\SimpleAdapter"
```

Basic usage (no verbose) example:
```bash
$ bin/cake build_search_index

Building search index

This may take a while, please wait... (Ctrl+C to abort)

Done. Summary:

> documents: 6
> events: 2
> folders: 13
> users: 9
```

Verbose mode example:
```bash
$ bin/cake build_search_index --verbose
Building search index

This may take a while, please wait... (Ctrl+C to abort)

Indexing audio

> audio: 0

Indexing documents

> ID 18 [test-epwefewfewfewf] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 21 [doc-di-prova] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 24 [teorema-di-pitagora] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 27 [algebra-lineare] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 33 [documento-orfano] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 47 [test] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> documents: 6

Indexing events

> ID 31 [test-0820c4] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 48 [test-d06dae] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> events: 2

Indexing files

> files: 0

Indexing folders

> ID 3 [subjects] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 4 [matematica] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 5 [storia] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 6 [filosofia] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 25 [geometria] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 26 [algebra] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 34 [estetica] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 35 [etica] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 36 [logica] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 43 [la-rivoluzione-francese] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 44 [la-prima-guerra-mondiale] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 45 [la-seconda-guerra-mondiale] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 46 [altra-cartella] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> folders: 13

Indexing images

> images: 0

Indexing links

> links: 0

Indexing locations

> locations: 0

Indexing news

> news: 0

Indexing profiles

> profiles: 0

Indexing publications

> publications: 0

Indexing users

> ID 1 [bedita] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 2 [__deleted-2] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 8 [__deleted-8] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 37 [gustavo-matematica] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 38 [gustavo-storia] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 39 [gustavo-viewer] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 40 [gustavo-manager] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 41 [gustavo-editor] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> ID 42 [user-gustavo-admin] [Adapter: BEdita\Core\Search\Adapter\SimpleAdapter]
> users: 9

Indexing videos

> videos: 0

Done. Summary:

> documents: 6
> events: 2
> folders: 13
> users: 9
```


plugins/BEdita/Core/src/Command/SearchCommand.php
```diff
@@ -0,0 +1,268 @@
...
+            'required' => false,
+        ]);
+        $parser->addOption('delete', [
+            'help' => 'Delete an object from index.',
+            'required' => false,
+        ]);
+        $parser->addOption('clear', [
+            'help' => 'Remove index for all or multiple objects.',
+            'required' => false,
+        ]);
```
fquffio: Personally I think these options should probably be changed.

Firstly, the `--clear` option shouldn't exist in my opinion: there are far more efficient ways to clear an index than looping over every single item. Also, since this operation is non-atomic, there is absolutely no guarantee that the index is empty after this command completes, as entities might have been reindexed as a consequence of API operations or other commands in the meanwhile.

If I understand well, the idea behind the `--delete` flag is to provide a way to remove a leftover indexed document from the index in case the search index becomes misaligned with the relational database. However, due to how it is designed, it would only work for objects that are _still_ present in the database, as they need to be loaded as BEdita objects in the first place, which (unless I'm mistaken) makes this feature much less useful, and theoretically if the adapter is configured correctly is the same thing that a normal reindex would do — if it's not, the object would be re-added to the index the first time it is saved anyway. If my assumptions are correct, I'd implement this feature as a separate command… or more likely I'd implement an adapter-specific shell to resolve possible issues that are specific to that adapter.

In other words, I'd leave just the "index" and "reindex". However, I'd probably avoid treating two very similar behaviours as two distinct flags. Instead I'd find more useful to have this shell reindex a bunch of objects, and add flags to filter the objects to be reindexed. For example, the flags to this command might be:

- `--type` **(multiple)**: reindex only objects from one or more specific types
- `--id` **(multiple)**: reindex only one or more specific objects by ID
- `--uname` **(multiple)**: same as `--id`, but using the object(s) unique names
- `--ancestor`: reindex only objects placed within a sub-tree (ID or `uname` of a folder, reindex its descendants)
- `--since`: reindex only objects modified or created _after_ the requested timestamp
- `--until`: reindex only objects modified or created _before_ the requested timestamp

(I just wrote down all the flags that came to my mind… to start with, `--type` and `--id` would be more than fine.)

Additionally, if possible I'd add a flag `--adapter` **(multiple)** to reindex only on the given set of adapters: this might be useful to populate a new index without messing with the other ones.

plugins/BEdita/Core/src/Command/SearchCommand.php
```diff
@@ -0,0 +1,268 @@
...
+    {
+        $types = array_filter(explode(',', (string)$args->getOption($cmdOperation)));
+        if (empty($types)) {
+            $result = $this->fetchTable('ObjectTypes')->find()->where(['enabled' => true])->toArray();
+            $types = (array)Hash::extract($result, '{n}.name');
+        }
+        foreach ($types as $type) {
+            $io->out(sprintf('Perform "%s" on type "%s"', $cmdOperation, $type));
+            $table = $this->fetchTable($type);
+            $query = $table->find('type', [$type])->where(['deleted' => false]);
```
fquffio: This might load tens of thousands of objects into memory all at once. I'd prefer implementing a pagination here…

plugins/BEdita/Core/src/Command/SearchCommand.php
```diff
@@ -0,0 +1,268 @@
...
+    public function execute(Arguments $args, ConsoleIo $io): ?int
+    {
+        $tmp = array_intersect_key($args->getOptions(), array_flip($this->operations));
+        $operation = empty($tmp) ? '' : (string)array_key_first($tmp);
+        $message = empty($operation) ? $this->getOptionParser()->help() : 'Perform ' . $operation . ' operation';
+        $io->out($message);
+        $this->Objects = $this->fetchTable('Objects');
+        $this->dryrun = $args->getOption('dry-run') !== null;
+
+        return empty($operation) ? Command::CODE_ERROR : $this->{$operation}($args, $io);
```
fquffio: If you like my suggestions per my previous comment, this method might become something like this:

```php
public function execute(Arguments $args, ConsoleIo $io): int
{
    foreach ($this->objectsIterator($args) as $object) {
        $this->doIndexResource($object, $args, $io); // Signature changed to filter adapters, if desired.
    }

    // ...
}

protected function objectsIterator(Arguments $args): iterable
{
    $table = $this->fetchTable('Objects');
    $query = $table->find()->orderAsc($table->aliasField('id'))->limit(200);
    if (!empty($args->getOption('type')) {
        $query = $query->find('type', (array)$args->getOption('type'));
    }
    // More filters...

    $lastId = 0;
    while (true) {
        $q = clone $query;
        $q = $q->where(fn (QueryExpression $exp): QueryExpression => $exp->gt($table->aliasField('id'), $lastId));
        $results = $q->all();
        if ($results->isEmpty()) {
            break;
        }

        foreach ($results as $entity) {
            $lastId = $entity->id;

            yield $entity;
        }
    }
}
```

plugins/BEdita/Core/src/Command/SearchCommand.php
didoda: Thanks for the detailed analysis. I renamed the class to `BuildSearchIndexCommand`. I added all the options you suggested, except `--ancestor`, `--since` and `--until` (they could be done separetely, in the future).

Good idea the flag `--adapter`. We have it available as well, now.

plugins/BEdita/Core/src/Command/SearchCommand.php
didoda: I used your code and changed just a bit, to have type iteration (reindexing by type order, even when no arguments are passed to the command).

Funny note: your `objectIterator` realizes an infinite loop, because results is fetched with the wrong variable (`$query` instead of `$q`); it took me some time to "see" it :-)

fquffio: Oops! Sorry about that! 😂

PR #2044, opened 2024-01-18
Add extra associations in object entity

paolosantarsiero:
This PR introduce possibility to add extra associations to object entity.

### DIFF
--- a/plugins/BEdita/Core/src/I18n/MessagesFileLoader.php
+++ b/plugins/BEdita/Core/src/I18n/MessagesFileLoader.php
@@ -45,7 +45,7 @@ class MessagesFileLoader extends BaseLoader
     /**
      * {@inheritDoc}
      *
-     * @param string[] $plugins Additional plugins to look up in for translations.
+     * @param string[] $name Additional plugins to look up in for translations.
      */
     public function __construct(string $name, string $locale, string $extension = 'po', array $plugins = [])
     {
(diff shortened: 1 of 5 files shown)


plugins/BEdita/Core/src/I18n/MessagesFileLoader.php
```diff
@@ -45,7 +45,7 @@ public function plugins(): array
     /**
      * {@inheritDoc}
      *
-     * @param string[] $plugins Additional plugins to look up in for translations.
+     * @param string[] $name Additional plugins to look up in for translations.
```
didoda: complete signature is `__construct(string $name, string $locale, string $extension = 'po', array $plugins = [])`.

The additional plugins to look up are `array $plugins`. I guess this change is wrong.

`@inheritDoc` provides the inheritance of the basic signature php doc, `$plugins` is the extra parameter. Right?

plugins/BEdita/Core/src/Model/Action/ListAssociatedAction.php
```diff
@@ -292,8 +292,8 @@ public function execute(array $data = [])
      * Get association sort by association and primary key.
      * When association name is "Children", use Folders.getSort($primaryKey).
      *
-     * @param mixed $primaryKey Primary key
-     * @param \Cake\ORM\Association $association Association
+     * @param mixed $association Primary key
+     * @param \Cake\ORM\Association $primaryKey Association
```
batopa:
```suggestion
     * @param \Cake\ORM\Association $association Association
     * @param  mixed $primaryKey Primary key
```

batopa (approved):

PR #2049, opened 2024-02-07
`objects_delete` command

### LINKED ISSUE #2047: Objects "hard" delete command
Objects can be softly deleted. It would be handy a command to hard delete objects, considering specific logics.

Examples:
```
// delete all users in trashcan since more than 3 months
$ bin/cake objects_delete --type users --since '-3 months'

// delete all documents in trashcan since more than 1 week
$ bin/cake objects_delete --type documents --since '-1 weeks'
```

didoda:
This PR resolves https://github.com/bedita/bedita/issues/2047

Command usage:
```bash
// delete all users in trashcan since more than 3 months
$ bin/cake objects_delete --type users --since '-3 months'

// delete all documents and events in trashcan since more than 1 week
$ bin/cake objects_delete --type documents,events --since '-1 weeks'
```

This also fixes a dependency issue with `symfony/console: ^7`, by adding a conflict rule in composer.json.

### DIFF
--- a/plugins/BEdita/Core/src/Command/ObjectsDeleteCommand.php
+++ b/plugins/BEdita/Core/src/Command/ObjectsDeleteCommand.php
@@ -52,8 +52,8 @@ class ObjectsDeleteCommand extends Command
      */
     public function execute(Arguments $args, ConsoleIo $io)
     {
-        $since = $args->getOption('since') ?? '-1 month';
-        $type = $args->getOption('type');
+        $since = $args->getOption('since');
+        $types = (array)$args->getOption('type');
         $message = 'Deleting from trash objects, since ' . $since;
         $message .= $type ? ', for type ' . $type : '';
         $io->info($message);
--- a/plugins/BEdita/Core/src/Command/ObjectsDeleteCommand.php
+++ b/plugins/BEdita/Core/src/Command/ObjectsDeleteCommand.php
@@ -55,20 +55,22 @@ class ObjectsDeleteCommand extends Command
         $since = $args->getOption('since');
         $types = (array)$args->getOption('type');
         $message = 'Deleting from trash objects, since ' . $since;
-        $message .= $type ? ', for type ' . $type : '';
+        $message .= $types ? ', for type ' . implode(',', $types) : '';
         $io->info($message);
         $conditions = ['deleted' => true, 'locked' => false, 'modified <' => new FrozenDate($since)];
         $deleted = $errors = 0;
-        foreach ($this->objectsIterator($type, $conditions) as $object) {
-            try {
-                $io->verbose(sprintf('Deleting object %s', $object->id));
-                $object->getTable()->deleteOrFail($object);
-            } catch (\Throwable $e) {
-                $io->error(sprintf('Error deleting object %s: %s', $object->id, $e->getMessage()));
-                $errors++;
-                continue;
+        foreach ($types as $type) {
+            foreach ($this->objectsIterator($type, $conditions) as $object) {
+                try {
+                    $io->verbose(sprintf('Deleting object %s', $object->id));
+                    $object->getTable()->deleteOrFail($object);
+                } catch (\Throwable $e) {
+                    $io->error(sprintf('Error deleting object %s: %s', $object->id, $e->getMessage()));
+                    $errors++;
+                    continue;
+                }
+                $deleted++;
             }
-            $deleted++;
         }
         $io->success(sprintf('Deleted from trash %d objects [%d errors]', $deleted, $errors));
         $io->info('Done');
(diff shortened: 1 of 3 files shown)


plugins/BEdita/Core/src/Command/ObjectsDeleteCommand.php
```diff
@@ -0,0 +1,104 @@
...
+    {
+        return $parser
+            ->addOption('since', [
+                'help' => 'Delete objects in trash since this date',
+                'required' => false,
+            ])
+            ->addOption('type', [
+                'help' => 'Delete objects in trash by type',
+                'required' => false,
+            ]);
```
fquffio: It should be possible to set the default in the option parser itself ([docs](https://book.cakephp.org/4/en/console-commands/option-parsers.html#using-options)). Additionally, it would be cool to have `--type` flag be allowed multiple times, and as such treated as an array in the code:

```suggestion
                'required' => false,
                'default' => '-1 month',
            ])
            ->addOption('type', [
                'help' => 'Delete objects in trash by type',
                'required' => false,
                'multiple' => true,
            ]);
```

plugins/BEdita/Core/src/Command/ObjectsDeleteCommand.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function execute(Arguments $args, ConsoleIo $io)
+    {
+        $since = $args->getOption('since') ?? '-1 month';
+        $type = $args->getOption('type');
```
fquffio: If the default is set via option parser, the `??` is redundant. Additionally, if `--type` flag is allowed more than once, `$type` becomes an array (casting is redundant, but helps PHPStan):

```suggestion
        $since = $args->getOption('since');
        $types = (array)$args->getOption('type');
```

plugins/BEdita/Core/src/Command/ObjectsDeleteCommand.php
```diff
@@ -0,0 +1,125 @@
...
+     *
+     * @param string|null $type The object type
+     * @param array $conditions The conditions
+     * @return iterable
+     */
+    private function objectsIterator(?string $type, array $conditions): iterable
+    {
+        $table = $this->fetchTable('Objects');
+        $query = empty($type) ? $table->find() : $table->find('type', [$type]);
+        $query = $query->where($conditions)->limit(200);
```
batopa, line 109: Maybe it's better to sort by id asc since it is used to fetch results where  `'id' > $lastId` in the while cycle below.

plugins/BEdita/Core/tests/TestCase/Command/ObjectsDeleteCommandTest.php
```diff
@@ -0,0 +1,125 @@
...
+        $deleted = $errors = 0;
+        $io = $this->getMockBuilder('Cake\Console\ConsoleIo')->getMock();
+        $object = new ObjectEntity([]);
+        $command = new ObjectsDeleteCommand();
+        $reflectionClass = new \ReflectionClass($command);
+        $method = $reflectionClass->getMethod('deleteObject');
+        $method->setAccessible(true);
+        $method->invokeArgs($command, [$io, $object, &$deleted, &$errors]);
+        $this->assertEquals(0, $deleted);
+        $this->assertEquals(1, $errors);
```
batopa: In general instead of change the method visibility to test it, I think it would be better testing it throught the expected flow.

So, for example, you could count trash items before run the command and count items after the execution testing the expected result in different situations (using `since`, `type` options, ...)

didoda: good point. in this case it's difficult to test it (covering all `deleteObject`, including `catch`) because the exception cannot be raised, unless you pass a wrong entity to the private method.

```php
foreach ($this->objectsIterator($objectType, $conditions) as $object) {
    $this->deleteObject($io, $object, $deleted, $errors);
}
```
This code won't raise exception, because `objectsIterator` always return entities that can be deleted.

batopa: You could try to reach that condition with a listener on 'Model.beforeDelete`.
Something as

```php
$throwError = function () {
    throw new \Exception('Fake error');
};
// add listener to global event manager
EventManager::instance()->on('Model.beforeDelete', $throwError);

// count trash items

// run command

// ensure to off listener from global event manager
EventManager::instance()->off('Model.beforeDelete', $throwError);

// count new trash items

// test assertion
```

didoda: done

batopa (approved):

PR #2068, opened 2024-03-19
OAuth2: verify token via callback and add authenticator/identifier config

stefanorosanelli:
This PR improves OAuth2 authentication process supporting providers like Apple that requires a different approach.

User verification is not performed by calling the OAuth2 provider with an access token, but via JWT `id_token` decrypting and validation.

* OAuth2 identifier and authenticator classes can now be configured via `config.authenticator` in `auth_providers.params` JSON field 
* if a `credentials_callback` callable has been defined in `auth_providers.params` this callback is used to verify the authenticated used otherwise the usual OAuth2 provider call is performed

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Identifier/OAuth2IdentifierTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Identifier/OAuth2IdentifierTest.php
@@ -181,5 +181,13 @@ class OAuth2IdentifierTest extends TestCase
         ]);
 
         static::assertNotEmpty($result);
+
+        $result = $identifier->identify([
+            'auth_provider' => 'example',
+            'provider_username' => 'another_user',
+            'access_token' => 'very-log-string',
+        ]);
+
+        static::assertNull($result);
     }
 }
--- a/plugins/BEdita/API/tests/TestCase/Identifier/OAuth2IdentifierTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Identifier/OAuth2IdentifierTest.php
@@ -153,4 +153,33 @@ class OAuth2IdentifierTest extends TestCase
         }
         static::assertSame($expected, $result);
     }
+
+    /**
+     * Test `identify` method with callback.
+     *
+     * @return void
+     * @covers ::identify()
+     */
+    public function testIdentifyWithCallback(): void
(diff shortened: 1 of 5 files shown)


batopa (approved):

plugins/BEdita/API/tests/TestCase/Identifier/OAuth2IdentifierTest.php
```diff
@@ -153,4 +153,33 @@ public function testIdentify(?array $expected, array $credentials, array $oauthR
...
+        $identifier = new OAuth2Identifier();
+        $identifier->setConfig(compact('authProvider'));
+
+        $result = $identifier->identify([
+            'auth_provider' => 'example',
+            'provider_username' => 'first_user',
+            'access_token' => 'very-log-string',
+        ]);
+
+        static::assertNotEmpty($result);
```
didoda, line 183: This looks fine. We should provide another test for `return false callback`, and `assertNull`

plugins/BEdita/Core/tests/TestCase/Model/Action/SignupUserActionTest.php
```diff
@@ -352,6 +352,43 @@ public function testExecuteExtAuth($expected, array $data, array $oauthResponse)
...
+            'username' => 'testsignup',
+            'email' => 'testsignup@example.com',
+            'auth_provider' => 'example',
+            'provider_username' => 'not-found',
+            'provider_userdata' => [],
+            'access_token' => 'incredibly-long-string',
+        ];
+        $action = new SignupUserAction();
+        $result = $action(compact('data'));
+        static::assertTrue((bool)$result);
```
didoda, line 379: We should test also the `UnauthorizedException` case

stefanorosanelli: Done

plugins/BEdita/API/tests/TestCase/Identifier/OAuth2IdentifierTest.php
stefanorosanelli: Done

didoda (approved):

PR #2070, opened 2024-03-25
Handle svg in thumbs

didoda:
This provides an update in `/media/thumbs` and `/media/thumbs/{id}`: when stream's mime type is `image/svg+xml`, no thumb is requested.


plugins/BEdita/API/src/Controller/MediaController.php
```diff
@@ -112,6 +112,12 @@ public function thumbs()
             ->map(function (Stream $stream) use ($options, $preset) {
                 $id = $stream->object_id;
                 $uuid = $stream->uuid;
+                if ($stream->mime_type === 'image/svg+xml') {
```
batopa: I would move this check in `Thumbnail::get()` after `private_url` check.
In this way it is consistent with handle of `private_url` on streams.

Another note. I'm not sure we should always skip the generation of thumbnails of svg. I think it could depend by the thumbnail generator used.

batopa (changes requested):

plugins/BEdita/API/src/Controller/MediaController.php
didoda: Ok for the first part (moving logic to `Thumbnail::get()`.

This PR wanted to fix a buggy behaviour, with most thumbnail generators... Which generator could be fine with svg?

batopa: Theoretically using `GlideGenerator` with imagemagick driver should work, but I'm not sure. Anyway you could develop a custom generator to do that.

I think it could be useful in the situation of big dimensions SVG when you need to display a little preview.

Maybe at the moment we could skip the generation thinking about that if we needed.

plugins/BEdita/Core/src/Filesystem/Thumbnail.php
```diff
@@ -106,6 +106,16 @@ public static function get(Stream $stream, $options = 'default'): array
...
             ];
         }
+
+        if ($stream->mime_type === 'image/svg+xml') {
+            $ready = true;
+            $url = $stream->get('url');
+            $id = $stream->object_id;
+            $uuid = $stream->uuid;
+
+            return compact('ready', 'url', 'id', 'uuid');
```
batopa: `id` and `uuid` are set in `MediaController`
```suggestion
            return [
                'ready' => true,
                'url' => $stream->get('url'),
            ];
```

batopa (approved):

PR #2071, opened 2024-04-09
Categories and tags structure to store translations

didoda:
This PR introduces categories and tags structure changes to provide translations.

Some changes for `Categories` and `Tags` tables:

- adding a new JSON field `labels`
- copying data from `label` into `labels.default`
- removing field `label`
- `label` becomes a virtual field

We expect that a single category or tag translation is stored in `labels` field, with a value like the following example:
```
{
  "default": "Musica",
  "de": "Musik",
  "en": "Music",
  "es": "Música",
  "fi": "Musiikki",
  "it": "Musica"
}
```


plugins/BEdita/Core/config/Migrations/20240409130149_CategoriesLabels.php
```diff
@@ -0,0 +1,59 @@
...
+            ->addColumn('label', 'string', [
+                'default' => null,
+                'limit' => 255,
+                'null' => true,
+                'after' => 'name',
+            ])
+            ->update();
+        // copy labels.default into label
+        $this->query(
+            'UPDATE categories SET label = JSON_UNQUOTE(labels->>"$.default")'
```
le0m: PostgreSQL: `UPDATE categories SET label = labels->>'default'`
SQLite: `UPDATE tags SET label = labels->>'$.default'`

didoda: I didn't try the rollback on those dbs :sweat_smile: Thanks, I'll patch this soon

didoda: done, on tag migration too

le0m (approved):

PR #2073, opened 2024-04-11
Improve query performance

batopa:
This PR slightly improves performance of query built in `BuildFilterBehavior::allowed_to_userFilter()`.

### DIFF
--- a/bedita-app/models/behaviors/build_filter.php
+++ b/bedita-app/models/behaviors/build_filter.php
@@ -1183,9 +1183,10 @@ class BuildFilterBehavior extends ModelBehavior {
 
                 if (!empty($forbiddenPub) || !empty($forbiddenSection)) {
 
-                    $query = "SELECT {$s}Tree{$e}.{$s}id{$e}
+                    $query = "SELECT DISTINCT {$s}Tree{$e}.{$s}id{$e}
                              FROM {$s}trees{$e} AS {$s}Tree{$e}
-                             WHERE {$s}BEObject{$e}.{$s}id{$e} = {$s}Tree{$e}.{$s}id{$e}";
+                             INNER JOIN {$s}objects{$e} AS {$s}BEObject{$e} ON {$s}BEObject{$e}.{$s}id{$e} = {$s}Tree{$e}.{$s}id{$e}
+                             WHERE 1=1";
 
                     if (!empty($forbiddenPub)) {
                         $forbiddenPubList = implode(',', $forbiddenPub);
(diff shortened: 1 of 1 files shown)


bedita-app/models/behaviors/build_filter.php
```diff
@@ -1183,9 +1183,10 @@ protected function allowed_to_userFilter($s, $e, $userid) {
 
                 if (!empty($forbiddenPub) || !empty($forbiddenSection)) {
 
-                    $query = "SELECT {$s}Tree{$e}.{$s}id{$e}
+                    $query = "SELECT DISTINCT {$s}Tree{$e}.{$s}id{$e}
                              FROM {$s}trees{$e} AS {$s}Tree{$e}
-                             WHERE {$s}BEObject{$e}.{$s}id{$e} = {$s}Tree{$e}.{$s}id{$e}";
+                             INNER JOIN {$s}objects{$e} AS {$s}BEObject{$e} ON {$s}BEObject{$e}.{$s}id{$e} = {$s}Tree{$e}.{$s}id{$e}
+                             WHERE 1=1";
```
fquffio, line 1189:
```suggestion
                             INNER JOIN {$s}objects{$e} AS {$s}BEObject{$e} ON {$s}BEObject{$e}.{$s}id{$e} = {$s}Tree{$e}.{$s}id{$e}";
```

Should work without `WHERE` clause too…

bedita-app/models/behaviors/build_filter.php
```diff
@@ -1208,9 +1209,9 @@ protected function allowed_to_userFilter($s, $e, $userid) {
                     $this->conditions[] = "(
                         (
                             NOT EXISTS (
-                                SELECT {$s}Tree{$e}.{$s}id{$e}
+                                SELECT DISTINCT {$s}Tree{$e}.{$s}id{$e}
                                 FROM {$s}trees{$e} AS {$s}Tree{$e}
-                                WHERE {$s}Tree{$e}.{$s}id{$e} = {$s}BEObject{$e}.{$s}id{$e}
+                                INNER JOIN {$s}objects{$e} AS {$s}BEObject{$e} ON {$s}BEObject{$e}.{$s}id{$e} = {$s}Tree{$e}.{$s}id{$e}
```
fquffio, line 1214: This changes the query meaning… it becomes a non-dependent subquery, which checks if _any_ object is on the tree. It is a condition that is basically never satisfied.

batopa: I think you're right :disappointed:

batopa:
No needed. I close it.

PR #2078, opened 2024-05-31
Upgrade to CakePHP 4.5

batopa:
This PR updates CakePHP to 4.5 version replacing #2043


plugins/BEdita/Core/src/ORM/Inheritance/Query.php
```diff
@@ -15,147 +15,18 @@
...
 
-use Cake\Database\Expression\QueryExpression;
+use BEdita\Core\ORM\Inheritance\Query\InheritanceQueryTrait;
 use Cake\ORM\Query as CakeQuery;
-use Cake\ORM\Table as CakeTable;
 
 /**
  * Extends `\Cake\ORM\Query` to set `FROM` clause and add default types and fields.
  *
+ * It will be removed in 6.x.
```
le0m, line 24: could use a phpdoc comment like `@deprecated 5.24.0 Will be removed in 6.x`

batopa: done, thanks!

le0m (approved):

PR #2082, opened 2024-06-26
Property default

didoda:
This PR introduces `default_value` in `properties` table.

Project model can use `default_value` as in following example:
```json
{
    "properties": [
        {
            "name": "age",
            "description": null,
            "is_nullable": false,
            "read_only": false,
            "object": "profiles",
            "property": "integer",
            "default_value": 30
        }
    ]
}
```


plugins/BEdita/Core/config/Migrations/20240626123658_UpdateProperties.php
```diff
@@ -0,0 +1,39 @@
...
+    public function up(): void
+    {
+        $this->table('properties')
+            ->addColumn('default_value', 'text', [
+                'after' => 'read_only',
+                'comment' => 'Default value',
+                'length' => null,
+                'null' => true,
+            ])
+            ->addColumn('property_options', 'text', [
```
le0m: why not setting as a json column?

```suggestion
            ->addColumn('property_options', 'json', [
```

and add `->setColumnType('property_options', 'json')` in `PropertiesTable::getSchema()`. You can then json-encoding in `Property` utility and set the `property_options` property as `array` in the entity

plugins/BEdita/Core/src/Model/Entity/Property.php
```diff
@@ -35,6 +35,8 @@
  * @property bool $is_nullable
  * @property bool $read_only
  * @property bool $required
+ * @property string|null $default_value
+ * @property string|null $property_options
```
le0m:
```suggestion
 * @property array|null $property_options
```

plugins/BEdita/Core/src/Utility/Properties.php
```diff
@@ -77,13 +77,17 @@ public static function create(array $properties, array $options = []): array
...
+            $options = is_array($options) ? json_encode($options) : $options;
             $property = $Properties->newEntity([
                 'name' => $p['name'],
                 'property_type_name' => $p['property'],
                 'object_type_name' => $p['object'],
                 'description' => Hash::get($p, 'description'),
                 'is_nullable' => (bool)Hash::get($p, 'is_nullable', true),
                 'read_only' => (bool)Hash::get($p, 'read_only', false),
+                'default_value' => Hash::get($p, 'default_value', null),
+                'property_options' => $options,
```
le0m:
```suggestion
            $property = $Properties->newEntity([
                'name' => $p['name'],
                'property_type_name' => $p['property'],
                'object_type_name' => $p['object'],
                'description' => Hash::get($p, 'description'),
                'is_nullable' => (bool)Hash::get($p, 'is_nullable', true),
                'read_only' => (bool)Hash::get($p, 'read_only', false),
                'default_value' => Hash::get($p, 'default_value', null),
                'property_options' => Hash::get($p, 'options', null),
```

plugins/BEdita/Core/src/Utility/Properties.php
```diff
@@ -136,6 +140,8 @@ public static function update(array $properties, array $options = []): array
         foreach ($properties as $p) {
             static::validate($p);
             $objectType = $ObjectTypes->get(Inflector::camelize($p['object']));
+            $options = Hash::get($p, 'options', null);
+            $options = is_array($options) ? json_encode($options) : $options;
```
le0m, line 143:
```suggestion
```

plugins/BEdita/Core/src/Utility/Properties.php
```diff
@@ -148,6 +154,8 @@ public static function update(array $properties, array $options = []): array
             $property->set('description', Hash::get($p, 'description'));
             $property->set('is_nullable', (bool)Hash::get($p, 'is_nullable', true));
             $property->set('read_only', (bool)Hash::get($p, 'read_only', false));
+            $property->set('default_value', Hash::get($p, 'default_value', null));
+            $property->set('property_options', $options);
```
le0m:
```suggestion
            $property->set('property_options', Hash::get($p, 'options', null));
```

fquffio:
While I see the purpose of `default_value`, I think that `property_options` overlaps with the JSON Schema in `property_types` in a way that might be misleading on the long run.

Both the examples you've provided would probably be best represented by a new property type `int_10_100` with a JSON schema (see [docs](https://json-schema.org/understanding-json-schema/reference/numeric#range)):

```json
{
  "type": "integer",
  "minimum": 10,
  "maximum": 100
}
```

In addition to avoiding overlapping between property types and property options, JSON Schema has a very broad set of options. Probably even _too_ broad for us to support them all, but we might easily start with a subset – like `minimum` and `maximum` for numbers and integers.

plugins/BEdita/Core/src/Utility/Properties.php
didoda: no need, as we removed `property_options`

PR #2088, opened 2024-07-23
Feat: default relation params

le0m:
This PR adds support for default values in relation parameters.

For example, adding a relation with params:
```
    "params": {
                "type": "object",
                "properties": {
                    "name": {
                        "type": "string",
                        "description": "Name of the cat",
                        "default": "Garfield"
                    }
                }
            }
```
ensures that when using the API endpoint `/{objectType}/{id}/relationships/{relationName}` to add a related object, the corresponding entry in the junction table `object_relations` will contain `{"name":"Garfield"}` in its `params` field.

The default value is applied in the `UpdateAssociatedAction` (used by the aforementioned endpoint) so that it's not too hard to circumvent this behavior.


plugins/BEdita/API/src/Model/Action/UpdateAssociatedAction.php
```diff
@@ -165,12 +166,56 @@ function ($a, $b) use ($targetPrimaryKeys) {
...
+        );
+        if (count($defaultParams) === 0) {
+            return $meta;
+        }
+        if ($meta === null) {
+            $meta = ['params' => []];
+        }
+
+        foreach ($defaultParams as $name => $data) {
+            if (empty($meta['params'][$name])) {
```
fquffio: I'm really not sure whether the `empty` check is what we really want here. This would result in a field of type `{"type": "boolean", "default": true}` to be essentially always true, cause any attempt to set it as `false` would fall into this branch. Same for an empty string, or `0` as a number.

I'd much rather prefer a `isset` check here, but I'm unsure whether the best thing to do would be to check for key existence altogether using `array_key_exists($meta['params'], $name)`.

fquffio: Assuming we can easily tell whether that parameter is nullable, the (arguably) correct solution would be:

```php
if (!array_key_exists($meta['params'], $name) || (!$paramIsNullable && !isset($meta['params'][$name]))) {
  // Use default
}
```

le0m: You're right, should be fixed now

didoda (approved):

PR #2099, opened 2024-09-30
Add column `params` to table `Trees`

le0m:
This PR adds a `Trees.params` column, to store custom parameters for children association.

The parameters' JSON-schema is read from the `ChildrenParams` configuration, if available.

### DIFF
--- a/plugins/BEdita/Core/config/Migrations/20240930121402_TreesAddParams.php
+++ b/plugins/BEdita/Core/config/Migrations/20240930121402_TreesAddParams.php
@@ -1,0 +1,26 @@
+<?php
+declare(strict_types=1);
+
+use Migrations\AbstractMigration;
+
+class TreesAddParams extends AbstractMigration
+{
+    /**
+     * Change Method.
+     *
+     * @return void
+     */
+    public function change()
+    {
+        $columnTypes = $this->getAdapter()->getColumnTypes();
+        $type = in_array('json', $columnTypes) ? 'json' : 'text';
+        $this->table('trees')
+            ->addColumn('params', $type, [
+                'comment' => 'Folder parameters (JSON)',
+                'default' => null,
+                'limit' => null,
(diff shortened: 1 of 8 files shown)


plugins/BEdita/Core/config/Migrations/20240930121402_TreesAddParams.php
```diff
@@ -0,0 +1,26 @@
...
+     *
+     * @return void
+     */
+    public function change()
+    {
+        $columnTypes = $this->getAdapter()->getColumnTypes();
+        $type = in_array('json', $columnTypes) ? 'json' : 'text';
+        $this->table('trees')
+            ->addColumn('params', $type, [
+                'comment' => 'Folder parameters (JSON)',
```
fquffio: Maybe a pedantry, but these are not "folder parameters", but rather parameters associated with the tree position of an object.

```suggestion
                'comment' => 'Parameters for the position on tree (JSON)',
```

plugins/BEdita/Core/src/Model/Table/TreesTable.php
```diff
@@ -125,9 +128,39 @@ public function validationDefault(Validator $validator): Validator
         $validator
             ->boolean('canonical');
 
+        $validator
+            ->allowEmptyArray('params', null, static::jsonSchema(null) === true)
+            ->requirePresence('params', fn ($context) => $context['newRecord'] && static::jsonSchema(null) !== true)
```
fquffio: Same, but without a closure:

```suggestion
            ->requirePresence('params', static::jsonSchema(null) === true ? false : 'create')
```

fquffio (approved):
Nice to have, but maybe not strictly needed right now: an API integration test case that attempts to set/update params on a new/existing child (i.e. `POST/PATCH /folders/:id/relationships/children`).

PR #2102, opened 2024-10-14
Sort object relationship related data

didoda:
This introduces the endpoint `/{object_type}/{id}/relationships/{relationship}/sort`.

Use this endpoint to force save related data using a specific sort.

Usage example:
```
# document 999, relation seealso, sort by title desc
PATCH /documents/999/relationships/seealso/sort
{
    "meta": {
        "field": "title",
        "direction": "desc"
    }
}
```

This also updates postman collection, adding "PATCH Sort related" in section "4. Objects".

Bonus: fix some phpunit tests warnings.

### DIFF
--- a/plugins/BEdita/API/src/Controller/ObjectsController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectsController.php
@@ -529,19 +529,22 @@ class ObjectsController extends ResourcesController
     public function relationshipsSort(): ?Response
     {
         $this->request->allowMethod(['patch']);
-        $id = $this->request->getParam('id');
-        $relationship = $this->request->getParam('relationship');
-        $payload = json_decode((string)$this->request->getBody(), true);
-        $field = (string)Hash::get($payload, 'meta.field');
-        $direction = (string)Hash::get($payload, 'meta.direction');
-        $association = $this->findAssociation($relationship);
-        $this->setRelationshipsAllowedMethods($association);
-        $action = new GetObjectAction(['table' => $this->Table, 'objectType' => $this->objectType]);
-        $entity = $action(['primaryKey' => $id]);
-        $action = new SortRelatedObjectsAction(compact('association'));
-        $count = $action(['entity' => $entity, 'field' => $field, 'direction' => $direction]);
-        if ($count === false) {
-            throw new InternalErrorException(__d('bedita', 'Could not sort and update relationship "{0}"', $relationship));
+        try {
+            $id = $this->request->getParam('id');
+            $relationship = $this->request->getParam('relationship');
+            $payload = json_decode((string)$this->request->getBody(), true);
+            $field = (string)Hash::get($payload, 'meta.field');
+            $direction = (string)Hash::get($payload, 'meta.direction');
+            $association = $this->findAssociation($relationship);
+            $this->setRelationshipsAllowedMethods($association);
+            $action = new GetObjectAction(['table' => $this->Table, 'objectType' => $this->objectType]);
+            $entity = $action(['primaryKey' => $id]);
+            $action = new SortRelatedObjectsAction(compact('association'));
+            $count = $action(['entity' => $entity, 'field' => $field, 'direction' => $direction]);
+        } catch (\Exception) {
+            throw new InternalErrorException(
+                __d('bedita', 'Could not sort and update relationship "{0}"', $relationship)
+            );
         }
         if ($count === 0) {
             return $this->response->withStatus(204);
--- a/plugins/BEdita/API/src/Controller/ObjectsController.php
+++ b/plugins/BEdita/API/src/Controller/ObjectsController.php
@@ -541,9 +541,9 @@ class ObjectsController extends ResourcesController
             $entity = $action(['primaryKey' => $id]);
             $action = new SortRelatedObjectsAction(compact('association'));
             $count = $action(['entity' => $entity, 'field' => $field, 'direction' => $direction]);
-        } catch (\Exception) {
+        } catch (\Exception $e) {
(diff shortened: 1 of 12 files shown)


plugins/BEdita/Core/src/Model/Action/SortRelatedObjectsAction.php
```diff
@@ -0,0 +1,83 @@
...
+namespace BEdita\Core\Model\Action;
+
+use BEdita\Core\Model\Entity\ObjectEntity;
+use Cake\Log\LogTrait;
+use Cake\Utility\Hash;
+
+/**
+ * Replace related objects with the same objects ordered by a specific field, with a direction.
+ *
+ * @since 4.0.0
```
batopa: minor
```suggestion
 * @since 5.31.0
```

plugins/BEdita/Core/src/Model/Action/SortRelatedObjectsAction.php
```diff
@@ -0,0 +1,83 @@
...
+     * - `direction` (string) the direction of the sort
+     */
+    public function execute(array $data = [])
+    {
+        $required = ['entity', 'field', 'direction'];
+        foreach ($required as $key) {
+            if (!array_key_exists($key, $data) || empty($data[$key])) {
+                $this->log(sprintf('Missing required key "%s"', $key), 'error');
+
+                return false;
```
batopa: Maybe we could throw a `\BEdita\Core\Exception\InvalidDataException` instead of return false

plugins/BEdita/Core/src/Model/Action/SortRelatedObjectsAction.php
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+        /** @var \BEdita\Core\Model\Entity\ObjectEntity $entity */
+        $entity = Hash::get($data, 'entity');
+        $primaryKey = $entity->get('id');
+        $field = (string)Hash::get($data, 'field');
+        $direction = (string)Hash::get($data, 'direction');
+        $association = $this->getConfig('association');
```
batopa: I would add a check on association instance. It should be a `\BEdita\Core\ORM\Association\RelatedTo` association since it describes a relation between objects.

plugins/BEdita/API/src/Controller/ObjectsController.php
```diff
@@ -502,6 +507,53 @@ public function relationships()
...
+            $this->setRelationshipsAllowedMethods($association);
+            $action = new GetObjectAction(['table' => $this->Table, 'objectType' => $this->objectType]);
+            $entity = $action(['primaryKey' => $id]);
+            $action = new SortRelatedObjectsAction(compact('association'));
+            $count = $action(['entity' => $entity, 'field' => $field, 'direction' => $direction]);
+        } catch (\Exception $e) {
+            throw new InternalErrorException(
+                __d('bedita', 'Could not sort and update relationship "{0}": {1}', $relationship, $e->getMessage())
+            );
+        }
```
batopa: I would avoid "try catch" here and I would leave that the actions decide which error thrown and therefore which status code for the response.

For example if `GetObjectAction` fails to find the object, the action will throw a `RecordNotFoundException` and the response will be a `404 Not Found`.

The same with `SortRelatedObjectsAction`. If missing some required data it throws an `InvalidDataException` that will be a `400 Bad Request` in the response and so on.

batopa (approved):

plugins/BEdita/Core/src/Model/Action/SortRelatedObjectsAction.php
```diff
@@ -0,0 +1,91 @@
...
+        $relatedEntities = $relatedEntities->toArray();
+        usort(
+            $relatedEntities,
+            function (ObjectEntity $item1, ObjectEntity $item2) use ($field, $direction) {
+                $val1 = strtolower((string)$item1->get($field));
+                $val2 = strtolower((string)$item2->get($field));
+
+                return $direction === 'asc' ? $val1 <=> $val2 : $val2 <=> $val1;
+            }
+        );
```
fquffio: I'm not sure about what the best approach is.

One one hand, the current implementation works for calculated fields as well. On the other hand, it works well for string fields but might fail for other types—which I assume are a lot less common in this scenario.

I wonder if letting MySQL do the job would yield more stable results, i.e. something that might look like this:

```suggestion
        $action = new ListRelatedObjectsAction(compact('association'));
        $relatedEntities = $action(compact('primaryKey'))
            ->sort([$association->aliasField($field) => $direction])
            ->toArray();
```

didoda: Your suggestion seems safer than the previous approach. I'm gonna adopt it

fquffio (approved):

plugins/BEdita/Core/src/Model/Action/SortRelatedObjectsAction.php
```diff
@@ -0,0 +1,81 @@
...
+            $join = $related->get('_joinData')->toArray();
+            $priorities = [
+                'priority' => $priority++,
+            ];
+            $related->set('_joinData', array_filter(array_merge($join, $priorities)));
+        }
+        $association->junction()->getBehavior('Priority')->setConfig('disabled', true);
+        $action = new SetRelatedObjectsAction(compact('association'));
+        $action(compact('entity', 'relatedEntities'));
+        $association->junction()->getBehavior('Priority')->setConfig('disabled', false);
```
fquffio: This should be wrapped inside a transaction, to ensure replaced data is consistent with what has been read.

PR #2124, opened 2025-01-21
Add auth provider username array filter for external atuh

paolosantarsiero:
This PR:
- Provide a possibility to use array with external auth provider username filter, like `GET /users?filter[external_auth][auth_provider]=example&filter[external_auth][username][]=24&filter[external_auth][username][]=25`
- fix typing in check authorization of external auth forcing cast to string


plugins/BEdita/Core/src/Model/Table/UsersTable.php
```diff
@@ -292,9 +292,16 @@ protected function findExternalAuth(Query $query, array $options = [])
...
+                if (is_array($options['username'])) {
+                    $query = $query->where(fn(QueryExpression $exp) => $exp->in(
+                        $this->ExternalAuth->aliasField('provider_username'),
+                        $options['username']
+                    ));
+                } else {
+                    $query = $query->where([
+                        $this->ExternalAuth->aliasField('provider_username') => $options['username'],
+                    ]);
+                }
```
fquffio: There's nothing wrong in using `IN` for a single element, in case you wish to make this slightly simpler.

```suggestion
                $query = $query->where(fn (QueryExpression $exp) => $exp->in(
                    $this->ExternalAuth->aliasField('provider_username'),
                    (array)$options['username'],
                ));
```

fquffio (approved):

batopa (approved):

PR #2126, opened 2025-01-29
Fix clone entity preserve original `HasMany` associations (i.e. date_ranges)

didoda:
This PR fixes a problem in cloning an object with association(s).

The problem: the clone object is fine, but the source object association(s) data is empty.

### DIFF
--- a/plugins/BEdita/Core/src/Model/Action/CloneObjectAction.php
+++ b/plugins/BEdita/Core/src/Model/Action/CloneObjectAction.php
@@ -22,9 +22,9 @@ use BEdita\Core\Utility\SchemaTools;
 use BEdita\Core\Utility\Text;
 use Cake\Datasource\EntityInterface;
 use Cake\Http\Exception\UnauthorizedException;
+use Cake\ORM\Association\HasMany;
 use Cake\ORM\Locator\LocatorAwareTrait;
 use Cake\Utility\Hash;
-use Cake\Utility\Inflector;
 
 /**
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/Model/Action/CloneObjectAction.php
```diff
@@ -131,6 +132,18 @@ protected function setEntityField(array $schemaInfo, ObjectEntity $sourceEntity,
...
+                $entityClass = get_class($data);
+                unset($data['id']);
+                unset($data['object_id']);
+
+                return new $entityClass($data->toArray());
+            }, $value);
+            $entity->set($field, $value);
+
+            return $value;
+        }
```
batopa: :thinking: Since the associations could be of many types (`HasOne`, `BelongsTo`, `HasMany`, `BelongsToMany`) it doesn't always work.

To keep it simple we could refer to `HasMany` association for now doing something as

```php
$association = $sourceEntity->getTable()->associations()->getByProperty($field);
if ($association instanceof HasMany) {
    $targetTable = $association->getTarget();
    // foreignKey could be an array... but let's keep it simple
    $foreignKey = $targetTable->getForeignKey();
    $value = array_map(function ($data) {
        unset($data['id']);
        unset($data[$foreignKey]);

        return $targetTable->newEntity($data->toArray());
    }, $value);
    $entity->set($field, $value);

    return $value;
}
```

PR #2132, opened 2025-02-27
Tree slugs and new tree_paths endpoint

le0m:
This PR is for the transition from uname-based URLs to slug-based ones. These will allow to have multiple objects with the same "name" reachable from different paths. For example, `/hello/world` and `/good/morning/world` could reference different objects based on their paths even if the last piece of the URL is the same. This is currently not possible, since the path is built using the objects' unames and those must be unique across the whole database.

This PR adds a `slug` field to the `trees` table.

The migration also populates the new field with the current `uname` value.

A controller `TreeSlugs` has been added to handle `/tree_paths/*` routes, which work like the `/trees/*` ones and will replace them when the stransition will be complete (**TODO** tests for the controller).

This change should be backward-compatible.

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Controller/TreePathsControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/TreePathsControllerTest.php
@@ -28,7 +28,6 @@ class TreePathsControllerTest extends IntegrationTestCase
      *
      * @return void
      * @covers ::index()
-     * @covers ::objectDetails()
      * @covers ::pathDetails()
      */
     public function testIndexValidSlugPath(): void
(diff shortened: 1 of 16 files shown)


plugins/BEdita/API/tests/TestCase/Controller/TreePathsControllerTest.php
```diff
@@ -0,0 +1,255 @@
...
+        ];
+
+        static::assertEquals($expected, $response);
+    }
+
+    /**
+     * Test `index` with an invalid slug.
+     *
+     * @return void
+     * @covers ::index()
```
le0m, line 126:
```suggestion
     * @covers ::index()
     * @covers ::objectDetails()
     * @covers ::pathDetails()
```

Vavaste, line 125: i've moved utility and data access methods from the controller to TreesTable and used @covers in TreesTables's test file

didoda (approved):

PR #2140, opened 2025-05-22
Add `?sort=published` for objects, to sort by `publish_start ?? created`

fquffio:
This PR adds a custom sorting option for tables that allow sorting by `publish_start` to make it possible to sort by `COALESCE(publish_start, created)`. This is useful in conjunction with `filter[ancestor]` to sort all descendants in a publication by their publication date and find the latest.

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
@@ -1047,6 +1047,219 @@ class ObjectsControllerTest extends IntegrationTestCase
     }
 
     /**
+     * Test index method.
+     *
+     * @return void
+     * @covers ::index()
+     * @covers ::initialize()
+     * @covers ::addCount()
+     * @covers ::prepareFilter()
+     * @covers ::prepareInclude()
+     */
+    public function testIndexSortPublished()
+    {
+        $expected = [
+            'links' => [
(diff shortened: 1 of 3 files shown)


plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
```diff
@@ -1046,6 +1046,219 @@ public function testIndex()
...
+                        'uname' => 'title-one',
+                        'title' => 'title one',
+                        'description' => 'description here',
+                        'body' => 'body here',
+                        'extra' => [
+                            'abstract' => 'abstract here',
+                            'list' => ['one', 'two', 'three'],
+                        ],
+                        'lang' => 'en',
+                        'publish_start' => '2016-05-13T07:09:23+00:00',
```
didoda: Maybe we could use a date different from `meta.created` for this test, just to be sure that `publish_start` is considered, instead of `meta.created`

didoda (approved):

plugins/BEdita/API/tests/TestCase/Controller/ObjectsControllerTest.php
fquffio: Really great catch! You're absolutely right, this test might well be passing by chance.

I've changed it to make it meaningful. Thanks!

PR #2142, opened 2025-06-18
Add support for 'nin' operator in fieldsFilter

Vavaste:
Implements the `nin` (not in) operator in `fieldsFilter`.

This allows users to write filters like `filter[field][nin]=x,y,z`, which internally maps to a `NOT IN` condition on the specified field.

The change is covered by unit tests under `QueryFilterTraitTest`.

### DIFF
--- a/plugins/BEdita/Core/src/ORM/QueryFilterTrait.php
+++ b/plugins/BEdita/Core/src/ORM/QueryFilterTrait.php
@@ -85,15 +85,28 @@ trait QueryFilterTrait
                 }
 
                 $in = [];
+                $nin = [];
                 foreach ($conditions as $operator => $value) {
                     if (is_numeric($operator)) {
                         $in[] = $value;
                         continue;
                     }
-                    $exp = $this->operatorExpression($exp, $operator, $field, $value);
+
+                    switch ($operator) {
+                        case 'nin':
+                            $nin = is_string($value) ? explode(',', $value) : (array)$value;
+                            break;
+
+                        default:
+                            $exp = $this->operatorExpression($exp, $operator, $field, $value);
+                    }
                 }
+
                 if (!empty($in)) {
                     $exp = $exp->in($field, $in);
+                }
+                if (!empty($nin)) {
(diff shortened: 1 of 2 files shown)


plugins/BEdita/Core/src/ORM/QueryFilterTrait.php
```diff
@@ -85,16 +85,29 @@ public function fieldsFilter(Query $query, array $options)
...
                     if (is_numeric($operator)) {
                         $in[] = $value;
                         continue;
                     }
-                    $exp = $this->operatorExpression($exp, $operator, $field, $value);
+
+                    switch ($operator) {
+                        case 'nin':
+                            $nin = is_string($value) ? explode(',', $value) : (array)$value;
+                            break;
```
fquffio: You should be able to put the same logic inside `operatorExpression()`. Doing so you would leave this method untouched and simply add a few lines there.

```php
switch ($operator) {
    case 'eq':
    // ... 

    case 'nin':
    case 'notin':  // Optionally add a few aliases
        $exp = $exp->notIn($field, is_string($value) ? explode(',', $value) : (array)$value);

    // ...
```

plugins/BEdita/Core/src/ORM/QueryFilterTrait.php
```diff
@@ -90,11 +90,13 @@ public function fieldsFilter(Query $query, array $options)
...
+
                     $exp = $this->operatorExpression($exp, $operator, $field, $value);
                 }
-                if (!empty($in)) {
-                    $exp = $exp->in($field, $in);
-                }
+            }
+
+            if (!empty($in)) {
+                $exp = $exp->in($field, $in);
```
fquffio: Careful. This should be restored where it used to be, inside `foreach ($options as $field => $conditions)`.

Vavaste: Thanks, ammended, i'm going to add a test case for this

didoda (approved):

batopa (approved):

PR #2151, opened 2025-10-06
Add PostgreSQL support in custom props finder

stefanorosanelli:
This pull request adds support for PostgreSQL in handling custom properties stored as JSON, alongside existing MySQL support. It updates both the database migration logic and the application code to ensure correct querying and filtering of JSON custom properties in both databases. The test suite is also updated to validate functionality for PostgreSQL.

* Added a new migration to change the `objects.custom_props` column type from `TEXT` to `JSONB` for PostgreSQL and to `JSON` for MySQL/MariaDB, improving performance and native JSON support.
* Updated `CustomPropertiesBehavior` to support custom property filtering for PostgreSQL, using the  `->>` operator. 
* Modified integration and unit tests

### DIFF
--- a/plugins/BEdita/Core/config/Migrations/20251006090244_CustomPropsJson.php
+++ b/plugins/BEdita/Core/config/Migrations/20251006090244_CustomPropsJson.php
@@ -21,14 +21,13 @@ class CustomPropsJson extends AbstractMigration
             // For PostgreSQL, use raw SQL with explicit casting
             $this->execute('ALTER TABLE objects ALTER COLUMN custom_props TYPE jsonb USING custom_props::jsonb');
         } else {
-            // For other databases, use Phinx changeColumn
-            $json = 'text';
-            if (in_array('json', $columnTypes)) {
-                $json = 'json';
+            // For other databases having a 'json' column type, use Phinx changeColumn
+            if (!in_array('json', $columnTypes)) {
+                return;
             }
 
             $this->table('objects')
-                ->changeColumn('custom_props', $json, [
+                ->changeColumn('custom_props', 'json', [
                     'comment' => 'object custom properties (JSON format)',
                     'default' => null,
                     'null' => true,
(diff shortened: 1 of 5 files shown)


plugins/BEdita/Core/config/Migrations/20251006090244_CustomPropsJson.php
```diff
@@ -0,0 +1,55 @@
...
+                $json = 'json';
+            }
+
+            $this->table('objects')
+                ->changeColumn('custom_props', $json, [
+                    'comment' => 'object custom properties (JSON format)',
+                    'default' => null,
+                    'null' => true,
+                ])
+                ->update();
```
batopa, line 35: Do we need to alter table also for SQLite db?

plugins/BEdita/Core/src/Model/Behavior/CustomPropertiesBehavior.php
```diff
@@ -328,23 +330,37 @@ public function findCustomProp(Query $query, array $options): Query
...
+                                [
+                                    new FunctionExpression(
+                                        'JSON_EXTRACT',
+                                        [$field => 'identifier', sprintf('$.%s', $key)]
+                                    ),
+                                ]
+                            ),
+                            new FunctionExpression('JSON_UNQUOTE', [json_encode($value)]) // trick to normalize values compared
+                        );
+                    } else {
```
batopa: There isn't need of `else` control structure, we can remove it.

plugins/BEdita/Core/config/Migrations/20251006090244_CustomPropsJson.php
stefanorosanelli: ok, we can skip if we don't have a 'json' column type

stefanorosanelli: done

batopa (approved):

PR #2152, opened 2025-10-07
Update thumbnail(s) on image upload

didoda:
This provides an enhancement on image upload by triggering an event that updates thumbnails for image.

### DIFF
--- a/plugins/BEdita/API/src/Controller/Component/UploadComponent.php
+++ b/plugins/BEdita/API/src/Controller/Component/UploadComponent.php
@@ -22,6 +22,7 @@ use Cake\Event\Event;
 use Cake\Event\EventInterface;
 use Cake\Event\EventManager;
 use Cake\ORM\Locator\LocatorAwareTrait;
+use Exception;
 use Laminas\Diactoros\Stream;
 
 /**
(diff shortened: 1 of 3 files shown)


plugins/BEdita/API/src/Controller/Component/UploadComponent.php
```diff
@@ -80,6 +82,7 @@ public function upload($fileName, ?int $objectId = null): EntityInterface
         $private = filter_var($request->getQuery('private_url', false), FILTER_VALIDATE_BOOLEAN);
         $entity->set('private_url', $private);
         $data = $action(compact('entity', 'data'));
+        EventManager::instance()->dispatch(new Event('Thumbnails.update', $this, ['data' => $data]));
```
batopa: Maybe I would call `'stream'` the associative key of the array

plugins/BEdita/Core/src/Event/ImageThumbsHandler.php
```diff
@@ -65,6 +66,33 @@ public function afterSaveAssociated(Event $event): void
...
+            return;
+        }
+        $objectId = $stream->get('object_id');
+        if (!$objectId) {
+            return;
+        }
+        $image = $this->fetchTable('Objects')
+            ->find('type', ['images'])
+            ->where(['id' => $objectId])
+            ->first();
```
batopa: Maybe it's more safe to get all image data since we are going to save it.

```suggestion
        $image = $this->fetchTable('Images')
            ->find('type', ['images'])
            ->where(['id' => $objectId])
            ->first();
```

plugins/BEdita/API/src/Controller/Component/UploadComponent.php
batopa: I would add a try catch (logging the error) since `ImageThumbsHandler::updateThumbs()` throws an exception if there is a failure saving the image object.
I think that the thumbs creation must not block the upload action.

batopa (approved):

PR #2168, opened 2025-11-24
Fix thumbnail generation that needs to read EXIF data

batopa:
This PR is a porting of  #2166 into 6 version.


batopa:
Under investigation. In BEdita 6 seems that the behavior is correct.

PR #2172, opened 2026-01-02
Add custom properties sorting support

stefanorosanelli:
This pull request adds support for sorting on custom properties in API pagination, specifically for tables using the `CustomProperties` behavior and compatible with MySQL or PostgreSQL. It introduces a new method to generate database order clauses for custom property fields and extends the paginator and its tests to handle this feature. The most important changes are:

### Feature: Custom Properties Sorting Support

* Added logic in `JsonApiPaginator::validateSort()` to detect when sorting is requested on a custom property and, if so, generate an appropriate order clause using the new `customPropOrderClause()` method from the `CustomPropertiesBehavior` class. This only applies for MySQL and PostgreSQL drivers.
* Introduced the `customPropOrderClause()` method in `CustomPropertiesBehavior` to create a database order clause expression for a given custom property and direction, ensuring compatibility with MySQL and PostgreSQL.
* Updated imports and dependencies to include the necessary database driver and expression classes in relevant files. 

### Testing Improvements

* Enhanced integration and unit tests to cover sorting on custom properties, including driver checks to ensure tests only run for supported databases. Added new test cases and fixtures for profiles and properties.

### DIFF
--- a/plugins/BEdita/API/tests/TestCase/Datasource/JsonApiPaginatorTest.php
+++ b/plugins/BEdita/API/tests/TestCase/Datasource/JsonApiPaginatorTest.php
@@ -198,7 +198,7 @@ class JsonApiPaginatorTest extends TestCase
      *
      * @return void
      */
-    public function testValidateSortCustomProperty()
+    public function testValidateSortCustomProperty(): void
     {
         $paginator = new JsonApiPaginator();
         $repository = TableRegistry::getTableLocator()->get('Profiles')->find()->getRepository();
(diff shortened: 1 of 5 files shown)


plugins/BEdita/API/tests/TestCase/Datasource/JsonApiPaginatorTest.php
```diff
@@ -187,6 +193,27 @@ public function testValidateSort($expected, $sort = null)
         static::assertEquals($expected, $options['order']);
     }
 
+    /**
+     * Test `validateSort()` method with a CustomProperties key.
+     *
+     * @return void
+     */
+    public function testValidateSortCustomProperty()
```
batopa:
```suggestion
    public function testValidateSortCustomProperty(): void
```

batopa (approved):

PR #2205, opened 2026-09-10
feat: Async job options

didoda:
This PR add a column `job_options` (`options` is a reserved word) to `async_jobs` table. 

When scheduling the job, if available the job options are saved (to specific use in job process).


batopa (changes requested):
It seems ok! I would add `->setColumnType('job_options', 'json')` in `AsyncJobsTable::initialize()` to treat `job_options` as `json` also if not supported as in SQLite.

batopa (approved):

